PR #695, opened 2014-10-24
API -- GET url

vzvenyach:
This is a simple API endpoint that allows you to return a GUID for any arbitrary URL and returns a 404 if it doesn't exist.


phillipsm:
Yeah!

We're rolling out a full featured API soon. Would love to get your thoughts on it.

vzvenyach:
That'd be awesome!

PR #780, opened 2015-01-25
Remove concept of groups.

jcushman:
Instead of using groups, test user permissions by whether users have `is_staff`, `registrar_id`, or `vesting_org_id` set.

The best way to test the latter two is with `is_registrar_member()` and `is_vesting_org_member()`, so we can just upgrade those functions when we support multiple vesting orgs and such.

The only visible change here should be that the user edit form for staff no longer has a group dropdown, and staff now see an "edit in admin console" link in the user list. In order to change someone's status, you can click that and add or remove registrar/vesting org with the Django admin.


phillipsm:
Flippin' awesome!

PR #782, opened 2015-02-02
API permissions and validation.

jcushman:
A bunch of changes to fit the API to Perma's permissions model:

- Remove /registrars, /assets, and /vesting_orgs endpoints.
- Require user authentication for remaining endpoints -- /archives, /folders, and /user/.
- Add a /public/archives endpoint that does not require authentication. This endpoint excludes the archived_by, vested_by_editor, created_by, and notes fields, and excludes unvested archives.
- Filter out results from all archive endpoints where `user_deleted=True`.
- Remove `folders` field from all archive endpoints, because it varies from user to user. (TODO: This could be added back as /user/archives/<id>/folders if it's useful.)
- Additional folder authorization -- can't move folder with vested links outside of vesting org, handle moving folders with subfolders, moving folders into themselves.
- When vesting, 'folder' is only required if the archive isn't currently in a folder belonging to the vesting org.
- Some errors are now validation (400 Bad Request) instead of authorization (401 Unauthorized), so they can return helpful error messages to the client. (E.g. renaming user's root folder.)
- folder.parent is now returned as an ID (29) instead of a URL ("/api/v1/folders/29"). This makes it match folder.id, which is handy when you're munging a list of folders into a tree.

Frontend changes:

- Frontend JS should now show useful error messages on BadRequest errors (e.g. dragging a vested link out of the shared folder).
- Vesting folder dropdown now shows folders with tree indentation. (I decided to try doing this in frontend code to see how messy it was, vs. returning a tree structure from the API.)

Implementation details:

- Killed the unused folder 'slug' field in the Folder model.


perma_web/api/authorizations.py
```diff
@@ -1,124 +1,107 @@
...
         return True
 
     def delete_detail(self, object_list, bundle):
+
+        if not self.can_access(bundle.request.user, bundle.obj):
+            raise Unauthorized()
+
         if bundle.obj.is_shared_folder:
-            raise Unauthorized("Shared folders cannot be deleted.")
+            raise BadRequest("Shared folders cannot be deleted.")
```
leppert: Seems weird to throw `BadRequest` in an `Authorization` rather than a `Validation`. I seem to remember talking about this, and if there was a reason such as an early exit to avoid an expensive DB call that would take place before reaching the `Validation`, then I'm okay with that. Otherwise, it should probably be moved to a `Validation`.

perma_web/api/authorizations.py
```diff
@@ -1,124 +1,107 @@
...
-        if user.is_vesting_org_member():
-            return user.vesting_org == vesting_org
-        elif user.is_registrar_member():
-            return user.registrar == vesting_org.registrar
-        elif user.is_staff:
-            return True
-        else:
-            return False
+    def read_list(self, object_list, bundle):
+        return object_list.filter(vested=True)
```
leppert: That was easy enough.

perma_web/api/authorizations.py
jcushman: I tried doing this in `FolderValidation.is_valid` first, but Tastypie doesn't seem to call validations for DELETE -- they're triggered by trying to save the modified object.

PR #787, opened 2015-02-07
Fix mirroring + API integration bugs.

jcushman:
Couple random bugs I ran across when checking out the cross-origin thing.


jcushman:
Resending updated pull.

PR #816, opened 2015-02-23
Mirror updates.

jcushman:


jcushman:
I just pushed some more changes that make this into a more substantial pull req.

- Fix bug with display of preview image on creation page when mirroring is enabled.
- Add a basic "integrity check" as a celery task, where mirror will progressively download missing archive files.
- Add a mirror status page for registry users on the dashboard to view progress of archive syncing.

PR #825, opened 2015-03-02
Thumbnailing logic refactor

phillipsm:
Created a basic service to handle thumbnail creation.

We're using similar logic as we were before, on top of the sorl-thumbnail library, but we first check our cache to make sure we're not kicking off the same thumbnailing job more than once.


phillipsm:
Ack. Just spotted some logic that's off.

Issue #775, opened 2015-01-21
User guide styling broken. Buttons.


phillipsm:
Something's up with the styling on the buttons/menu on the user guide.
![screen shot 2015-01-21 at 3 46 55 pm](https://cloud.githubusercontent.com/assets/196089/5844817/e35eba38-a184-11e4-93dc-9ae7716f3f88.png)

abziegler:
I'd like to amp up our user guide/docs to hopefully help manage some of the inbound email traffic. Any objection to me taking a run at it, including a single page restyling approach using this template with our styles: http://codyhouse.co/gem/css-faq-template/ ?

kdulin:
None from me

phillipsm:
Go for it.

A couple of notes:
1. I used the basics of the manual for the API docs (/docs/developer on the staging server). We may or may not want to keep the doc navigation consistent.
2. It can be really nice to provide an anchors into the FAQ/user guide. We do this with the FAQ now, https://perma.cc/docs/faq#delete-links

PR #826, opened 2015-03-02
Thumbnailing logic refactor

phillipsm:
Created a basic service to handle thumbnail creation.

We're using similar logic as we were before, on top of the sorl-thumbnail library, but we first check our cache to make sure we're not kicking off the same thumbnailing job more than once.


phillipsm:
Sorry. I should have split this across a couple of branches. cd3d857a2bbb95603b1f8375bd3d6e364b880413 is a quick fix for #775

PR #827, opened 2015-03-03
Thumbnails2

jcushman:
Here's a revised version of Matt's pull req. I added image size checks in thumbnail creation and also in capturing. It would be nice to allow partial captures, but for now simply refusing to take screenshots above 50 megapixels is better than crashing.


phillipsm:
Nice. Thanks.

PR #863, opened 2015-04-21
Reduce celery logging.

jcushman:
The celery log was getting huge because warc captures were doing a ton of play-by-play logging, including a line for each resource in the page. This should cut those down.


jcushman:
(Note that the build errored because Travis failed to build the pip requirements -- not actually a problem with the pull req.)

PR #865, opened 2015-04-23
Add memento endpoints

jcushman:
Opening a pull req for comments ...


perma_web/perma/settings/deployments/settings_testing.py
```diff
@@ -8,6 +9,9 @@
     PROJECT_ROOT,
 )
 
+# Default to local functional tests
+USE_SAUCE = os.environ.get('USE_SAUCE', 'False') == 'True'
```
jcushman: Note that any setting can be overridden by an env variable already, via something like DJANGO__USE_SAUCE=True. (This is enabled by https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/settings/utils/environmental_settings.py)

Are the ergonomics worth special-casing this?

perma_web/perma/tasks.py
```diff
@@ -352,16 +352,22 @@ def meta_thread():
...
         print "Saving WARC."
         try:
-            temp_warc_path = os.path.join(warc_writer.directory, warc_writer._f_finalname)
+            temp_warc_path = os.path.join(warc_writer.directory,
+                                          warc_writer._f_finalname)
             with open(temp_warc_path, 'rb') as warc_file:
                 warc_name = default_storage.store_file(warc_file, warc_path)
                 save_fields(asset, warc_capture=warc_name)
+
+            print "Writing CDX lines to the DB"
```
jcushman: Let's comment this out -- I went and commented out all the other play-by-play in this task.

perma_web/warc_server/pywb_config.py
```diff
@@ -17,58 +15,149 @@
...
+            link = Link.objects.get(guid=guid)
+        except Link.DoesNotExist:
+            raise NotFoundException()
+
+        try:
+            line = CDXLine.objects.get(urlkey=urlkey,
+                                       asset__link_id=guid)
+        # Legacy archives didn't generate CDXLines during
+        # capture so generate them on demand if not found
+        except CDXLine.DoesNotExist:
```
jcushman: Does this distinguish between an index we haven't created yet, and an asset we really don't have? It's common for a complex playback like /ABCD-1234/http://cnn.com/ to trigger a bunch of requests like /ABCD-1234/http://assets.cnn.com/something-we-missed.css -- sometimes an optimistic javascript will even kick off continual requests for the same missing asset. We don't want to be re-running the indexing step each time.

Unless I'm missing something, maybe we need to add an asset.cdx_created field to avoid extra indexing? Or we could run the index only if there are no CDXLines at all -- not sure if that opens up weird edge cases though?

(Let's chat when you have a minute about the best way to normalize archives and assets in general -- I don't love what we have now, but not sure if there's a better way.)

perma_web/warc_server/pywb_config.py
```diff
@@ -17,58 +15,149 @@
...
+        urlkey = surt(wbrequest.wb_url_str)
+
+        try:
+            # This will filter out links that have user_deleted=True
+            link = Link.objects.get(guid=guid)
+        except Link.DoesNotExist:
+            raise NotFoundException()
+
+        try:
+            line = CDXLine.objects.get(urlkey=urlkey,
```
jcushman: This query and the custom_params cache need to handle multiple lines with the same urlkey. For example, here are the first two CDX lines for a capture of "google.com", which redirects to "www.google.com":

```
com,google)/ 20150423160350 http://google.com/ text/html 301 c79f5572f672361bc097676cb5da9d4aa956c8b9 - - 712 1673 2015/4/23/16/3/BY3G-VVCQ/archive.warc.gz
com,google)/ 20150423160350 http://www.google.com/ text/html 302 43c3c326039fde13b0473b08d729e6fcd7c643b3 - - 1012 2960 2015/4/23/16/3/BY3G-VVCQ/archive.warc.gz
```

Note that the surt for both lines is the same because the www is normalized away. In a search for `com,google)/` we need to return *both* lines -- then pywb will detect the self-redirect in the first line, discard it, and move on to the second one. Strange but true.

services/travis/settings.py
```diff
@@ -18,4 +19,7 @@
 HOST = 'perma.cc'
 
 # Where we store our generated assets (phantomjs images)
-MEDIA_ROOT = '/tmp/perma/assets'
\ No newline at end of file
+MEDIA_ROOT = '/tmp/perma/assets'
+
+SAUCE_USERNAME = os.environ.get('SAUCE_USERNAME')
```
jcushman: As above, we can just use DJANGO__SAUCE_USERNAME without special handling for these.

perma_web/perma/settings/deployments/settings_testing.py
leppert: Nope, didn't realize we could do that. I'll revise.

perma_web/perma/tasks.py
leppert: Will do.

services/travis/settings.py
leppert: Got it.

leppert:
Woohoo! Let me know if there's any trouble getting it set up, testing on Semaphore.

PR #869, opened 2015-05-08
Freshen Vagrant.

jcushman:
- New build of our vagrant box based on Ubuntu 14, includes imagemagick, updated Python packages.
- Temporarily disable integration tests that were giving me trouble.
- Some Heroku tweaks.


jcushman:
(Pulled this right in since it's about testing/dev-env stuff that doesn't affect live site, but could be annoying for devs if left out.)

PR #871, opened 2015-05-13
Adds multi vesting org logic

### LINKED ISSUE #684: Allow vesting users to be associated with multiple vesting orgs
Right now we allow vesting users to be associated with only one vesting org at a time. It's common, however, for students to work for multiple journals, for faculty to be associated with multiple groups, and for lawyers to be associated with multiple firms, etc. 

We should support multiple org associations. At time link is vested, a vesting user should select which org applies.

phillipsm:
fixes #684


perma_web/perma/forms.py
```diff
@@ -200,18 +248,44 @@ class UserAddVestingOrgForm(forms.ModelForm):
...
+        if 'target_user_id' in kwargs:
+            target_user_id = kwargs.pop('target_user_id')
+
+        super(UserAddVestingOrgForm, self).__init__(*args, **kwargs)
+
+        target_user = LinkUser.objects.get(pk=target_user_id)
+
+        # Vesting managers can only edit their own vesting members
+        if registrar_id:
+            # Get the union of the user's and the registrar member's vesting orgs
```
jcushman: This seems like getting the difference, not the union? I'm not clear what the 'exclude' is for.

Since it's the same in all three branches, maybe move it down after the if/else block, like

`vesting_orgs = vesting_orgs.exclude(pk__in=target_user.vesting_org.all())`

with a comment explaining what we're excluding?

perma_web/api/authorizations.py
```diff
@@ -17,7 +17,9 @@ def can_access(self, user, obj):
         # shared folders
         elif obj.vesting_org_id:
             if user.is_vesting_org_member():
-                return user.vesting_org_id == obj.vesting_org_id
+                for vo in user.vesting_org.all():
```
jcushman: Ideally we would rename `vesting_org` to `vesting_orgs`. Up for a big, careful find-and-replace?

jcushman: It's probably more efficient to replace these three lines with `return user.vesting_orgs.filter(pk=obj.vesting_org_id).exists()` -- it avoids fetching the data and instantiating an object for each vesting org, just a SQL count query.

perma_web/perma/models.py
```diff
@@ -215,21 +221,21 @@ def all_folder_trees(self):
...
             vesting_orgs = self.registrar.vesting_orgs.all()
         else:
-            vesting_orgs = [self.get_default_vesting_org()]
+            vesting_orgs = list(self.get_default_vesting_org())
+
+
         return [self.root_folder.get_descendants(include_self=True)] + \
             ([vesting_org.shared_folder.get_descendants(include_self=True) for vesting_org in vesting_orgs if vesting_org])
 
     def get_default_vesting_org(self):
```
jcushman: It looks like this function should now be renamed to something like `get_accessible_vesting_orgs` -- that's how it's being used in `all_folder_trees`, `LinkManager.user_access_folder`, and `FolderManager.user_access_folder` where it's called.

This also means that:

- `if self.is_registrar_member():` should return `self.registrar.vesting_orgs.all()`
- `all_folder_trees`, `LinkManager.user_access_folder` and `FolderManager.user_access_folder` should drop the `is_registrar_member()` branch -- they can use the same logic for everyone

perma_web/perma/models.py
```diff
@@ -276,7 +282,7 @@ def is_registrar_member(self):
 
     def is_vesting_org_member(self):
         """ Is the user a member of a vesting org? """
-        return bool(self.vesting_org_id)
+        return bool(self.vesting_org.all())
```
jcushman: Should be more efficient to do `return self.vesting_org.exists()`

(I'm pretty sure that works -- otherwise `bool(self.vesting_org.count())`)

perma_web/api/authorizations.py
phillipsm: Yes. Vesting orgs makes sense. Added an issue so we don't lose track, https://github.com/harvard-lil/perma/issues/873

perma_web/perma/models.py
phillipsm: For sure. It's getting awfully crufty in there. I'm out of steam now. Punting this cleanup to the next release with this issue, https://github.com/harvard-lil/perma/issues/876

jcushman:
Just pulled in these changes ...

PR #879, opened 2015-05-19
Use NFS and more CPU/RAM for Vagrant speed.

jcushman:


jcushman:
(Pulled this right in since it doesn't affect live code.)

PR #931, opened 2015-07-14
Fix bug that stopped registry member from adding a registrar member

phillipsm:
If you're logged in as a registry member you would receive a 500 if attempting to promote a user to a registrar member. This patch fixes the bug that was causing the 500.

### DIFF
--- a/perma_web/perma/views/user_management.py
+++ b/perma_web/perma/views/user_management.py
@@ -687,20 +687,19 @@ def registrar_user_add_user(request):
         if ((form and form.is_valid()) or form == None) and not cannot_add:
             if target_user == None:
                 target_user = form.save()
-                is_new_user = True
-    
-            if request.user.is_registrar_member():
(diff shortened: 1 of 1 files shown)


perma_web/perma/views/user_management.py
```diff
@@ -687,20 +687,19 @@ def registrar_user_add_user(request):
...
             if target_user == None:
                 target_user = form.save()
-                is_new_user = True
-    
-            if request.user.is_registrar_member():
-                target_user.registrar = request.user.registrar
-    
-            if is_new_user:
                 target_user.is_active = False
-                email_new_user(request, target_user)
```
jcushman: What's the logic here? It seems like if an admin or registrar user creates a *new* account, we need to send the email_new_user message to let the new account be activated. And if an existing account is upgraded to registrar (whether it's a peer or an admin doing it), we should send the "you've been upgraded" message from email_new_registrar_user.

PR #946, opened 2015-07-21
Add missing protocol to urls built in JS.

jcushman:


leppert:
Any reason not to make these protocol-relative or even explicitly https?

jcushman:
Yeah, they're the actual Perma Links for display to the user. Last time we talked the majority view was to make those http.

PR #956, opened 2015-07-22
Warc storage for Perma Links

jcushman:
For review and (eventual) merging ...

- All capture tasks are routed through PhantomJS (no separate PDF capture)
- All disk assets for a Perma Link are stored in a single .warc file 
- Refactor thumbnailing code to create and cache thumbnails drawing from the .warc file
- Some misc JS/HTML changes

There's also a migration element to this -- each existing Link needs export_warc() called to generate the new .warc bundle from the old disk assets. Right now that's just done on-demand, to allow for a smooth deploy process, but I should add a fab task to migrate all the existing Links in a batch as well.


phillipsm:
Required an upgrade of PhantomJS. 1.9.8 is working beautifully for me on OS X.

PR #960, opened 2015-07-29
Learn more and sign up

anniejocain:
Provide more info for courts, journals and faculty.


perma_web/perma/forms.py
```diff
@@ -75,6 +75,41 @@ class Meta:
...
+
+    def __init__(self, *args, **kwargs):
+        super(CreateUserFormWithCourt, self).__init__(*args, **kwargs)
+        self.fields['requested_account_note'].label = "Your court"
+        self.fields['first_name'].label = "Your first name"
+        self.fields['last_name'].label = "Your last name"
+        self.fields['email'].label = "Your email"
+        
+        
+class CreateUserFormWithUniversity(CreateUserForm):
```
jcushman: Couple notes on this class:

- docstring should be updated
- seems very similar to CreateUserFormWithCourt -- maybe it should derive from that one or vice versa?
- it's not clear why we're changing the labels for first_name, last_name, and email in CreateUserFormWithCourt but not the other three -- maybe add a comment if that's a special case?

perma_web/perma/templates/registration/check_email.html
```diff
@@ -9,6 +9,11 @@
 			<h1>Thanks for creating an account</h1>
 			
 			<p>Look for an email from Perma and click on the link to activate your account.</p>
+			    {% if messages %}
+        {% for message in messages %}
+            <p>{% if 'safe' in message.tags %}{{ message|safe }}{% else %}{{ message }}{% endif %}</p>
+        {% endfor %}
+    {% endif %}
```
jcushman: Nitpick -- let's consistently indent blocks with 4 spaces in templates, the same way we do in Python. So, `{% if %}` would line up with `<p>`, and `{% endif %}` would line up with `{% if %}`. It's not crazy confusing here, but it can definitely lead to confusion and bugs as templates get edited.

perma_web/perma/templates/registration/sign-up-courts.html
```diff
@@ -0,0 +1,69 @@
...
+				  {% endif %}<br/>
+				{% endfor %}  
+				{% if not request.user.is_authenticated %}
+				<div class="checkbox">
+                    <label>
+                      <input id="id_create_account" name="create_account" type="checkbox" checked> Create a free personal account to try
+                    </label>
+                </div>
+                <p class="terms-statement"><small>By registering, you agree to the <a href="{% url 'terms_of_service' %}">terms of service</a>.</small></p>
+                {% endif %}
```
jcushman: Ditto the indenting nitpick throughout this file.

perma_web/perma/urls.py
```diff
@@ -70,11 +70,16 @@
     url(r'^login/account-is-deactivated/?$', 'user_management.account_is_deactivated', name='user_management_account_is_deactivated'),
     url(r'^logout/?$', 'user_management.logout', name='logout'),
     url(r'^register/?$', 'user_management.register', name='register'),
+    url(r'^sign-up/?$', 'user_management.sign_up', name='sign_up'),
+    url(r'^sign-up/courts/?$', 'user_management.sign_up_courts', name='sign_up_courts'),
+    url(r'^sign-up/faculty/?$', 'user_management.sign_up_faculty', name='sign_up_faculty'),
+    url(r'^sign-up/journals/?$', 'user_management.sign_up_journals', name='sign_up_journals'),
```
jcushman: As a general note, there's a whole lot of repeated code and HTML among these four views. This invites bugs in the future, because some things have to be updated in four different places when they change (without any explicit warning), so it's easy for the views to drift. Meanwhile, other things should only be updated in one place when they change. As the code drifts (and it's not fresh in your mind, or someone else has to update it), it becomes harder to figure out whether something should be fixed in one place or four places.

Taking a random example, if I was maintaining this later, it wouldn't be clear to me why some of the templates have `$('input[type=...]:first').focus()` in the script block, while others have that commented out. It looks like that would make sense in all of them, but maybe there's a good reason for the variation that I can't see right away. If they were drawing from the same form, the template might look like `{% if some_reason %}$('input[type=text]:first').focus();{% endif %}`, making the difference between the forms self-documenting.

So, my thought for this kind of thing would be to have a single URL route, like `r'^sign-up/?(?P<group>courts|faculty|journals)?/?$'`, and then a single view like `def sign_up(request, group='default')`, and a single template, with the form logic and template varying based on the `group` variable.

Not a huge deal, just an option to keep in mind ...

phillipsm:
Merged this in. Let's work through some of the style issues and comments in future pull reqs. Find the right balance for the team.

PR #977, opened 2015-08-14
Rename heroku settings file.

jcushman:
Just a tweak so we don’t have so many settings.py files …


phillipsm:
Cool. Let's put this in v0.54.

PR #980, opened 2015-08-17
Org and registrar users can see and edit their own info

### LINKED ISSUE #951: Links to orgs in manage users interface are broken
If I'm signed in as an org user and I click on a link to an org, I get redirected to the login screen.

`user_management.manage_organization` doesn't allow for the org user type and it probably should (?).

anniejocain:
Link orgs in manage users panel.


phillipsm:
Cool. Let's put this in v0.54.

PR #1013, opened 2015-08-28
Dow styles header

thisisdano:
In general, my technique for scoping design changes to a specific element is to maintain a broad sass/css file, but target its contents only at elements that include (or fall within) a class called `__permaNew`.

These changes are targeted at `header.__permaNew`.

I've made some html changes to the header, in addition to the styles, and modified bootstrap3.js to add/remove a class flag on `header` at a toggle trigger, and to remove the transition animation.

Please review these js changes. The `header` flag should only be added when the mobile nav expand button is triggered. Not for other expand/collapse items that might be triggered in other places on the site.


thisisdano:
OK — I think I fixed it so the dropdown menu looks good at all widths, at all signin/credential levels.

PR #1026, opened 2015-09-04
Upgrade dependencies.

phillipsm:
Forcing the inclusion of a recent version of pyscss to fix sass compilation strangeness on some machines. Thanks, @jcushman .

Also included a slight warcprox version bump. 1.3's been working
well on my machine.


jcushman:
Cool. I'm going to kick in a warcprox 1.4 update at some point, which requires a slight code change -- I was hoping it would fix the SSL issues we've had with some sites. TL;DR turns out it doesn't.

PR #1034, opened 2015-09-12
Make heroku deploys easier

leppert:
Everything that was handled in the `heroku_push` fab task is now done within [heroku-buildpack-perma](https://github.com/leppert/heroku-buildpack-perma) on push to Heroku (ex: `git push heroku develop:master --force`)

This has the advantage of making deploys to Heroku much faster as the entire repo doesn't need to be pushed each time as with the previous method. The only compromise is that the buildpacks file needs to live at `/.buildpacks` rather than under `/services/heroku/buildpacks`. This is to enable `heroku-buildpack-multi` to kick in when the app is git pushed.


jcushman:
Nice!

PR #1042, opened 2015-09-14
Debug captures on heroku

leppert:


services/heroku/heroku_settings.py
```diff
@@ -74,6 +74,7 @@
 AWS_QUERYSTRING_AUTH = False
 
 # archive creation
+PHANTOMJS_BINARY = '/app/vendor/phantomjs/bin/phantomjs' # celery won't find this otherwise
```
leppert: @jcushman I'm not thrilled about hardcoding this, but the celery worker doesn't seem to have phantomjs in its path for some reason ([see rollbar](https://rollbar.com/perma-staging/perma-staging/items/10/)) despite the fact that Django seems to find it just fine from the shell. Any ideas?

jcushman: It looks like the PATH is relative: 

```
~ $ echo $PATH
/app/.heroku/python/bin:/usr/local/bin:/usr/bin:/bin:vendor/phantomjs/bin
```

So it probably works depending on current working directory. 

Relative PATH is an anti-pattern, right? Presumably this is being set by the phantomjs buildpack we're using. Maybe fork it and send the change upstream? Or just switch us to another solution if you think that makes sense. Thanks!

(On the horizon are (a) upgrading to PhantomJS 2.0 or (b) just switching to headless Firefox, which I'm leaning toward. So the simplest fix is probably best for now.)

leppert:
@jcushman this is fixed and ready to go in

jcushman:
Thanks!

PR #1081, opened 2015-09-26
Reconfigure for adhoc heroku pipeline deploys

leppert:


leppert:
@jcushman next hurdle is removing the need for `HOST` since that'll be dynamic per instance, but I'll do that in a separate branch. Let me know if any gotchas come to mind there.

PR #1142, opened 2015-10-01
Use whitenoise to deliver static assets

jcushman:
This uses the WhiteNoise project for serving static assets, [which is recommended as a best practice by Heroku](https://devcenter.heroku.com/articles/django-assets). Static assets are now served by the same WSGI process that serves our Django app with no special handling needed, instead of a separate `/static` route in nginx or an S3 domain.

(We do still need to run collectstatic if DEBUG is off, as usual.)

On Heroku in particular, this is a win because the static assets end up bundled into the app slug. This way we don't have to re-run collectstatic when promoting a slug from perma-staging to prod.

### DIFF
--- a/perma_web/perma/settings/deployments/settings_common.py
+++ b/perma_web/perma/settings/deployments/settings_common.py
@@ -56,9 +56,9 @@ MONITOR_ROOT = '/tmp/perma/monitor'
 MONITOR_URL = '/monitor/media/'
 
 # static files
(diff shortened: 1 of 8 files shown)


jcushman:
(n.b. still debugging this on Heroku -- don't pull yet)

jcushman:
This is now ready to pull, and confirmed working on perma-stage.org. Recommend pulling this before other code, as Heroku is now configured with a new STATIC_URL setting based on this change.

PR #1159, opened 2015-10-02
Show capture progress in button.

jcushman:
This is what's currently running on perma-test ...

### DIFF
--- a/perma_web/api/resources.py
+++ b/perma_web/api/resources.py
@@ -344,7 +344,7 @@ class AuthenticatedLinkResource(BaseLinkResource):
     # folders = fields.ToManyField(FolderResource, 'folders', readonly=True, null=True)
     archive_timestamp = fields.DateTimeField(attribute='archive_timestamp', readonly=True)
     is_private = fields.BooleanField(attribute='is_private')
-    capture_progress_display = fields.CharField(attribute='capture_progress_display', blank=True)
+    capture_progress_display = fields.CharField(attribute='capture_progress_display', null=True, blank=True)
 
     class Meta(BaseLinkResource.Meta):
         authorization = AuthenticatedLinkAuthorization()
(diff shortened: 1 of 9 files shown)


phillipsm:
The status updates are cool. I like seeing them as a pro user with behind the scenes knowledge. My gut reaction though is that the user won't *get it* and that it'll lead to confusion. Best to keep it simple.

Would love to get @cdemarco's, @anniejocain's, @abziegler's, and @thisisdano's take.

thisisdano:
I'm curious to see it... I like the idea of seeing progress, but maybe not at the expense of clarity.

jcushman:
You can see a demo of this idea at https://perma-test.law.harvard.edu/manage/create . Maybe we should show it at the Perma meeting and kick it around. The motivation here is that we now make people wait up to a minute longer before seeing any kind of positive feedback that their capture is working, which is like infinity in internet time -- hard to smooth over with just a spinner. Curious to kick around solutions for that.

jcushman:
A nice middle ground that Matt mentioned before might be a progress bar, tied to the same capture steps I'm showing now in this demo. For a geeky easter egg we could throw in the actual capture steps as a tooltip.

thisisdano:
I kinda like the changing update message. We should think about the language -- communicating progress in addition to state. maybe even better when combined with progress bar?

anniejocain:
I think it could work. Another idea would be to just say 'step 1 of 5' either with or without a simplified description

phillipsm:
Let's freshen this up. Maybe we sketch it out with @jcushman and @anastasia ?

phillipsm:
Let's rework this back in. Maybe get @anastasia's take. Or, let's close this. (I think we should adjust a little and merge it in.)

anastasia:
Assigning this to me because apparently I've been missing this conversation

PR #1167, opened 2015-10-03
Dev.post vesting

thisisdano:
Just a few little fixes after the heat of Wednesday; and adapting to the new Create/Manage page. (Good idea!)

### DIFF
--- a/perma_web/perma/templates/user_management/user_add_registrar.html
+++ b/perma_web/perma/templates/user_management/user_add_registrar.html
@@ -4,22 +4,24 @@
 
 {% block content %}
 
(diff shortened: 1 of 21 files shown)


thisisdano:
I can't tell what what this error is — any clues?

perma_web/perma/templates/user_management/user_add_registrar.html
```diff
@@ -1,25 +1,27 @@
-{% extends "admin-layout.html" %}
+{% extends "maange-layout.html" %}
```
jcushman: This is causing the test error -- typo in "manage".

PR #1174, opened 2015-10-05
Record page permissions and tests.

jcushman:
- Add unit tests for the record page -- logged out and logged in, normal, dark archived, deleted.
- Fix 500 error for dark archived pages.
- Fix permissions checks to include dark_archived_robots_txt_blocked (although really we want to change to a `visible` flag and maybe a `visibility_reason` choicefield, right?)


phillipsm:
> (although really we want to change to a visible flag and maybe a visibility_reason choicefield, right?

It feels like it.

PR #1182, opened 2015-10-08
More efficient playbacks.

jcushman:
- Cache CDX lookups so we don't have to hit the database for each sub-resource when playing back a page.
- Close Django database connections in pywb playbacks (since they're outside a request and don't close manually).

This should fix instability we've been having on stage caused by exceeding max_user_connections.


jcushman:
(pulling this in to see if it fixes playback problem I'm having on stage)

PR #1203, opened 2015-10-14
Infinite scroll for folder links.

### LINKED ISSUE #1141: Pagination in folders
Folders should only show the first N links until you scroll down.

(Sort of high priority because this will cause the Create page to display slowly for people with a ton of existing, unorganized links.)

jcushman:
Folders now show only their first 20 links, with more loaded when you get to the bottom. 

Closes #1141


phillipsm:
yeah! waypoints is the way to go. nice.

Issue #1210, opened 2015-10-20
Logo carousel not responding to admin console upload


cdemarco:
I've been uploading logos using the admin console upload, but they're not getting pulled into the carousel - NY Law school has asked a couple of times, and I've been putting them off, presuming that it was coming in another release, but chatted with Adam, and that shouldn't be the problem ...

Apparently some people are still really into the logos!  Can you help, Jack?

jcushman:
Yeah, I held off on implementing this, because I wasn't sure where the carousel would end up in the new design -- it's still just static for now.

@thisisdano -- I take it we're keeping the carousel on the home page and about page, and I should go ahead and set it to auto-update?

jcushman:
Carousel and map are now auto-updated.

PR #1215, opened 2015-10-21
Partner display improvements.

jcushman:
- The partner logo carousel is now generated from the database instead of hardcoded. Logos are automatically thumbnailed and grayscaled so we don't have to upload tiny gray versions.
- The partner map is generated from the database instead of hardcoded in javascript. This requires running some sql to load the lat-lon for existing partners: https://www.dropbox.com/s/zipzrof9kh5kxtq/registrar_locations.sql?dl=0

(todo: update the fixtures so the map and carousel show something on stage)


jcushman:
This fixes #1210

PR #1242, opened 2015-12-03
Automatic screenshots

jcushman:
This is just a little proof-of-concept for automatic re-generation of screenshots for our docs, using selenium to control Firefox. I set up a fab command to capture the first three screenshots. You can run it (from outside vagrant) with `fab dev.screenshots`.

If we like this we could set it up to capture all the other screenshots too -- I didn't want to go too far since the docs as a whole need rewriting anyway.


phillipsm:
so cool.

PR #1257, opened 2015-12-04
Switch Heroku DB for stage.

jcushman:
This should fix the timezone issue that was messing up the monthly link count on perma-stage -- I switched to a Heroku DB provider that has timezones configured.

### DIFF
--- a/app.json
+++ b/app.json
@@ -78,7 +78,7 @@
     }
   },
   "addons":[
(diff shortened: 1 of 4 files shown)


jcushman:
Pulling this in, since it's already running on stage and I don't want it to revert if something else gets pulled.

PR #1263, opened 2015-12-08
Deleted unnecessary FAQ template

### LINKED ISSUE #1258: We have two FAQ pages. We likely don't need both (?).
Remove one?

abziegler:
FAQ is now part of docs. We don't need this one. Closes #1258


jcushman:
We also have to remove a line from urls.py that depends on this template to get the tests to pass. Closing this and sending that separately to keep things simple.

PR #1298, opened 2015-12-14
Added realtime counts for users, orgs, libs, and links to the faq page.

### LINKED ISSUE #1282: Landing page stats need to be connected to real data

phillipsm:
We need to rework the stats, but this will hopefully hold us over.

Fixes #1282


jcushman:
As far as I can tell this all made it in already.

PR #1305, opened 2015-12-18
Link foldering

anastasia:
I just realized I think I meant to call this branch folder-linking. Oh well, link-foldering it is.

This pull-request links the tree folder view with the folder dropdown. By locally storing the tree node and the folder attributes (organization ID and folder ID), we can sync both parts of the app.

### DIFF
--- a/perma_web/static/js/links-list.js
+++ b/perma_web/static/js/links-list.js
@@ -215,7 +215,7 @@ $(function() {
       
       var stringPath = path.join(" &gt; ");
       if (!node.data.organization_id) {
-        stringPath += "<span class='links-remaining'>" + links_remaining + "<span></a></li>"
+        stringPath += "<span class='links-remaining'>" + links_remaining + "<span></a></li>";
       }
 
       $('#organization_select_form').find('.dropdown-toggle').html(stringPath);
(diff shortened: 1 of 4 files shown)


perma_web/static/js/links-list.js
```diff
@@ -158,6 +168,56 @@ $(function() {
...
+
+    function updateLocalStorage(node) {
+      var folder = getFolderByNode(node)
+      localStorage.setItem("perma_selected_node", JSON.stringify(node));
+      localStorage.setItem("perma_selected_folder", JSON.stringify(folder));
+    }
+
+    function updatePathWithSelected(node) {
+      var nodePath = folderTree.get_path(node).join(" &gt; ");
+      var myLinks = nodePath.includes("My Links");
```
anastasia: I'm assuming that there are no other folders called My Links. Faulty assumption, but might work in the meantime?

jcushman:
I'm looking at this now -- great, fast work! Will add a few specific comments.

perma_web/static/js/links-list.js
jcushman: Good note -- I like the idea of something more bullet-proof. Won't node.data.organization_id be null if we're a My Links subfolder?

anastasia: Yes! 
I think I just pushed a dumb test fix that will actually be solved better by your line of thought. Is My Links the only place where organization.id can be null?

jcushman:
Great work -- pulled these changes in so they can go up on stage for people to look at. Still one or two things to iron out.

PR #1316, opened 2015-12-22
Cleanup

anastasia:
Just deals with static/js and static/css and static/fonts, for now. Beginning of hopefully much wider cleanup.

/static
   * /js
      * actual perma script files like links-list.js, create.js, etc
      * /helpers
         * polyfills and such. map helpers, handlebar helpers, etc.
   * /vendors
       * folders like /bootstrap that contain all bootstrap css and js 
       * /jquery contains all jquery plugins and jquery itself
   * /fonts
      * contains all font files, font-awesome was moved here as well.


jcushman:
Sweet!

PR #1318, opened 2015-12-23
Add Firefox and Chrome capture support behind feature flag.

jcushman:
This allows headless captures to be done with Firefox or Chrome instead of PhantomJS.

Currently experimental -- e.g. Chrome screenshots only show the visible part of the screen. By default captures are still done with PhantomJS.


jcushman:
@phillipsm: the issue caught by the tests is fixed now.

PR #1339, opened 2016-02-01
Temporarily disable org delete.

jcushman:


phillipsm:
It looks like return 404 is throwing off a test. It's looking for a 200. Maybe we ignore the view in the tests?

PR #1340, opened 2016-02-01
add - making user choose a folder if none is set in their preferences

anastasia:
@phillipsm @jcushman don't know if "Please select a folder" is the right wording there?


jcushman:
This looks great -- it's going to help a ton!

One issue -- this currently doesn't handle the case where a user has no orgs and hasn't logged in and clicked a folder before. I think in that case we want to select My Links by default -- otherwise "Create Perma Link" is greyed out for no reason. This is also causing functional_tests to fail.

PR #1351, opened 2016-02-04
fix - top margin

anastasia:
capturing header height using js, sizing iframe and its wrappers using css

@jcushman: flexbox can't deal with something (like our header) being position:fixed. The only way that I could see to handle this situation, where a header is not only fixed in place but also dynamically sized, is through js on window size event (unfortunately).
However, we're no longer sizing the iframe through javascript! So that's sort of something, right?


jcushman:
I'm happy with whatever works! I think the idea with flexbox would be that the top header wouldn't be position:fixed, but just regular position. The content div would then be the height of the remaining window with overflow:scroll.

PR #1403, opened 2016-03-07
Internet Archive

anastasia:
- checking the last 48-24 hour old perma links every hour
- uploading to internet archive if they aren't set to private
- removing as much information as possible from internet archive if they are set to private and are older than 24 hours


perma_web/api/resources.py
```diff
@@ -451,17 +453,35 @@ def obj_create(self, bundle, **kwargs):
                 content_type='image/png',
             ).save()
 
-            # kick off capture task
-            run_task(proxy_capture.s(link.guid, bundle.request.META.get('HTTP_USER_AGENT', '')))
+            try:
+                run_task(proxy_capture.s(link.guid, bundle.request.META.get('HTTP_USER_AGENT', '')))
+            except Exception as e:
+                print "Getting error:",e
```
jcushman: Is this just left over from debugging, or is there something we want to do around exception handling here?

perma_web/api/resources.py
```diff
@@ -451,17 +453,35 @@ def obj_create(self, bundle, **kwargs):
...
         if bundle.data.get('folder', None):
             bundle.obj.move_to_folder_for_user(bundle.data['folder'], bundle.request.user)
+
+        if 'is_private' in bundle.data:
+            today = datetime.today()
+            newest_date = today - timedelta(days=1)
+            newest_date = timezone.make_aware(newest_date)
+            if bundle.obj.creation_timestamp < newest_date:
+                going_private = bundle.data.get("is_private")
+                print "this should not print"
```
jcushman: Debugging?

anastasia: Aaahhh debugging!

perma_web/api/resources.py
```diff
@@ -451,17 +453,35 @@ def obj_create(self, bundle, **kwargs):
...
         bundle = super(LinkResource, self).obj_update(bundle, skip_errors, **kwargs)
 
         if bundle.data.get('folder', None):
             bundle.obj.move_to_folder_for_user(bundle.data['folder'], bundle.request.user)
+
+        if 'is_private' in bundle.data:
+            today = datetime.today()
+            newest_date = today - timedelta(days=1)
+            newest_date = timezone.make_aware(newest_date)
+            if bundle.obj.creation_timestamp < newest_date:
```
jcushman: I think the previous four lines can be done as `if bundle.obj.archive_timestamp() < timezone.now()`, which also solves the problem that this is just comparing date rather than time. See `LinkUser.can_delete()` for an example.

Actually to make this even more readable and DRY, could we move that comparison to a `Link.is_archive_eligible()` function (or some similar name that makes more sense), and call it both from this line and from `LinkUser.can_delete()`?

perma_web/api/resources.py
```diff
@@ -478,6 +498,11 @@ def obj_delete(self, bundle, **kwargs):
 
         bundle.obj.safe_delete()
         bundle.obj.save()
+        try:
+            if bundle.obj.uploaded_to_internet_archive:
+                run_task(delete_from_internet_archive.s(link_guid=bundle.obj.guid))
+        except Exception as e:
+            print "getting error in delete:",e
```
jcushman: Debugging?

perma_web/perma/tasks.py
```diff
@@ -592,45 +593,115 @@ def update_stats():
         current_week.save()
 
 
-@shared_task(
-    bind=True,
-    default_retry_delay=10*60) # 10 minute delay between retries
+@shared_task(bind=True)
+def delete_from_internet_archive(self, link_guid):
+    identifier = settings.INTERNET_ARCHIVE_IDENTIFIER_PREFIX + link_guid
```
jcushman: Does this need `if not settings.UPLOAD_TO_INTERNET_ARCHIVE: return` at the top like the other function?

perma_web/perma/tasks.py
```diff
@@ -592,45 +593,115 @@ def update_stats():
...
+
+@shared_task()
+def upload_all_to_internet_archive():
+    # find all links created 48-24 hours ago
+    # include timezone
+    today      = datetime.today()
+    start_date = today - timedelta(days=2)
+    end_date   = today - timedelta(days=1)
+    start_date = timezone.make_aware(start_date)
+    end_date   = timezone.make_aware(end_date)
```
jcushman: This can be avoided by starting with `today = timezone.now()`

anastasia:
@jcushman tested that background queue works:
running celery and background queues in two terminal windows logging in two separate files yields two different sets of logs:
```
$ cat tmp-celery-task.txt 
   perma.tasks.update_stats[1da6712...]: running update_stats
$ cat tmp-background-task.txt
   perma.tasks.upload_to_internet_archive[4fa03888...]: running upload_to_internet_archive
```

Issue #458, opened 2014-02-28
Pretty maintenance page.


jcushman:
I'm not sure what exactly is involved from the technical end, but we should put together a nicely styled page to switch to when we're upgrading the server -- our current one looks like something might have gone wrong.

jcushman:
This is a design thing -- our static maintenance page lives in static/maintenance/maintenance.html. It would be nice if it looked like everything was happy and under control.

PR #1407, opened 2016-03-08
New maintenance page

anastasia:
finishes #458


jcushman:
Can we make this a self-contained HTML file? Like, same look, but move the CSS inline, and copy the gif contents into a `data:` url. It looks tricky to get nginx to serve multiple files in maintenance mode ...

PR #1440, opened 2016-03-29
Js window.onerror handling

anastasia:
saving any unexpected errors to a table, posting to /manage/errors


perma_web/perma/models.py
```diff
@@ -944,3 +944,15 @@ def is_revisit(self):
...
+    user_agent = models.TextField()
+    stack = models.TextField()
+    name = models.TextField()
+    message = models.TextField()
+    custom_message = models.TextField()
+    user = models.ForeignKey(LinkUser, null=True, blank=True)
+    created_at = models.DateTimeField(default=timezone.now)
+
+    resolved = models.BooleanField(default=False)
+    resolved_by_user = models.CharField(null=True, max_length=767)
```
jcushman: This should be a `models.ForeignKey` rather than a CharField. (Then you'll need a related_name for the two ForeignKeys to distinguish them -- something like `related_name="errors_triggered"` for `user`, and `related_name="errors_resolved"` for `resolved_by_user`.)

jcushman:
Looks like a couple of tests are actual fails -- you can't save the user when `request.user.is_anonymous`.

anastasia:
blurg! On it.

anastasia:
How?! I added a test!

perma_web/perma/views/common.py
```diff
@@ -49,14 +49,14 @@ def get_context_data(self, **kwargs):
                     context[key] = value
         return context
 
-
+@csrf_exempt
```
jcushman: Should this be on `post_new` instead of `landing`?

Issue #1423, opened 2016-03-23
Edited folder name doesn't update in dropdown


abziegler:
I edited the `.selected` folder name, but the old name remained in drop-down
![screen shot 2016-03-23 at 9 49 58 am](https://cloud.githubusercontent.com/assets/2140548/13987192/bc7a0cd0-f0dc-11e5-8282-7aed2e88ee8f.png)

anastasia:
this is an existing bug (maybe bump?)

jcushman:
Yep, bumped to 0.61.

jcushman:
Deployed!

PR #1448, opened 2016-04-05
Updating dropdown path on folder name edit

anastasia:
finishes #1423 

+ added WARC_HOST to settings_dev since we're requiring it now
+ added more margin to show error message for error warc

### DIFF
--- a/perma_web/static/js/folder-tree.module.js
+++ b/perma_web/static/js/folder-tree.module.js
@@ -187,7 +187,7 @@ FolderTreeModule.domTreeInit = function () {
                 .done(function () {
                   self.allowedEventsCount++;
                   self.folderTree.rename_node(node, newName);
-                  var folderPath = self.folderTree.get_path(node)
+                  var folderPath = self.folderTree.get_path(node);
                   var folderId = node.data.folder_id;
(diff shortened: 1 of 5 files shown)


perma_web/static/js/folder-tree.module.js
```diff
@@ -103,23 +103,25 @@ FolderTreeModule.editNodeName = function (node) {
...
-    var folderId = node.data.folder_id;
-    var orgId = node.data.organization_id;
-    data = JSON.stringify({path: folderPath, orgId:orgId, folderId:folderId});
+    var data = {};
+    data.folderId = node.data.folder_id;
+    data.orgId = node.data.organization_id;
+    data.path = folderPath = this.folderTree.get_path(node);
   }
+  return data;
+}
```
jcushman: I'm confused what this function returns if `node.data` is false. Maybe `var data = {}` wants to be outside the `if`?

anastasia: data would return null and handleSelectionChange in create.module.js wouldn't attempt to get orgId and path vars from data. But you're right, it's super convoluted and can be restructured.

PR #1473, opened 2016-04-21
Fix retry on getting browser title

trifle:
Hi,
looking through your tasks.py I noticed that there's a bug that prevents additional attempts to get the title:

            def get_title():
                if browser.title:
                    save_fields(link, submitted_title=browser.title)

This grabs browser.title, which may contain the correct title or '' (which evaluates to False) - but it never raises an Exception! [1].

Hence, the subsequent `repeat_while_exception(get_title, timeout=10)` will never get a chance to repeat.

This tiny patch just raises a KeyError for empty titles, allowing for retries of up to 10 seconds.

[1] http://seleniumhq.github.io/selenium/docs/api/py/_modules/selenium/webdriver/remote/webdriver.html#WebDriver.title

### DIFF
--- a/perma_web/perma/tasks.py
+++ b/perma_web/perma/tasks.py
@@ -393,7 +393,7 @@ def proxy_capture(self, link_guid, user_agent=''):
                     save_fields(link, submitted_title=browser.title)
                 else:
                     title_element = browser.find_element_by_tag_name("title")
-                    save_fields(link, submitted_title=title_element.text)
+                    save_fields(link, submitted_title=title_element.get_attribute("text"))
             repeat_while_exception(get_title, timeout=10)
 
             # check meta tags
--- a/perma_web/perma/tasks.py
+++ b/perma_web/perma/tasks.py
@@ -155,14 +155,16 @@ def add_thread(thread_list, func):
     new_thread.start()
     thread_list.append(new_thread)
 
-def repeat_while_exception(func, exception=Exception, timeout=10, sleep_time=.1):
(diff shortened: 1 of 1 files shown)


jcushman:
Thanks! Great catch.

It looks like your patch is currently failing our tests because not every page we want to capture has a title -- we still want a capture to succeed even if it has `<title></title>` or a weird title that Selenium can't read. We also don't want to delay captures by 10 seconds in the case of a blank title.

Actually what I think we want is three cases:

(1) If the page does have a title, save it.
(2) If the page has a blank title, immediately continue.
(3) If selenium just doesn't know the title yet, wait up to 10 seconds for the browser to sort itself out before giving up and going on with the rest of the capture.

Maybe the way to do that is to use something like `browser.find_elements_by_tag_name('title')` instead of `browser.title`, and then try to detect whether an error indicates a timeout that might fix itself or a not-found that is permanent?

trifle:
Right, those are use cases I didn't think about.
We only check for titles in text/html, so your proposal should work (see updated patch):

1. get browser.text
2. if False (= empty string): use title element
3. if title element is present (even empty), use that as title
4. if there is no <title>, a selenium.common.exceptions.NoSuchElementException is raised
5. after 10 seconds, give up and continue without title.

The timeout only affects worst cases and normally shouldn't delay the capture.

trifle:
Okay, so the tests are failing because we can't raise even after the timeout.
I've added an argument to `repeat_while_exception` to swallow the final Exception if explicitly desired.

jcushman:
Elegant solution -- thanks!

PR #1482, opened 2016-05-04
Upgrade fabric

jcushman:
Fixes test error.


jcushman:
Merged this right in since it fixes a test error on our develop branch. Apparently one of Fabric's own dependencies has a new release (Paramiko v.2.0), which caused our version of Fabric to stop working.

PR #1483, opened 2016-05-04
Pinned requirements

jcushman:
This uses requirements.in and requirements.txt files with the pip-compile tool to make sure that all of our sub-dependencies are pinned to specific versions.

pip-compile required me to resolve dependency conflicts, which meant updating the `internetarchive` and `requests` packages to their latest versions.


anastasia:
This one's for @bjohnson747

PR #1492, opened 2016-05-10
User forms refactor

jcushman:
- Drop a lot of forms we were no longer using, or that were redundant, from perma/forms.py. 
- Simplify and add comments for remaining forms.
- For the organization_user_add_user view that adds an existing user to an organization, make sure that a user can only be added if they are not already a registrar or admin; otherwise show an error message.
- Add tests for the organization_user_add_user view.

Commits are a suggested order for review.


jcushman:
(hold on this one, I think it will make sense to fold in the stuff I'm currently working on)

jcushman:
OK, additional changes added. This and my other PR got pretty large -- maybe easier to review in person?

PR #1498, opened 2016-05-12
Internet archive metadata fix

anastasia:
before merging to production, make sure to run: 
```
mysql -e "select guid from perma_link where uploaded_to_internet_archive=True" -uperma -pperma perma > links_without_metadata.txt
```

### DIFF
--- a/perma_web/perma/tasks.py
+++ b/perma_web/perma/tasks.py
@@ -682,25 +682,30 @@ def upload_to_internet_archive(self, link_guid):
     if link.organization:
         metadata["sponsor"] = "%s - %s" % (link.organization, link.organization.registrar)
 
+
     identifier = settings.INTERNET_ARCHIVE_IDENTIFIER_PREFIX + link_guid
-    with default_storage.open(link.warc_storage_file(), 'rb') as warc_file:
-        success = internetarchive.upload(
-                        identifier,
-                        warc_file,
-                        metadata=metadata,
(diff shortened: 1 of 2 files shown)


perma_web/api/resources.py
```diff
@@ -481,11 +481,11 @@ def obj_update(self, bundle, skip_errors=False, **kwargs):
                 going_private = bundle.data.get("is_private")
                 # if link was private but has been marked public
                 if is_private and not going_private:
-                    run_task(upload_to_internet_archive.s(link_guid=bundle.obj.guid))
+                    run_task(upload_to_internet_archive.s(link=bundle.obj))
```
jcushman: Passing `link` instead of `link_guid` won't work for production, where RUN_TASKS_ASYNC=True. This line raises `EncodeError: <Link: 2HWD-BUB9> is not JSON serializable`. To pass tasks to background celery the parameters have to be JSON serializable.

(It would be neat if unit tests found this issue -- not sure if there's a good way to do that though.)

(Even if we used the `pickle` serializer which lets you pass Django objects directly, doing it this way raises the possibility that the link object will be updated before the job is actually run, and the task won't see the updates. It's hard to figure out whether that's a problem in a given case -- simpler just to fetch an up-to-date copy when the task starts.)

anastasia: Oh, oops! Thanks for the note, will fix.

jcushman:
Ran the SQL -- output is stored in /perma/home/perma/links_without_metadata.txt

Issue #1444, opened 2016-04-04
Capture fail message not displaying properly on small screens


abziegler:
Tried this URL: http://www.mondaq.com/unitedstates/x/435110/Financial+Services/Madden+v+Midland+Funding+LLC

Got capture fail message, which itself didn't display right at first (see below), but now does:

![screen shot 2016-04-04 at 11 55 42 am](https://cloud.githubusercontent.com/assets/2140548/14254382/66373d22-fa5d-11e5-8cd4-3f71d0c3dfa0.png)

abziegler:
May be related to #1439

anastasia:
May be resolution related (here's mine):
<img width="1280" alt="screen shot 2016-04-04 at 12 08 22 pm" src="https://cloud.githubusercontent.com/assets/1494102/14254504/01b61ff2-fa5e-11e5-8dd8-cba7aacb0cd6.png">

jcushman:
Creating separate bug for the capture problem here (#1445), and removing 'capture error' label from this bug.

anastasia:
https://perma.cc/U2Z7-TEHV
Added extra padding to error message, should be mostly fixed but need to double check if the code is actually solid long-term.

jcushman:
Per @anastasia, remaining work is to get the message showing on small screens.

Issue #1452, opened 2016-04-07
Installation instructions out of date


anastasia:
In https://github.com/harvard-lil/perma/blob/develop/install.md
`python manage.py syncdb --noinput`
via https://docs.djangoproject.com/en/1.8/ref/django-admin/#syncdb
Deprecated since version 1.7:
This command has been deprecated in favor of the migrate command, which performs both the old behavior as well as executing migrations.

PR #1515, opened 2016-05-18
fixes capture error sizing issue  + install instructions

anastasia:
fix - finishes #1444


anastasia:
finishes #1452

anastasia:
going to break out into two easy pull requests

PR #1522, opened 2016-05-19
internet archive metadata + status fixes

anastasia:
- we're now storing ia upload status on link models
- retrying to upload if warc exists but ia upload failed for some reason
- not retrying if warc does not exist (marking link as failed permanently)
- if archive WAS uploaded but later was deleted (made private, for instance), only the file was getting reuploaded. the metadata was not. this is now fixed — we're checking if item exists and then applying "modify_metadata" method if it's been "Removed".
I think we might need to think a little more about how we're dealing with IA. If a warc has failed to be created, do we want the metadata stored somewhere? What are all the edge cases, do we do a good job in keeping perma archives and internet archive archives in sync?

### DIFF
--- a/perma_web/perma/migrations/0006_add_internetarchive_status.py
+++ b/perma_web/perma/migrations/0006_add_internetarchive_status.py
@@ -5,39 +5,24 @@ from django.db import migrations, models
 
 def update_upload_to_ia_field(apps, schema_editor):
     Link = apps.get_model('perma', 'Link')
-    for link in Link.objects.all():
-        if link.uploaded_to_internet_archive:
-            link.internet_archive_upload_status = 'completed'
-        elif not link.uploaded_to_internet_archive:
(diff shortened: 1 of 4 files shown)


perma_web/perma/migrations/0006_add_internetarchive_status.py
```diff
@@ -0,0 +1,71 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+def update_upload_to_ia_field(apps, schema_editor):
+    Link = apps.get_model('perma', 'Link')
```
jcushman: These will deploy a *lot* faster on prod if we do them as something like, `Link.objects.filter(uploaded_to_internet_archive=True).update(internet_archive_upload_status='completed')`

PR #1537, opened 2016-05-26
Will activate New Relic on deploy to production

jayed:
###             In that case, then *this* commit should be the final state for all of the New Relic stuff

+ added and customized newrelic.ini file
  + chaned app name to Perma
  + chaged logging directory to perma's logs
  + turned off monitoring for a staging environment

+ import and activcate agent on application start
  + just a few lines in the wsgi.py file


jcushman:
Thanks! I adapted this to work with staging and other environments, in a new pull request ( #1541 ). Closing this one.

PR #1540, opened 2016-05-26
Patch archive with your own version

anastasia:


perma_web/api/authorizations.py
```diff
@@ -196,6 +196,9 @@ def update_detail(self, object_list, bundle):
         if not self.can_access(bundle.request.user, bundle.obj):
             raise Unauthorized()
 
+        if bundle.data.get('file') and bundle.obj.is_archive_eligible():
+            raise Unauthorized()
+
         return True
```
anastasia: if we're changing file but the archive is older than 24 hours, unauthorize

perma_web/api/resources.py
```diff
@@ -508,25 +520,43 @@ def obj_create(self, bundle, **kwargs):
         return bundle
 
     def obj_update(self, bundle, skip_errors=False, **kwargs):
-        is_private = bundle.obj.is_private
-        bundle = super(LinkResource, self).obj_update(bundle, skip_errors, **kwargs)
+        uploaded_file = bundle.data.get('file')
+        if uploaded_file and bundle.request.method == 'PATCH':
+            if kwargs['request']:
+                del kwargs['request']
```
anastasia: request ends up in kwargs because of the hacky patch_detail up above. If it exists in kwargs, delete! Otherwise everything will fail.

PR #1541, opened 2016-05-26
Newrelic integration

jcushman:
This pulls out our Heroku-specific New Relic support, and adds general New Relic support that can be enabled with an environment variable. Based on Jay's pull request but adapted to play well with stage/dev as well as prod.

### DIFF
--- a/perma_web/perma/wsgi.py
+++ b/perma_web/perma/wsgi.py
@@ -13,7 +13,9 @@ import os
 use_newrelic = os.environ.get("USE_NEW_RELIC", False)
 if use_newrelic:
     import newrelic.agent
-    newrelic.agent.initialize(os.path.join(os.path.dirname(__file__), '../../services/newrelic/newrelic.ini'))
+    newrelic_config_file = os.environ.get('NEW_RELIC_CONFIG_FILE',
+                                          os.path.join(os.path.dirname(__file__), '../../services/newrelic/newrelic.ini'))
+    newrelic.agent.initialize(newrelic_config_file)
 
 # env setup
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "perma.settings")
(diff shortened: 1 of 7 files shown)


jcushman:
Pulling this in to avoid issues with our Heroku buildpack, which was updated at the same time.

services/newrelic/newrelic.ini
```diff
@@ -191,14 +186,14 @@ thread_profiler.enabled = true
...
 [newrelic:development]
-monitor_mode = false
+monitor_mode = true
 
 [newrelic:test]
-monitor_mode = false
+monitor_mode = true
 
 [newrelic:staging]
 app_name = Python Application (Staging)
```
jayed: This should be changed to "Perma (Staging)". Alternatively you can use the environment variable NEW_RELIC_APP_NAME

perma_web/perma/wsgi.py
```diff
@@ -9,36 +9,36 @@
 """
 import os
 
-import newrelic
-from werkzeug.wsgi import DispatcherMiddleware
-
-newrelic.agent.initialize('/srv/www/perma/services/newrelic/newrelic.ini')
+# Newrelic setup
+use_newrelic = os.environ.get("USE_NEW_RELIC", False)
```
jayed: Instead of using a custom environment variable, your should use the standard NEW_RELIC_ENVIRONMENT variable  to set the environment to one of the environments specified at the bottom of the file and then use monitor_mode=true/false to control collection.

services/newrelic/newrelic.ini
```diff
@@ -24,11 +24,6 @@
 
 [newrelic]
 
-# You must specify the license key associated with your New
-# Relic account. This key binds the Python Agent's data to your
-# account in the New Relic service.
-license_key = 889221e414af81d4877cc57b9bfd9413ac122dfe
```
jayed: Removing the license key is awesome. But I can't tell from the commit message what you replaced it with? Was it the env variable NEW_RELIC_LICENSE_KEY? And, if so, how are you currently distributing it out?

PR #1545, opened 2016-05-31
Develop.1402

phillipsm:


perma_web/perma/tests/test_link_count_caching.py
```diff
@@ -0,0 +1,61 @@
+from perma.models import *
+
+from .utils import PermaTestCase
+
+class CaptureJobTestCase(PermaTestCase):
```
jcushman: Nitpick: this should be something like LinkCountCachingTestCase

perma_web/perma/tests/test_link_count_caching.py
```diff
@@ -0,0 +1,61 @@
...
+
+    fixtures = [
+        'fixtures/users.json',
+        'fixtures/folders.json',
+    ]
+
+    def setUp(self):
+        super(CaptureJobTestCase, self).setUp()
```
jcushman: Ditto naming nitpick

perma_web/perma/views/user_management.py
```diff
@@ -501,30 +501,33 @@ def list_users_in_group(request, group_name):
     users = users.filter(
```
jcushman: I think this whole `users = users.filter()` block is now unnecessary -- true?

PR #1555, opened 2016-05-31
Rename members to users

### LINKED ISSUE #952: Clean up test accounts
Our test account login names are a little ragged. Sometimes we use *_user@example.com and sometimes we use *_member@example.com.

Test accounts should follow the same pattern, and maybe a shorter one? How about the following?

user@example.com
org_user@example.com
registrar_user@example.com
registyr_user@example.com

phillipsm:
Fixes #952

Rename test accounts from *_member to *_user
Rename code references from member to user
Migration to update existing loaded user accounts names

**notes, I think we may have created more test accounts in some envs and we should update those in a one-off manner**


jcushman:
LGTM. Let's pull in after we deploy tomorrow.

#dismembered

jcushman:
Actually @phillipsm, what do you think of renaming test_registry_user -> test_admin_user while you're at it? I've been phasing out 'registry' since it's confusing ...

phillipsm:
Yes! That's smart, @jcushman. I'll do that.

jcushman:
sweet this is perfect

PR #1579, opened 2016-06-07
Upgrade to Django 1.9

jcushman:
Also upgrades tastypie and ratelimit to latest versions for compatibility.


anastasia:
sorry — @jcushman — merged out of order and now we have conflicts

PR #1598, opened 2016-06-13
Appending hash to warc_url name in file replace

### LINKED ISSUE #1556: Update link won't allow upload, then re-upload
On perma-stage, logged in as test_org_user:

1. archived example.com
2. clicked 'edit record'
3. uploaded png
4. clicked 'edit record' again
5. tried to upload different png

Did not swap out new png. Kept the first one instead.

anastasia:
Only appending hash if a file is getting replaced
Fixes #1556

### DIFF
--- a/perma_web/api/resources.py
+++ b/perma_web/api/resources.py
@@ -1,5 +1,5 @@
-import hashlib
-import time
+import re
+import random
 import json
 from django.conf import settings
 from django.conf.urls import url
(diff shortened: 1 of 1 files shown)


anastasia:
Perma team: this is not really a fix, but I am not getting a content-type mismatch when re-uploading file. Let me know if it's not sufficient.

PR #1601, opened 2016-06-14
Monitor celery on staging

bensteinberg:
This is a first cut at monitoring celery with NewRelic on staging. It'll monitor celery and the Django application together; next step is to monitor them separately.


bensteinberg:
I wonder if this will need to specify the virtualenv, as described at https://discuss.newrelic.com/t/need-some-help-setting-up-new-relic-with-our-celery-workers/30181/4

PR #1616, opened 2016-06-28
New functional tests

HolaHansi:
The **motivation** behind this commit is

1) To put the helper functions outside test_all(), but within the same class, so we can reuse these helpers in other tests cases. 

2) To break up test_all() into multiple components - again to promote modularity. 

I do this because I'm working on creating a larger test suite and hence, modularity is going to be very useful. I'm still working on getting more of the logic into a setUpClass function, which turned out to be difficult as calls to local methods like : 

> self.server_thread.httpd.set_app(self.server_thread.static_handler(wsgi_app)) 

found in setUp()

Is not allowed within class methods. I'm not sure how to best resolve this issue, so for now, all setup logic happens in setUp(), which explains why the tests are so slow.

### DIFF
--- a/perma_web/functional_tests/tests.py
+++ b/perma_web/functional_tests/tests.py
@@ -8,7 +8,6 @@ import re
 import datetime
 import sys
 from urlparse import urlparse
-
 from pyvirtualdisplay import Display
 from selenium import webdriver
 from selenium.common.exceptions import ElementNotVisibleException, NoSuchElementException
@@ -102,9 +101,6 @@ if USE_SAUCE:
 else:
     browsers = [{}]  # single PhantomJS test case with empty dict -- no special desired_capabilities
 
-
-## helpers
-
 def on_platforms(platforms):
     """
         Generate a version of the test case class for each platform.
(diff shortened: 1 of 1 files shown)


jcushman:
It looks like this pull currently includes the entire phantomjs binary distribution. Could you remove those files? (By rolling back rather than by an additional commit -- I don't want to blow up our git repository size.) You can move phantomjs to `perma_web/lib/` if you want to keep it local, or install it globally in vagrant using these steps: https://github.com/harvard-lil/perma/blob/develop/services/vagrant/provision.sh#L55

perma_web/functional_tests/tests.py
```diff
@@ -134,22 +132,24 @@ class FunctionalTest(BaseTestCase):
...
+                self.server_thread.host = LOCAL_SERVER_DOMAIN
 
-            self.server_url = self.live_server_url
+                self.server_url = self.live_server_url
 
-        if USE_SAUCE:
-            self.setUpSauce()
-        else:
-            self.setUpLocal()
+            if USE_SAUCE:
```
jcushman: Suggestion: this if/else block is the part that's too slow to repeat for each test, because that's where we spin up the headless browser. Maybe move this block to `setUpClass`, and just leave stuff that has to be there in `setUp`? Note that `tearDown` then has to become `tearDownClass` as well, because it's the inverse of this block.

Also, nitpick: we follow [pep8](https://www.python.org/dev/peps/pep-0008/) for python code, including 4 spaces for indent. This block became 8 spaces somehow. :)

PR #1617, opened 2016-07-01
Fix typo

thatandromeda:
The docs reference `is_organization_member` but no such property appears on the user model - presumably `is_organization_user` was intended.


anastasia:
thanks!

PR #1633, opened 2016-07-27
DO NOT MERGE add - if item exists but for some reason it's been marked as not started, mark it as completed

anastasia:


anastasia:
I'm actually not totally sure this is right. Will close out by tomorrow.

PR #1638, opened 2016-07-29
Capture assets from audio, video, embed, and object tags

### LINKED ISSUE #1635: just push play?
if we're archiving a target and it contains a video embed, should we click the play button and capture the video? should we incorporate webrecorder.io's tech as an _advanced capture_ feature?

jcushman:
This is a start to capturing audio/video assets. It just captures assets included via normal html5 audio/video tags, as well as oldschool object and embed tags.

Example page where assets are now captured: http://www.tutorialspoint.com/html5/html5_audio_video.htm

To see what this is looking for, check out perma/tests/assets/target_capture_files/test_wav_inner.html

Closes #1635


jcushman:
Tests fixed.

PR #1657, opened 2016-08-22
Small fixes

anastasia:
- removed created-links
- moved handlebars templates compilation to handlebarshelpers module


perma_web/static/js/helpers/handlebars_helpers.js
```diff
@@ -14,3 +17,29 @@ Handlebars.registerHelper ('truncatechars', function (str, len) {
...
+  });
+}
+
+HandlebarsHelpers.renderTemplate = function(templateId, args) {
+  var $this = $(templateId);
+  var name = templateId.replace('-template', '').replace(/-/g, '_');
+  templates[name] = Handlebars.compile($this.html());
+  return templates[name](args)
+}
+
```
jcushman: Since you're refactoring anyway ... :)

- Why don't we take it all the way and make `renderTemplate` the only way that templates are rendered? Then we don't need the `init` function or `renderAllTemplates`, and we don't have confusing code hanging around in the remaining places that refer to `templates` directly.
- Maybe `HandlebarsHelpers.templateCache` would be clearer than `window.templates` for that variable?
- Once we get rid of direct access to `templates`, there's no need for the `replace` stuff for `templateId`.
- As an optimization, can skip the `compile` if it's already cached.
- Maybe HandlebarsHelpers should have a sweet theme song? https://soundcloud.com/hamburgerhelper/hamburger-helper-prod-retro-spectro

perma_web/static/js/stats.js
```diff
@@ -1,7 +1,24 @@
 // House our sum data here. We should refactor this.
+window.StatsModule = window.StatsModule || {};
```
anastasia: Refraining from refactoring whole stats page for now, but new code follows the same style of other js modules

PR #1675, opened 2016-09-08
Fix private link playback in IE and Safari

jcushman:


jcushman:
Taking another pass ...

PR #1691, opened 2016-09-17
Simplify capture job queue

jcushman:
This implements @rebeccacremona's idea of ordering CaptureJobs at the time they're created. Seems a lot simpler!

The algorithm is this: suppose User X submits `new_job`. We need to assign a `new_job.order` float value so it gets processed in the right order relative to all existing pending jobs.

`new_job` should definitely be processed after all previously submitted User X jobs, so look just at the pending jobs with higher `order` values than the highest previous job belonging to User X. 

The "fair" point to process `new_job` is right before processing a *second* job by someone else in the queue. So we walk through those other-user jobs in order, until we find a second job belonging to a user who already got one. `new_job` gets an order value halfway between that second job and the one previous.

If no second job belonging to another user is found, `new_job` gets an order value 1 higher than the highest previous.

For example, suppose we add three jobs from A, and then three jobs from B, and then three jobs from C. Here are the order values after each addition according to this algorithm:

A=1
A=1 **A=2**
A=1 A=2 **A=3**
A=1 **B=1.5** A=2 A=3
A=1 B=1.5 A=2 **B=2.5** A=3
A=1 B=1.5 A=2 B=2.5 A=3 **B=4**
A=1 B=1.5 **C=1.75** A=2 B=2.5 A=3 B=4
A=1 B=1.5 C=1.75 A=2 B=2.5 **C=2.75** A=3 B=4
A=1 B=1.5 C=1.75 A=2 B=2.5 C=2.75 A=3 B=4 **C=5**

Thus we end up with the round-robin order ABCABCABC as expected.

Limitations:
* Since we're trying to avoid locking, this doesn't claim a lock on the pending rows. So in a race condition it's possible for two new entries to come in with the same order value. Ties are broken by primary key, which roughly maps to timing anyway, so I think this only creates trivial unfairness between jobs that arrived at basically the same time.
* We can't entirely avoid locking -- there's still a `select_for_update` in `get_next_job`. When a capture task wants to claim a job to process, it has to lock that row long enough to change status from 'pending' to 'in_progress'. That seems like a simple use of locking that's unlikely to cause deadlock, though.

Does this seem good? Any ways to make it more clear what's going on?


bensteinberg:
Does the locking still present in `get_next_job` work when the table engine is MyISAM?

jcushman:
Good question -- thinking about it more, we can avoid pessimistic locking in `get_next_job` and use optimistic locking, which doesn't care about the table engine. In optimistic locking we just try the update, and if no rows were updated that means someone else got to the job first and we try again for another one. I pushed an update that does that.

Note that any sort of locking only occurs in `get_next_job(reserve=True)`, which is only called by celery workers, so there's no possibility of deadlock in web threads regardless.

perma_web/perma/models.py
```diff
@@ -983,118 +984,46 @@ class CaptureJob(models.Model):
...
+        # If this job does not have an order yet (just created),
+        # examine all pending jobs to place this one in a fair position in the queue.
+        # "Fair" means round robin: this job will be processed after every other job submitted by this user,
+        # and then after every other user waiting in line has had at least one job done.
+        if not self.order:
+
+            # get all pending jobs, in reverse priority order
+            pending_jobs = CaptureJob.objects.filter(status='pending', human=self.human).order_by('-order').select_related('link')
+            # narrow down to just the jobs that come *after* the most recent job submitted by this user
+            pending_jobs = list(itertools.takewhile(lambda x: x.link.created_by_id != self.link.created_by_id, pending_jobs))
```
rebeccacremona: Use of itertools and a lambda function all at once! Love it.

PR #1698, opened 2016-09-21
Rework settings/organizations page to make user affiliations clearer

rebeccacremona:
At last week's Perma meeting, we decided to title this page "affiliations" rather than "organizations", since "organization" is a term of art in Permaland, making it inappropriate to refer to a registrar as an organization. This request makes that change both textually (menus, headings, title-tags) and in the code (file names, function names). For consistency, it also changes the route in question from /settings/organizations to settings/affiliations. (Neither route appears anywhere in the user guide; it's unlikely anybody has that url bookmarked...)

The page was also reworked to display users' orgs grouped by registrar. The look and feel of the page as seen by registrars was also tweaked.

/settings/affiliations as seen by an org user (with three affiliations, two with one registrar and one with another):

org-user

...as seen by an approved registrar:

approved-registrar

...as seen by a pending registrar:

pending-registrar


perma_web/perma/templates/user_management/settings-affiliations.html
```diff
@@ -0,0 +1,86 @@
...
+        </h4>
+        {% for org in orgs %}
+          <div class="settings-block">
+            <p>
+              {{ org }}
+              <span class="_hideMobile">
+                {% if org.default_to_private %}
+                  (Links are <strong>Private</strong> by default)
+                {% else %}
+                  (Links are <strong>Public</strong> by default)
```
jcushman: Small suggestion -- let's not include `(Links are Public by default)`, only Private, because most users will *only* have public orgs and never think about it.

perma_web/perma/views/user_management.py
```diff
@@ -985,6 +985,9 @@ def settings_organizations(request):
     if pending_registrar:
         messages.add_message(request, messages.INFO, "Thank you for requesting an account for your library. Perma.cc will review your request as soon as possible.")
 
+    organizations = request.user.organizations.all().order_by('registrar')
+    orgs_by_registrar = {registrar : [org for org in orgs] for registrar, orgs in itertools.groupby(organizations, lambda x: x.registrar)}
```
jcushman: Total nitpick -- seems like this works as just `orgs_by_registrar = dict(itertools.groupby(organizations, lambda x: x.registrar))`

perma_web/perma/templates/user_management/settings-affiliations.html
rebeccacremona: Thanks, great idea. I had observed they looked too similar in any case, and was considering adding styling to make "Private" more distinct.... your thought sounds exactly right.

perma_web/perma/views/user_management.py
rebeccacremona: Aha. Much cleaner. I think I was initially doing something else with the data, hence the contortions, but as is, absolutely, simple call to dict is the way to go!

rebeccacremona: Actually: nope. That does not work. Happy to show you on my machine if you like: I've altered a test user to have lots of affiliations, which is where problems show up.

rebeccacremona: The issue with using dict() as opposed to building the dict yourself using the dict and list comprehensions is not the key, which is the registrar object in both cases, but in the value. When using dict(), the value is an <itertools._grouper object>, which as far as I can tell, the Django templating language can't iterate through correctly.

jcushman: Oops, yes, I didn't try with enough affiliations. This looks good!

Issue #1681, opened 2016-09-12
IE support


rebeccacremona:
The FAQ says that we support IE 8+
```
Which web browsers does Perma.cc support?
Perma.cc works on Chrome, Firefox, Safari and IE8+.
```
I don't know if we hold ourselves to that, but, since Microsoft has dropped support for IE 8, 9, and 10, maybe we should as well. And do we test with Microsoft Edge? Would that free us up in any way? (some possibilities: faster cross-browser testing, slicker/cleaner CSS, lighter weight version of JQuery? hmmm. other things?)

jcushman:
Last time we discussed this, we decided to bump support to IE10+ based on usage logs. Looks like we missed the FAQ -- could you update it?

rebeccacremona:
Sure.

rebeccacremona:
https://github.com/harvard-lil/perma/pull/1700

PR #1700, opened 2016-09-22
FAQ tweaks

rebeccacremona:
Only claim support for IE 10+ per #1681. Also tweak a few things I noticed while the file was open.


jcushman:
> (let's hold on to "accessible" for later after we fix some things)

♥

Issue #336, opened 2013-11-20
Create a calendar view


phillipsm:
If multiple users have archived the same URL, it would be useful to see different versions over time. A calendar view could work, like how the Wayback Machine does it.
![cal-view](https://f.cloud.github.com/assets/196089/1586560/576279e8-522a-11e3-9aa9-9774348e504b.png)

anastasia:
Basic structure is done — calendar exists, displaying only public links, showing search results for fuzzy urls. Need feedback on what we want our calendar UX to look like

anastasia:
This is done? I think?

rebeccacremona:
See https://github.com/harvard-lil/perma/issues/1728

jcushman:
So done.

PR #1704, opened 2016-09-26
Calendar search

### LINKED ISSUE #1680: Old search results show up in new search
If you do a search then scroll, your results repeat forever
repeated search results

anastasia:
screen shot 2016-09-26 at 4 29 28 pm

finishes #336 
fixes #1680


rebeccacremona:
Very attractive-looking calendar! A couple small suggestions: 

- If you are using your keyboard to navigate, it is pretty difficult to see which date on the calendar has keyboard focus. Might consider adding an additional :focus style for those elements
- The prev/next links can't be accesses via keyboard, because, though they are wrapped in `<a>`, there's no href attribute. Using a `button` tag would take care of that, and would let people press the button with "space" instead of "enter" (=expected behavior for buttons). Checking into whether this library permits that markup to be customized.....
- The search input field and the search button should have screenreader-only labels
- The title of the page still says "Stats"

anastasia:
Added aria-labels only, for now. And fixed "Stats" title, of course!

rebeccacremona:
We're going to keep talking about the :focus styling on the calendar, but don't want to hold this up for that!

PR #1707, opened 2016-09-27
NEW: prefix subject line with [perma-contact] for easier parsing and …

rebeccacremona:
…filtering


rebeccacremona:
FWIW, this is why I don't fully trust coveralls. These changes didn't decrease test coverage at all. I made a string a little longer, is all......

PR #1715, opened 2016-09-28
FIX: Add default referer so that hidden uses of send_admin_email don'…

rebeccacremona:
…t break.


bensteinberg:
Does the order of arguments to the function matter here?

rebeccacremona:
Doh. Of course. Fixing.......

PR #1759, opened 2016-10-18
Catch urlerror in capture task

jcushman:
Fixes capture failure recently reported by user


rebeccacremona:
Small suggestion: how about a comment to explain what this error is, and why it's okay to "pass" if an exception occurs?

jcushman:
Good idea -- added some comments.

jcushman:
Good idea -- added a couple of comments.

PR #1778, opened 2016-10-24
Nudge org users to use org folders

rebeccacremona:
- dropdown color changes
- "unlimited" badge 
- "My Links" to "Personal Links"
- warning message for org users nearing their personal limits
dropdown
warning

N.B. change to perma_web/perma/templates/docs/developer/index.html
Do we anticipate anybody using our API relies on a folder being called "My Links"? Surely not.

### DIFF
--- a/perma_web/perma/migrations/0017_rename_mylinks_to_personallinks.py
+++ b/perma_web/perma/migrations/0017_rename_mylinks_to_personallinks.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+from django.core.exceptions import ObjectDoesNotExist
+
+
+def rename_folders_helper(apps, schema_editor, old, new):
+    Folder = apps.get_model("perma", "Folder")
+    try:
+        folders = Folder.objects.filter(name=old)
+        folders.update(name=new)
+    except ObjectDoesNotExist:
+        pass
+
+def rename_folders(apps, schema_editor):
+    rename_folders_helper(apps, schema_editor, "My Links", "Personal Links")
+
+def reverse_rename_folders(apps, schema_editor):
+    rename_folders_helper(apps, schema_editor, "Personal Links", "My Links")
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('perma', '0016_auto_20161004_2215'),
+    ]
+
+    operations = [
+        migrations.RunPython(rename_folders, reverse_code=reverse_rename_folders)
+    ]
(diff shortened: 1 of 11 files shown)


perma_web/perma/migrations/0017_rename_mylinks_to_personallinks.py
```diff
@@ -0,0 +1,30 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+from django.core.exceptions import ObjectDoesNotExist
+
+
+def rename_folders_helper(apps, schema_editor, old, new):
+    Folder = apps.get_model("perma", "Folder")
+    try:
+        folders = Folder.objects.filter(name=old)
```
jcushman: Suggest adding the filter `is_root_folder=True`, to avoid changing any folders that happen to be called "My Links".

perma_web/perma/templates/user_management/create-link.html
```diff
@@ -153,6 +153,15 @@ <h3 class="link-header body-ah">Your Perma Links</h3>
...
 
+  {% if org_user %}
+    <script>
+      if (links_remaining == 10){
+        var message = "Your personal links for the month are almost used up! Create more links in 'unlimited' folders."
+        Helpers.informUser(message, 'danger');
+      }
+    </script>
+  {% endif %}
+
```
jcushman: This probably wants to live in create.module.js. Could use a custom JS variable `is_org_user`, if you can't get that status from `current_user` defined just above.

(This is nitpicky now, but will be necessary post-JS-refactor when Helpers stops being in the global namespace and we start reporting JS test coverage and so on.)

perma_web/perma/views/link_management.py
```diff
@@ -58,7 +58,8 @@ def create_link(request):
         'links_remaining': links_remaining,
         'folder': folder,
         'suppress_reminder': suppress_reminder,
-        'max_size': max_size
+        'max_size': max_size,
+        'org_user': request.user.is_organization_user
```
jcushman: Total nitpick: `request` is always in the template context, so I would tend to just use `{{ request.user.is_organization_user }}`.

perma_web/perma/migrations/0017_rename_mylinks_to_personallinks.py
jcushman: (I'm hoping any API authors are using this folder flag/user property rather than relying on the folder name, incidentally.)

rebeccacremona:
All eminently sensible suggestions. Fixing now. I also observe that I left the message warning at 10 (so it displays in dev) instead of lowering it to the desired count, 3!

PR #1810, opened 2016-11-15
Email excitement

rebeccacremona:
This requires three new values in settings.py:

CAMPAIGN_MONITOR_AUTH = {'api_key':'fake'}
CAMPAIGN_MONITOR_REGISTRAR_LIST = 'fake'
INTERNAL_SERVICES_KEY = 'fake'

So, tests will fail until we add dummy values to Heroku.

- use templates for all emails sent by Perma
- route for syncing registrar users + stats to campaign monitor
- route for sending an email to all registrar users via Perma's existing email service (draft email text only)


perma_web/perma/views/service.py
```diff
@@ -1,37 +1,95 @@
...
-        "%s \n\n(This link is the Perma link)" % link_url,
-        email_address
-    )
+logger = logging.getLogger(__name__)
 
-    response_object = {"sent": True}
+@csrf_exempt
+@sensitive_post_parameters()
+@require_http_methods(["POST"])
+def cm_sync(request):
```
jcushman: As we discussed, if this wants to be run nightly, it might want to be a celerybeat task instead of a route:

https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/settings/utils/post_processing.py#L26

perma_web/perma/views/service.py
```diff
@@ -1,37 +1,95 @@
...
+                          'email/admin/sync_to_cm.txt',
+                          {"reports": reports})
+        return HttpResponse(json.dumps(reports), content_type="application/json", status=200)
+    else:
+        return HttpResponseForbidden("<h1>Forbidden</h1>")
+
+@csrf_exempt
+@sensitive_post_parameters()
+@require_http_methods(["POST"])
+def ping_registrar_users(request):
```
jcushman: As we discussed, if this is being run manually by devs, it might make sense as a fab task.

If it's being run manually by non-dev admins, it could instead be added as a button in the Django admin: http://stackoverflow.com/a/24553887/307769

PR #1831, opened 2016-12-08
Send contact form messages from org users to registrar users 

rebeccacremona:
All is unchanged for not-logged-in users, regular users, and registrar users.

Org users associated with a single org still see the normal contact form, but a hidden field naming their registrar is included in the form:
image

Org users associated with multiple orgs are presented with a drop down list:
image

If, in config, CONTACT_REGISTRARS = True (False by default via common.py), then when the form is submitted, an email is sent to all the active registrar users associated with the org, and both Perma and the sender are cc'ed. And, the user is delivered at a customized thank you page, listing the names and email addresses of everyone who has been emailed:

In all other situations, an email is sent to Perma as usual, and the normal thanks page is displayed.
image

### DIFF
--- a/perma_web/perma/email.py
+++ b/perma_web/perma/email.py
@@ -29,11 +29,11 @@ def send_user_email(to_address, template, context):
     return success_count
 
 # def send_mass_user_email(template, recipients):
-    '''
-        Opens a single connection to the mail server and sends many emails.
-        Pass in recipients as a list of tuples (email, context):
-        [('recipient@example.com', {'first_name': 'Joe', 'last_name': 'Yacoboski' }), (), ...]
-    '''
+    # '''
+    #     Opens a single connection to the mail server and sends many emails.
+    #     Pass in recipients as a list of tuples (email, context):
+    #     [('recipient@example.com', {'first_name': 'Joe', 'last_name': 'Yacoboski' }), (), ...]
+    # '''
     # to_send = []
     # for recipient in recipients:
     #     to_address, context = recipient
--- a/perma_web/perma/email.py
+++ b/perma_web/perma/email.py
@@ -3,7 +3,7 @@ from createsend import Subscriber, List
 from collections import defaultdict
 
 from django.conf import settings
-from django.core.mail import EmailMessage, send_mail, send_mass_mail
+from django.core.mail import EmailMessage, send_mail
 from django.template.loader import render_to_string
 
 from .models import Registrar, LinkUser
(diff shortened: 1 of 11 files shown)


perma_web/perma/views/common.py
```diff
@@ -257,17 +296,30 @@ def contact(request):
...
+                }
+            )
+        )
 
         context = RequestContext(request, {'form': form})
         return render_to_response('contact.html', context)
+
+def contact_thanks(request):
+    """
+    When a user hits a rate limit, send them here.
```
jcushman: Update comment ...

perma_web/perma/views/common.py
```diff
@@ -206,40 +208,77 @@ def contact(request):
...
+            if len(registrars) == 1:
+                form.fields['registrar'].widget = widgets.HiddenInput()
+                email = registrars.pop().email
+                form.fields['registrar'].initial = email
+                form.fields['registrar'].choices = [(email,email)]
+        else:
+            del form.fields['registrar']
+        return form
+
+    def should_send_to_registrar():
```
rebeccacremona: Good idea. In the design phase where we had a checkbox for "Is this a bug?" on the form, this function included "and not is_bug", so (at least some version of) it had to be here rather than in the model, but now.... no need! And thanks for the tip on AnonymousUser.

perma_web/perma/views/common.py
```diff
@@ -206,40 +208,77 @@ def contact(request):
...
-            return HttpResponseRedirect(reverse('contact_thanks'))
+            context = {
+                "message": form.cleaned_data['message'],
+                "from_address": from_address,
+                "referer": form.cleaned_data['referer'],
+                "affiliation_string": affiliation_string()
+            }
+            if should_send_to_registrar():
+                # Send to all active registar users for registrar and cc Perma
+                reg_email = form.cleaned_data['registrar']
```
rebeccacremona: So this is actually taken care of already. By using a ChoiceField, and by adding only the user's own registrars as field.choices, we ensure that form validation will fail if anything but one of the user's own registrars is posted. I have a test for this: test_contact_org_user_submit_invalid

perma_web/perma/views/common.py
```diff
@@ -206,40 +208,77 @@ def contact(request):
...
+                registrar_users = LinkUser.objects.filter(
```
rebeccacremona: Good idea, thanks.

perma_web/perma/views/common.py
rebeccacremona: FWIW, for the AnonymousUser trick to work, one has to remove the @cached_property decorated from LinkUser.is_organization_user(). That seems...... potentially like a risk worth taking? Do you have memories of performance problems related to that property?

perma_web/perma/email.py
```diff
@@ -61,6 +61,21 @@ def send_admin_email(title, from_address, request, template="email/default.txt",
         headers={'Reply-To': from_address}
     ).send(fail_silently=False)
 
+def send_user_email_copy_admins(title, from_address, to_addresses, request, template="email/default.txt", context={}):
```
rebeccacremona: Right, have made all suggested improvements except this one. Think I'd rather bite that off later, if you don't mind.

PR #1837, opened 2016-12-15
Prep for Django 1.10, part 1

rebeccacremona:
render templates using render, not render_to_response


jcushman:
lines changed: +27 −66 

<3

PR #1850, opened 2017-01-05
Capture srcset images

rebeccacremona:
Phanton JS does not support srcset: http://stackoverflow.com/questions/28129210/srcset-attribute-in-phantomjs

And, even if it did, we want to capture all the images, so that playbacks work correctly regardless of screen size or screen resolution.

So, let's grab them like favicons.

Test case: lil.law.harvard.edu and lil.law.harvard.edu/projects/perma-cc


rebeccacremona:
Do you mind reviewing this one, Jack? I haven't worked out how to test this yet; will tackle that some other time.

jcushman:
Cool! Welcome to the capture code. :)

My one suggestion is to take your new logic and move it down inside the `def get_media_tags(browser)` function. (I think you'll just need the part inside your `if have_html:`, since `get_media_tags()` is already returning a set of URLs to fetch.)

This means we'll get a few extra edge cases taken care of for free -- e.g. the srcset logic will be run in every frame, not just the top frame, and only URLs that haven't already been captured directly will be added.

rebeccacremona:
Oh, cool. Hadn't thought about frames at all. I'll tweak that section too so that this particular bit runs even if settings.ENABLE_AV_CAPTURE = False

rebeccacremona:
This should do it. Let me know if I mangled the style of your function.

jcushman:
So cool!

PR #1859, opened 2017-01-10
uploading all links to internet archive

anastasia:
instead of just those created in the last forty eight hours
also looking out for links that have been deleted previously but are public

### DIFF
--- a/perma_web/perma/tasks.py
+++ b/perma_web/perma/tasks.py
@@ -874,10 +874,12 @@ def upload_all_to_internet_archive():
         - links that have not started to upload
         - links that have failed to upload but not failed permanently (those should not exist, but doesn't hurt to look)
         - links that have previously been deleted from internet archive (is_private=True) but have since been made public
+        - only links that are older than 24 hours
     """
-    links = Link.objects.filter(Q(internet_archive_upload_status='not_started') |
-                                Q(internet_archive_upload_status='failed') | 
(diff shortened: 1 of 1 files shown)


anastasia:
There are currently 200,000+ links not previously archived to internet archive. We never ran upload on all the old links, so links have only been uploading since the feature has been added. I see no reason that we should be looking at just the last forty eight hours, but maybe you can think of one?

perma_web/perma/tasks.py
```diff
@@ -869,12 +869,18 @@ def delete_from_internet_archive(self, link_guid):
...
+    """
+        Links to upload:
+        - links that have not started to upload
+        - links that have failed to upload but not failed permanently (those should not exist, but doesn't hurt to look)
+        - links that have previously been deleted from internet archive (is_private=True) but have since been made public
+        - only links that are older than 24 hours
+    """
+    links = Link.objects.filter((Q(internet_archive_upload_status='not_started') |
+                                Q(internet_archive_upload_status='failed') |
+                                Q(internet_archive_upload_status='deleted', is_private=False)) &
```
anastasia: We delete previously public links that are going private from Internet Archive https://github.com/harvard-lil/perma/blob/develop/perma_web/api/resources.py#L632

We should upload links previously private that are going public TO Internet Archive

anastasia:
Closing pending fab synchronous command to clear the big backlog, will reopen when that's done.

PR #1862, opened 2017-01-12
Add queue and capture times to API

bensteinberg:


jcushman:
One tweak -- I believe that as written this will result in an additional SQL query per returned record. This can be avoided by including 'capture_job' in Meta.query_set.select_related().

To see what I mean, try adding this to settings.py:

```
# log all SQL statements:
LOGGING['loggers']['django.db.backends'] = {
            'level': 'DEBUG',
            'handlers': ['console'],
        }
```

I think under this pull request you'll see a `SELECT * FROM perma_capture_job WHERE id='whatever'` for each record returned by the API. With the right input to `select_related()` you can move that into a single join in the initial `SELECT * FROM perma_link`.

bensteinberg:
Thanks! See what you mean!

bensteinberg:
Although this is better than one additional query per GUID, it looks like it's still doing one additional query for all the capture_job times, rather than including it as a join in the main query. Will look at this some more.

bensteinberg:
Moving capture_job into prefetch_select results in two additional queries rather than one. Do you think it should be possible to do this with no additional queries?

jcushman:
I think this is working correctly. You're seeing a second query caused by `prefetch_related('captures')`, which was already happening and is distinct from `capture_job`.

PR #1884, opened 2017-02-14
Ensure all capture time fields are present

bensteinberg:


perma_web/perma/views/user_management.py
```diff
@@ -110,10 +110,11 @@ def stats(request, stat_type=None):
             )
             if capture_time_fields:
                 ctf_len = len(capture_time_fields)
-                capture_times = sorted(c['capture_end_time']-c['capture_start_time'] for c in capture_time_fields)
-                wait_times = sorted(c['capture_start_time']-c['link__creation_timestamp'] for c in capture_time_fields)
-                day['capture_time_dist'] = " / ".join(str(i) for i in [capture_times[int(ctf_len*.05)], capture_times[int(ctf_len*.5)], capture_times[int(ctf_len*.95)]])
-                day['wait_time_dist'] = " / ".join(str(i) for i in [wait_times[int(ctf_len*.05)], wait_times[int(ctf_len*.5)], wait_times[int(ctf_len*.95)]])
+                if ctf_len == 3:
```
jcushman: I don't think this'll quite do it. `capture_time_fields` is a list that looks like this:

    [
        {'capture_start_time':<date>, 'link__creation_timestamp':<date>, 'capture_end_time':<date>,
        {'capture_start_time':<date>, 'link__creation_timestamp':<date>, 'capture_end_time':None,
    ]

... and the issue is that we have to do something sensible when one of the dates is `None`, which is tricky and maybe not that useful.

Rather than try to solve that, what if we add an `.exclude(capture_start_time=None).exclude(capture_end_time=None)` to the initial query?

bensteinberg: Ah, right. Thanks!

PR #1905, opened 2017-03-03
Lazy load folders

jcushman:
This pull req makes folders load in their subfolders on demand instead of loading the entire tree at once.

API changes:

* /user/ includes a top_level_folders key with a list of the user's top level folders.
* /folders/:id/ includes an organization key with the organization the folder belongs to, if any, and a has_children boolean key that is true is the folder has subfolders.

Code changes:

The simple change in the code is to have jsTree hit the API and fetch subfolders when a folder is opened.

Most of the changes deal with remembering the previous folder the user had open. For that, we have to change localStorage to store the entire path to the folder instead of just the final folderId. Then on load, we have to fetch the folder contents at each level, and build the whole tree that leads to the selected folder. Finally we have jsTree's load_node event select the target node.

Todo:

This doesn't yet update the API docs to reflect the new keys.


anastasia:
@jcushman I think the files need to be recompiled

jcushman:
Thanks! Fixed.

PR #1906, opened 2017-03-03
added line break before "Troubleshooting info" line

bjohnson747:


rebeccacremona:
Per discussion, going to handle this a bit differently.

PR #1920, opened 2017-03-12
fix - buttons spacing on settings->tools page in mobile display

prastut:
Minor bug:

screen shot 2017-03-13 at 2 21 57 am

There should be some gap between install btn `Perma for Chrome` and btn `Install Perma for Firefox`. Added the styles to maintain visual consistency.

### DIFF
--- a/perma_web/static/bundles/global-styles.css
+++ b/perma_web/static/bundles/global-styles.css
@@ -8765,6 +8765,12 @@ a.cc:active {
   color: white;
 }
 
+@media only screen and (max-width: 767px) {
+  .btn + .btn {
+    margin-left: 16px;
+  }
+}
+
 @media only screen and (min-width: 768px) {
   .btn {
     white-space: nowrap;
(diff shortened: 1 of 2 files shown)


prastut:
@rebeccacremona makes sense. 

I tried to make the desired changes i.e writing the css without any media queries and I was running into problems which I earlier didn't notice. 

Generally the effect should work like the following: 
screen shot 2017-03-14 at 9 30 08 pm

^ This works on phone screen sizes having width above 412px. 

But as soon as you reduce the screen size, this happens:

rebeccacremona:
Yes, using an "!important" override here doesn't sound like the best plan. We also want to be quite careful here, because the .btn class is a very broad class that affects every button across the application, not just these bookmarklet buttons. (There is a .btn-bookmarklet class which, more safely, only affects these bookmarklet buttons.) Your idea of a 100% width button on phone-sized screens is a good one; I gave that styling a look locally, and found that it (without changing any margin-related code) looked good on phones, but looked rather peculiar on tablet-sized screens.

rebeccacremona:
I'm going to close this for now. Thanks again for pointing out the behavior of these buttons. Please feel free to reopen the pull request in the future if you are so inspired.

PR #1926, opened 2017-03-13
wait longer for metatags

rebeccacremona:
We're getting way too many failures here. ~115 out of the 160 links currently "private due to meta tag  analysis failure" are from sites other than NYTimes. All the non--NY Times links that I've spot checked so far can be made public.

In all cases, the problem seems to be that the page runs large numbers of slow-loading ad libraries and trackers that each significantly alter (or replace entirely?) the document head, possibly even rewriting the entire document. I suspect there is something improper about their behavior, but I can't pin it down. (But even if I did, I don't think we can work around it.)

I'm hoping if we wait longer (increasing the upper bound from from 10 to 30 seconds), this will happen much less frequently.

If it doesn't, we might want to revisit our approach.

### DIFF
--- a/perma_web/perma/tasks.py
+++ b/perma_web/perma/tasks.py
@@ -584,7 +584,7 @@ def proxy_capture(capture_job):
 
                 # get all meta tags
                 meta_tags = repeat_while_exception(lambda: browser.find_elements_by_tag_name('meta'),
-                                                   timeout=10)
+                                                   timeout=30)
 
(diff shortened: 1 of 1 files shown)


rebeccacremona:
p.s. These are seriously slow-loading pages. And, though they are at very mainstream urls, a lot of them have some seriously messed up js. I only spot checked 6. Of those 6, 4 had over 8000 console error messages after 60 seconds. 3 caused Chrome dev tools to crash after 3 min.

PR #1929, opened 2017-03-15
Folder drag fix

jcushman:
- Fix folder drag and drop problem reported by Adam.
- Handle case where second folder is created/dragged/renamed so as to end up with same name as existing folder.
- Unexpected jstree errors (such as duplicate folder errors) will be reported to user instead of swallowed, so we don't have things fail for no apparent reason.
- Refactor API test helpers to avoid duplication in rejected_<request method> functions.

### DIFF
--- a/perma_web/static/js/folder-tree.module.js
+++ b/perma_web/static/js/folder-tree.module.js
@@ -8,6 +8,7 @@ var Helpers = require('./helpers/general.helpers.js');
 var localStorageKey = Helpers.variables.localStorageKey;
 var allowedEventsCount = 0;
 var lastSelectedFolder = null;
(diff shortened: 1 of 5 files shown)


perma_web/static/js/folder-tree.module.js
```diff
@@ -8,6 +8,7 @@ var Helpers = require('./helpers/general.helpers.js');
 var localStorageKey = Helpers.variables.localStorageKey;
 var allowedEventsCount = 0;
 var lastSelectedFolder = null;
+var hoveredNode = null;  // track currently hovered node in jsTree
```
rebeccacremona: So bizarre that this works and the jquery version doesn't. What a crazy world.

PR #1935, opened 2017-03-20
If browser bails out mid capture

rebeccacremona:
Diff is somewhat hard to read because of changed indentation, or lines moved to a new spot.

Using browser_still_running repeatedly because the browser may crash at any point during capture; checking before each block is ugly, but should be more robust, and might make it easier to introduce custom error handling in the future.

Introduces a new setting PRVIATE_LINKS_ON_FAILURE, False by default, so we can more easily toggle this behavior, if the group decides a conservative approach to these links is prudent.

If this request (or something like it) is merged in, I'll clean up the production data to match the new behavior afterwards.


perma_web/perma/models.py
```diff
@@ -106,6 +107,9 @@ def safe_delete(self):
         self.user_deleted = True
         self.user_deleted_timestamp = timezone.now()
 
+class GenericStringTaggedItem(CommonGenericTaggedItemBase, TaggedItemBase):
```
rebeccacremona: Totally. Will add tomorrow.

jcushman:
Thanks!

PR #1937, opened 2017-03-23
Fab task for comparing WARCs in primary and secondary storage

bensteinberg:


bensteinberg:
It would be nice not to have to get all WARCs from S3. If Storage Inventory is too awkward, maybe get ETags via API calls? At the moment, this task doesn't care what kind of storage primary and secondary are.

perma_web/fabfile/dev.py
```diff
@@ -573,3 +573,89 @@ def check_s3_hashes():
...
+def check_storage():
+    """
+        Confirm that, for every link, there is a WARC in primary storage and in
+        secondary storage, and that their hashes match.
+
+        Derived from check_s3_hashes
+    """
+    from django.core.files.storage import default_storage
+    from perma.models import Link
+    from tqdm import tqdm
```
rebeccacremona: I had only heard of this as Spanish shorthand:
> tqdm means “progress” in Arabic (taqadum, تقدّم) and an abbreviation for “I love you so much” in Spanish (te quiero demasiado).

perma_web/fabfile/dev.py
```diff
@@ -573,3 +573,89 @@ def check_s3_hashes():
...
+        'match': 0,
+        'mismatch': 0,
+        'primary_only': 0,
+        'secondary_only': 0,
+        'no_warc': 0
+    }
+    cache_path = '/tmp/perma_path_list'
+
+    if not os.path.exists(cache_path):
+        print("Building state ...")
```
rebeccacremona: Very optional readability suggestion, likely occurring to me only because I'm braindead: "building link list" and "using cached link list" might be clearer than "state". I also wonder if it's worth using .txt as an extension for cache_path, and calling it cache_file rather than cache_path. Feel free to ignore.

perma_web/fabfile/dev.py
```diff
@@ -573,3 +573,113 @@ def check_s3_hashes():
...
+    if hasattr(default_storage, 'secondary_storage'):
+        storages.update({'secondary': {'storage': default_storage.secondary_storage, 'lookup': {}}})
+
+    # only use cache files when all are present: link cache, and one for each storage
+    link_cache = '/tmp/perma_link_cache.txt'
+    caches = [link_cache]
+    for key in storages:
+        caches.append('/tmp/perma_storage_cache_{0}.txt'.format(key))
+
+    if not all(map(os.path.exists, caches)):
```
rebeccacremona: Hooray for all and map! Since we are talking about a list of length 2 at this point, certainly no optimization is needed, but mentioning just for fun: it might be preferable to use a generator here instead of map. ```all(os.path.exists(p) for p in caches)``` would stop as soon as os.path.exists returned false; unless I'm mistaken, all(map...) will calculate os.path.exists for every p in caches, and then start looking for a falsy value. But again, totally doesn't matter here.

perma_web/fabfile/dev.py
```diff
@@ -573,3 +573,113 @@ def check_s3_hashes():
...
+            storage = storages[key]['storage']
+            with open('/tmp/perma_storage_cache_{0}.txt'.format(key), 'w') as tmp_file:
+                if hasattr(storage, 'bucket'):
+                    # S3
+                    for f in tqdm(storage.bucket.list('generated/warcs/')):
+                        # etag is a string like u'"3ea8c903d9991d466ec437d1789379a6"', so we need to
+                        # knock off the extra quotation marks
+                        # note that etags are not always md5sums, but should be in these cases; we can rewrite
+                        # or replace md5hash if necessary
+                        path = f.key[(len(storage.location)):]
```
rebeccacremona: Is this is where we are "chopping off the prefix, whether storage.location..."? Total nitpick: I personally would find it easier to read this code if that comment were here, modifying this line. It's not that the concept is confusing; I think I was thrown because the comment and the code are in such different scopes. No need to address if you disagree.

perma_web/fabfile/dev.py
```diff
@@ -573,3 +573,113 @@ def check_s3_hashes():
...
+                    else:
+                        # local file storage -- are there other possibilities to consider?
+                        base = storage.base_location
+                    for f in tqdm(storage.walk(os.path.join(base, 'warcs'))):
+                        # os.walk: "For each directory in the tree rooted at directory top (including top itself),
+                        # it yields a 3-tuple (dirpath, dirnames, filenames)" -- there should be exactly one WARC
+                        # per directory, so:
+                        if len(f[2]) == 1:
+                            full_path = os.path.join(f[0], f[2][0])
+                            path = full_path[len(base):]
```
rebeccacremona: Same here: a duplicative comment might not be a bad thing here, just to make it clearer what's happening where. Or not.

perma_web/fabfile/dev.py
```diff
@@ -573,3 +573,113 @@ def check_s3_hashes():
...
+                            hash = md5hash(full_path, storage)
+                            tmp_file.write("{0}\t{1}\n".format(path, hash))
+                            storages[key]['lookup'][path] = hash
+    else:
+        print("Reading storage caches ...")
+        for key in storages:
+            with open('/tmp/perma_storage_cache_{0}.txt'.format(key)) as f:
+                for line in f:
+                    path, hash = line[:-1].split("\t")
+                    storages[key]['lookup'][path] = hash
```
rebeccacremona: Since each path and hash pair for each storage are loaded into memory here, why go through the intermediate step of building a storage cache file? Just for record keeping purposes? (Which might be plenty reason to do it!)

perma_web/fabfile/dev.py
```diff
@@ -573,3 +573,113 @@ def check_s3_hashes():
...
+                if path not in storages[key]['lookup']:
+                    print("{0} not in {1}".format(path, key))
+                    file_present = False
+            if file_present and len(storages) > 1:
+                hashes = []
+                for key in storages:
+                    hashes.append(storages[key]['lookup'][path])
+                if hashes.count(hashes[0]) != len(hashes):
+                    print("Hash mismatch for {0}: {1}".format(path, str(zip(storages.keys(), hashes))))
+
```
rebeccacremona: In 660 and 667: would it be good to save these results to a file as well?

perma_web/fabfile/dev.py
rebeccacremona: Oh wait, these are in the tmp directory, so none of these caches stick around. So my guess that it's for record-keeping purposes must be wrong.

perma_web/fabfile/dev.py
```diff
@@ -573,3 +573,113 @@ def check_s3_hashes():
...
+            file_present = True
+            for key in storages:
+                if path not in storages[key]['lookup']:
+                    print("{0} not in {1}".format(path, key))
+                    file_present = False
+            if file_present and len(storages) > 1:
+                hashes = []
+                for key in storages:
+                    hashes.append(storages[key]['lookup'][path])
+                if hashes.count(hashes[0]) != len(hashes):
```
rebeccacremona: That's a clever approach! Here are a couple other ways I'm wondering about, which might be more readable. (They might also be a bit faster, but again, it's not like we are dealing with long lists here.) 
```len(set(hashes)) == 1```
or 
```all(hash == hashes[0] for hash in hashes)```

bensteinberg: I found this at http://stackoverflow.com/a/3844948/4074877 -- thought I'd put the link in a comment.... -- "A solution faster than using set() that works on sequences (not iterables) is to simply count the first element. This assumes the list is non-empty (but that's trivial to check, and decide yourself what the outcome should be on an empty list)"

perma_web/fabfile/dev.py
bensteinberg: Yeah, good idea. I thought I was overcommenting already, but.

perma_web/fabfile/dev.py
jcushman: I think this caching pattern is borrowed from the task I wrote previously, and the benefit of caching is that the hashing step can take hours and cost $$ to run. If you then hit some edge case that blows things up in the comparison step, it's nice not to have to run the hashing step again ...

rebeccacremona: Sold!

perma_web/fabfile/dev.py
rebeccacremona: Wow! That's impressively faster. Cool. Maybe the comment would be good, especially since we are talking about a list of length 2 rather than long lists like they were comparing?

perma_web/fabfile/dev.py
bensteinberg: I think I'm going to run this and pipe it to a file, fwiw.

PR #1945, opened 2017-03-29
Add start_date argument to check_storage

bensteinberg:


bensteinberg:
I could name the cache files with start_date so that there wouldn't be any ambiguity about using a cache with a different or no start date. Not sure it's worth it.

PR #1949, opened 2017-04-02
Django Rest Framework

jcushman:
This pull request reimplements our API based on Django Rest Framework.

The new API currently lives at /api/v2. You can tell the frontend to use it by setting API_VERSION=2. While this is handy for comparison, once we are satisfied with the new API, we may want to move it to /api/v1. The user-visible changes are small (but do exist, see below), and we probably don't want to maintain a dependency on Tastypie for /api/v1.

As a general comment on the change: it seems to be worth it. The new code is 70% as long as the old code, and feels much more straightforward. You can pretty much trace a request from beginning to end looking at just a couple of places in views.py and serializers.py. DRF mimics the usual Django urls.py -> views.py -> forms.py pipeline, but using APIViews and Serializers instead of regular views and forms. I like this balance -- we're getting a lot handled automatically for us in terms of serializing and deserializing, but it's not too far from plain old Django.

DRF also comes with a nice human-readable HTML interface, which you can reach by viewing the api endpoints (like perma.dev:8000/api/v2/) in your browser. I haven't played much with this and it might need polishing if we want to promote it to API users.

*Code review*

For code review, the changed files break down as follows:

- api/: small changes to add the /api/v2 endpoint to urls.py and maintain compatibility
- api2/: new API
    - to make sense of things, maybe look at README.md, then urls.py, then views.py, then serializers.py
- api2/tests/: copy of tests from api/tests/. This retains all substantive tests, but with changes to deal with minor differences between DRF and Tastypie (see below).
- perma/models.py: add `accessible_to(user)` method for all models that need to be permission-checked.
- perma/templates/ and static/js/: a couple of our frontend forms were using a hardcoded /api/v1 URL. Update these to use the same `api_path` JS variable as our other forms. Also update error handling to handle /v2 and /v1 error dictionaries (see below).

*User-visible changes*

Here are the things I know of that are different between /v2 and /v1:

1. Many endpoints return a different 4XX error code for failures. Some 401s or 404s become 403s, some 401s become 400s, etc. Basically I went with the DRF standard for returned error codes instead of patching things to emulate Tastypie. *Rationale:* it doesn't seem worth complicating the code to emulate Tastypie when there's no clear standard for which error code to return anyway.
2. The Tastypie /schema endpoints are gone (e.g. /api/v1/archives/schema). *Rationale:* those are a Tastypie-specific thing and don't follow a standard, and I don't think anyone's using them. We could add a [coreapi schema](http://www.django-rest-framework.org/api-guide/schemas/) later if we want.
3. Returned error dictionaries no longer include the name of the requested endpoint. For example, for a POST to /api/v1/archives, Tastypie would return `{"archives":{"url":"URL is required."}}`. DRF just returns `{"url":["URL is required."]}`. *Rationale:* I like the DRF way better. It seems annoying and non-DRY to have to keep track of the endpoint URL to parse the error dict.
4. Filter options for searching links have changed. v1 options are `q` (text search), `submitted_url` (url search), `date` (creation date search), and `date_range` (search within X months of `date`). v2 options are `q` (text search), `url` (url search), `date`, `min_date`, and `max_date` (creation date search). *Rationale:* I took the opportunity to clean these up. `submitted_url` doesn't make sense as the name of a query option, since we expose the field as `url` in API responses. `date_range` seems needlessly specific, instead of accepting arbitrary min and max dates. I don't think anyone is probably using these filters as-is, other than our /search page.
5. Several endpoints have been moved (although they also continue to work at the old locations for backwards compatibility, see urls.py). `/user/folders` is now merged into the existing `/folders` endpoint, `/user/archives` is now merged into `/archives`, `/user/organizations` is renamed `/organizations`, and `/user/capture_jobs` is renamed `/capture_jobs`. *Rationale:* we had a confusing split between endpoints under /user or not under /user. /capture_jobs and /organizations were just under /user. /folders and /archives were part under /user and part not -- the list parts were only under /user, but the POST parts were only not. This division was originally intended to ensure that urls outside of /user would always return the same thing for different logged-in users, which in theory is more RESTful, but in practice adds conceptual overhead for both us and API consumers without any clear benefit. Flattening everything out is simpler for everybody.
6. /archives now includes all archives accessible_to the user, not those created_by the user.  /folders now includes all top-level folders for the user, not all folders created_by the user. Rationale: this fixes a logic error in the v1 API. Since links and folders can be moved around and original creator can lose access to them, checking created_by made no sense.

Major questions for code review:

- What could be clearer or better documented? (Feel free to ask questions or add comments directly, whichever's easier.)
- Can we get away with swapping this in as /api/v1, given the changes?
- And of course the usual question of what's broken or left out.


perma_web/api2/README.md
```diff
@@ -0,0 +1,26 @@
...
+from start to end within a single file, and to see all of the routes and request methods exposed by the API.
+
+The two features of DRF that we do make heavy use of are class-based views (which basically work like Django's class-based
+views) and serializers (which basically work like Django's ModelForms).
+
+Using LOG_API_CALLS to Debug Tests
+----------------------------------
+
+When an API test fails, a great first step (and often the only required step) is to see what API endpoints are being
+called by the test, what user is calling them, and what they are returning.  You can enable logging of that information
```
rebeccacremona: Further reading will make this clearer to me, but: does this suggest to you I have some work to do on the tests? I would have imagined that only one endpoint would be hit per unit test, and that, like the rest of our tests, they would fail with a meaningful error message. I'll study up, but would also love to chat about this offline sometime when convenient.

perma_web/api2/serializers.py
```diff
@@ -0,0 +1,242 @@
...
+from utils import get_mime_type, mime_type_lookup
+from perma.models import LinkUser, Folder, CaptureJob, Capture, Link, Organization
+from perma.utils import ip_in_allowed_ip_range
+
+
+class BaseSerializer(serializers.ModelSerializer):
+    """ Base serializer from which all of our serializers inherit. """
+
+    def update(self, instance, validated_data):
+        """ When updating, requiring that our serializers provide a whitelist of fields that can be updated. """
```
rebeccacremona: The LinkUserSerializer does not include a field in Meta "allowed_update_fields". Is that an oversight? Or would this method be better like this?
```
    def update(self, instance, validated_data):
        """Update object, failing validation if a serializer provides allowed_update_fields and disallowed fields are passed in."""
        if getattr(self, 'Meta.allowed_update_fields', False) and set(validated_data.keys()) - set(self.Meta.allowed_update_fields):
            raise serializers.ValidationError('Only updates to these fields are allowed: %s' % ', '.join(self.Meta.allowed_update_fields))
        return super(BaseSerializer, self).update(instance, validated_data)

File: perma_web/api2/serializers.py [hunk 3]
```diff
@@ -0,0 +1,242 @@
...
+        model = Folder
+        fields = ('id', 'name', 'parent', 'has_children', 'organization')
+        extra_kwargs = {'parent': {'required': True}}
+        allowed_update_fields = ['name', 'parent']
+
+    def get_has_children(self, folder):
+        return not folder.is_leaf_node()
+
+    def validate_name(self, name):
+        if self.instance:
```
rebeccacremona: What does it mean if self.instance is falsy here? Does that mean we are in the process of trying to create a new folder, and so don't yet have a saved folder object? (I do see that the DRF has a default value of None for self.instance)

perma_web/api2/serializers.py
```diff
@@ -0,0 +1,242 @@
...
+            if self.instance.is_shared_folder:
+                raise serializers.ValidationError("Can't move organization's shared folder.")
+            if self.instance.is_root_folder:
+                raise serializers.ValidationError("Can't move user's main folder.")
+        return parent
+
+    def validate(self, data):
+        if 'name' in data:
+            # make sure folder name is unique in this location
+            parent_id = data['parent'].pk if 'parent' in data else self.instance.parent_id if self.instance else None
```
rebeccacremona: I'm a little confused about the parent field. Since extra_kwargs = {'parent': {'required': True}}, 'parent' should always explicitly be in data, no? (Even if it's "null" a.k.a = None). Or does this have to do with the order validators are run (i.e., this runs before DRF checks to make sure all required fields are present)?

perma_web/api2/serializers.py
```diff
@@ -0,0 +1,242 @@
...
+            if parent_id:
+                unique_query = Folder.objects.filter(parent_id=parent_id, name=data['name'])
+                if self.instance:
+                    unique_query = unique_query.exclude(pk=self.instance.pk)
+                if unique_query.exists():
+                    raise serializers.ValidationError({'name':"A folder with that name already exists at that location."})
+        return data
```
rebeccacremona: Similar question as above: since name is a required field (default behavior), I would have thought we could rely on it always being present in data, without an explicit check. Not the case?

perma_web/api2/serializers.py
```diff
@@ -0,0 +1,242 @@
...
+class CaptureSerializer(BaseSerializer):
+    class Meta:
+        model = Capture
+        fields = ('role', 'status', 'url', 'record_type', 'content_type', 'user_upload', 'playback_url')
+
+
+### LINK ###
+
+class LinkSerializer(BaseSerializer):
+    url = serializers.CharField(source='submitted_url', max_length=2100, required=False, allow_blank=True)
```
rebeccacremona: It's too bad it doesn't inherit the max_length from the model by default. Oh well.

perma_web/api2/serializers.py
```diff
@@ -0,0 +1,242 @@
...
+class AuthenticatedLinkSerializer(LinkSerializer):
+    """
+        Subclass of LinkSerializer with extra fields for users who own the link.
+    """
+    created_by = LinkUserSerializer(read_only=True)
+    organization = OrganizationSerializer(read_only=True)
+
+    class Meta(LinkSerializer.Meta):
+        fields = LinkSerializer.Meta.fields + ('notes', 'created_by', 'is_private', 'private_reason', 'archive_timestamp', 'organization')
+        allowed_update_fields = ['submitted_title', 'submitted_description', 'notes', 'is_private', 'private_reason']
```
rebeccacremona: I presume this code already works (and I'll double check shortly), but commenting so I don't forget: in LinkSerializer, the fields are called plain "title" and "description" rather than "submitted_title" and "submitted_description" like they are in the model. I would have expected we'd need to list the Serializer field name in this whitelist, rather than the model field name.

perma_web/api2/serializers.py
```diff
@@ -0,0 +1,242 @@
...
+        return url
+
+    def validate(self, data):
+        user = self.context['request'].user
+        errors = {}
+
+        # handle private_reason:
+        if not user.is_staff:
+            # only staff can manually set private_reason
+            data.pop('private_reason', None)
```
rebeccacremona: Triviality: does it make sense for this popping off of 'private_reason') to take place inside the "if self.instance" branch, to make it clear that staff can only affect the private_reason when updating an existing link? I think probably yes. If I'm reading this correctly, any private_reason provided by a staff member for a not-yet-saved link would be overwritten in the "else" below.

perma_web/api2/serializers.py
```diff
@@ -0,0 +1,242 @@
...
+                else:
+                    data['private_reason'] = 'user' if data['is_private'] else None
+        else:
+            # for new links, set private_reason based on is_private
+            data['private_reason'] = 'user' if data.get('is_private') else None
+
+        # check submitted URL for new link
+        if not self.instance:
+            if not data.get('submitted_url'):
+                errors['url'] = "URL cannot be empty."
```
rebeccacremona: Another interesting instance of model field name vs. serializer field name.

perma_web/api2/serializers.py
```diff
@@ -0,0 +1,242 @@
...
+                except UnicodeError:
+                    # see https://github.com/harvard-lil/perma/issues/1841
+                    errors['url'] = "Unicode error while processing URL."
+                except DjangoValidationError:
+                    errors['url'] = "Not a valid URL."
+                except TooManyRedirects:
+                    errors['url'] = "URL caused a redirect loop."
+
+        # check uploaded file
+        uploaded_file = self.context['request'].data.get('file')
```
rebeccacremona: This can just be data.get('file'), like in 197, right?

perma_web/api2/urls.py
```diff
@@ -0,0 +1,67 @@
+from django.conf.urls import url
+from rest_framework.routers import APIRootView
+
+from perma.urls import guid_pattern
+
+import views
```
rebeccacremona: Curious about a triviality: in the perma urls.py, signals.py, models.py, and possibly elsewhere, when importing things from the current directory, it's done with a ./ (e.g. ./views or from ./views import blah blah blah). Is either one better style, so far as you know?

jcushman: I'm obviously not careful about this, but in theory I prefer the explicit relative import approach (`import .views`) for imports within a Django app.

We have three options here -- `import views`, `import api2.views`, or `import .views`.

`import views` is super vague -- it could refer to api/views.py or perma_web/views.py or some third party `views` library. You can imagine a scenario where this actually matters, like if you have `import celery` referring to a local celery.py file and then delete that file, and `import celery` still works but unexpectedly refers to something totally different.

`import api2.views` means you have to change all the import lines if you rename the app, and it doesn't immediately show the difference between within-app imports and cross-app imports:

    # one of these things is not like the other ...
    from perma.views import foo_helper
    from api2.views import foo

So my favorite is `import .views` -- it clearly expresses that this is a relative import within this modular, movable thing we call an app.

In practice it's such a big hairy deal to rename folders and files that this particular thing may never matter that much ...

perma_web/api2/serializers.py
rebeccacremona: Nope, parent is demonstrably not always in the data. I will work out why I find this confusing, and will consider adding a paragraph to the README after I figure it out.
```
    def validate(self, data):
        import pdb; pdb.set_trace()
        if 'name' in data:
            # make sure folder name is unique in this location
            parent_id = data['parent'].pk if 'parent' in data else self.instance.parent_id if self.instance else None
            if parent_id:
                unique_query = Folder.objects.filter(parent_id=parent_id, name=data['name'])
                if self.instance:
                    unique_query = unique_query.exclude(pk=self.instance.pk)
                if unique_query.exists():
                    raise serializers.ValidationError({'name':"A folder with that name already exists at that location."})
        return data
(rename a folder from Sunday to Sundae)
(Pdb) data
OrderedDict([(u'name', u'Sundae')])
```

perma_web/api2/serializers.py
jcushman: No, because 'file' isn't actually a field on the model -- by this point DRF has filtered down data to just known fields. I'll add a comment to explain why we access that in a different way.

perma_web/api2/serializers.py
rebeccacremona: Aha, this is because of "partial" updates. FolderSerializer(<Folder: Sundae>, context={'request': <rest_framework.request.Request object>}, data={'parent': 27L}, partial=True)

jcushman: Right! We instantiate serializers in two different ways in views.py:

To create new models: `serializer_class(data=data)`. `self.instance` is None, all required fields must be supplied or raise validation errors.

To update existing models: `serializer_class(obj, data=data, partial=True)`. `self.instance` is the existing instance. `partial=True` means that fields not supplied in `data` retain their existing values, so required checks don't apply.

perma_web/api2/serializers.py
rebeccacremona: (just updated my comment to reference the correct line number)

rebeccacremona: Okay, reversing the question then: should 201 (under "Don't force URL resolution validation if a file is provided") be self.context['request'].data.get('file') as well? I'm creating a new capture via the upload form, and url validation is still taking place, despite the presence of a file. (And verified that data doesn't contain file at that point.) Is that the situation this is intended to catch? (Sorry for still being confused).
url-validation

jcushman: Good catch! Yes, will fix that. Either by checking self.context['request'].data.get('file') or registering excrample.com

perma_web/api2/urls.py
rebeccacremona: Thanks for the thorough explanation!

perma_web/api2/utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+        And
+
+            class LinkDetailView():
+
+                @load_parent
+                def get(request, pk):
+                    ...
+
+        This decorator will make sure that request.parent is set to Folder(pk=2),
```
rebeccacremona: Typo, I think. Folder(pk=1). Also, triviality: maybe use /folders/1/folders/2 as an example (since archive IDs are GUIDS rather than ints, and since you can access /folders/int/folders/int with a  GET request)

perma_web/api2/views.py
```diff
@@ -0,0 +1,483 @@
...
+            raise ValidationError(e.error_dict)
+
+    def get_object_for_user(self, user, queryset):
+        """
+            Get single object from queryset, making sure that returned object is accessible_to(user).
+        """
+        try:
+            obj = queryset.get()
+        except ObjectDoesNotExist:
+            raise Http404
```
rebeccacremona: It's right not to handle MultipleObjectsReturned, right? If this function is misused that way, it SHOULD raise 500.

jcushman: That sounds right to me. It would either mean the function was misused or we had a data integrity error, which either way should be a server error.

perma_web/api2/views.py
```diff
@@ -0,0 +1,483 @@
...
+            queryset = request.user.top_level_folders()
+        return self.simple_list(request, queryset)
+
+    @load_parent
+    def post(self, request, format=None):
+        """ Create folder. """
+        # if parent folder is not supplied in post data, try to get it from /folders/:parent_id:
+        data = request.data.copy()
+        if request.parent:
+            data.setdefault('parent', request.parent.pk)
```
rebeccacremona: Neat. I didn't know this method existed on regular dictionaries.

perma_web/api2/views.py
```diff
@@ -0,0 +1,483 @@
...
+        return self.folder_update(request, pk, {'parent': request.parent.pk})
+
+    @load_parent
+    def delete(self, request, pk, format=None):
+        """ Delete folder. """
+        folder = self.get_object_for_user_by_pk(request.user, pk)
+
+        # delete validations
+        if folder.is_shared_folder:
+            raise_validation_error("Shared folders cannot be deleted.")
```
rebeccacremona: Triviality: should we reword this message, since you can delete an org's subfolders if they are empty? How about "Organizations' top-level folders cannot be deleted" or something similar? Or how about combining if folder.is_shared_folder and elif folder.is_root_folder and have them both say "Top-level folders cannot be deleted."

perma_web/api2/views.py
```diff
@@ -0,0 +1,483 @@
...
+        except Link.DoesNotExist:
+            raise Http404
+        return self.simple_get(request, obj=obj)
+
+# /archives
+# /folders/:parent_id/archives
+class AuthenticatedLinkListView(BaseView):
+    serializer_class = AuthenticatedLinkSerializer
+    filter_class = LinkFilter
+    search_fields = PublicLinkListView.search_fields + ('notes',)  # private links can also be searched by notes field
```
rebeccacremona: Hmmm, PublicLinkListView's tuple of search_fields above contains notes. Sounds like it shouldn't.

perma_web/api2/views.py
```diff
@@ -0,0 +1,483 @@
...
+    def post(self, request, format=None):
+        """ Create new link. """
+        data = request.data
+
+        # Set target folder, in order of preference:
+        # - 'folder' key in data
+        # - parent folder, if posting to /folders/:parent_id/archives
+        # - user's personal folder
+        folder = self.get_folder_from_request(request) or request.parent or request.user.root_folder
+
```
rebeccacremona: What would you think of a future feature that, if folder is specified via post data and if posted to /folders/:parent_id/archives but the specified folder is not a child of request.parent, throws a validation error?

perma_web/api2/views.py
```diff
@@ -0,0 +1,483 @@
...
+
+# /folders/:parent_id/archives/:guid
+class MoveLinkView(BaseView):
+    serializer_class = AuthenticatedLinkSerializer
+
+    @load_parent
+    def put(self, request, guid, format=None):
+        """
+            Move folder.
```
rebeccacremona: Typo, I think? Move link, rather than move folder?

perma_web/perma/settings/deployments/settings_common.py
```diff
@@ -465,3 +470,20 @@ def _pywb_static_dir_location():
 # If technical problems prevent proper analysis of a capture,
 # should we default to private?
 PRIVATE_LINKS_ON_FAILURE = False
+
+
+REST_FRAMEWORK = {
+    'DEFAULT_AUTHENTICATION_CLASSES': (
+        'api2.authentication.TokenAuthentication',  # authenticate with ApiKey token
+        'rest_framework.authentication.SessionAuthentication',  # authenticate with Django login
+    ),
```
rebeccacremona: This simple config is unacceptable. I demand we return to something in the style of https://github.com/harvard-lil/perma/blob/develop/perma_web/api/authentication.py. For pity's sake.

perma_web/api2/views.py
jcushman: Good idea -- combining into a single "Top-level folders" message, for both deleting and renaming.

perma_web/api2/views.py
jcushman: Nope! Good catch.

perma_web/api2/views.py
jcushman: So the question is: if I POST to /folders/:folder_id/archives with data {'url':'http://example.com', 'folder':other_folder_id}, do we write the archive to other_folder_id (as this code does), or to folder_id, or raise a validation error?

My feeling is it's fine to write to other_folder_id. I don't see why anyone would actually do that, but I can vaguely imagine composing a couple of helper functions and having it work out that way?

perma_web/api2/views.py
jcushman: Fixed.

PR #1964, opened 2017-04-07
Make datetimes TZ-aware to suppress test warnings

bensteinberg:
There must be a better way....


bensteinberg:
(Note that the alleged warnings don't appear until you test with Django 1.10.)

jcushman:
Would it be worth throwing something into perma.utils like

    def tz_datetime(*args, **kwargs):
        return timezone.make_aware(datetime(*args, **kwargs))

? I don't immediately see a cleaner approach, but maybe worth wrapping in a helper. Or if you're not feeling it I'm fine with this as is.

PR #1967, opened 2017-04-08
Pytest support

jcushman:
This lets you optionally use `pytest` as your test runner instead of the builtin `manage.py test` which is called by `fab test`. The primary benefit right now is the ability to run tests locally in parallel for speed. Pytest also seems like it has nicer output and offers some cool features we could explore in the future.

Usage:

To allow for parallel tests, go into `mysql -uroot -p` and run `GRANT ALL ON *.* TO 'perma'@'localhost';` so pytest can create multiple test databases (we should do this in the vagrant image if this ends up being useful).

All tests: `pytest`
One test: `pytest -k test_should_provide_different_folder_data_relative_to_user`
Tests for one app: `pytest api/`
Tests in two processes in parallel: `pytest -n 2`
Disable stdout capture if you want to set a breakpoint with pdb: `pytest -s`

If running in parallel, a couple of tests may fail until the interdependent-tests pull req is merged in.

Changes in this pull req:

- Merge config files -- combine .pep8, pytest config, and coverage config into one settings.cfg file
- Have TestHTTPServer try multiple ports so tests can run in parallel
- Make perma.settings.deployments.settings_testing a complete settings file that can be provided to test runners directly, to avoid the sys.argv import hack at the bottom of setting_dev.py

(So, all stuff that's good or neutral regardless of whether we end up wanting to use pytest.)

Speed differences:

I tried running the api tests on my local vagrant with two virtual CPUs in a few different ways:

`fab test:api2`: 200 seconds
`./manage.py test --settings perma.settings.deployments.settings_testing api2` -- same as `fab test` but without recording test coverage: 163 second
`pytest api2/`: 181 seconds
`pytest api2/ -n 2`: 116 seconds

So, ballpark, running in parallel takes about 2/3 as long as running consecutively.

I also tried running in 8 processes outside vagrant and got down to 90 seconds, but with some failures I didn't diagnose ...

### DIFF
--- a/perma_web/.pep8
+++ b/perma_web/.pep8
@@ -1,8 +1,0 @@
-[flake8]
-ignore = E12,
-         E2,W2,
-         E3,W3,
-         E4,
-         E501,
-         F403
-exclude = node_modules,static
--- a/perma_web/api/tests/utils.py
+++ b/perma_web/api/tests/utils.py
@@ -165,7 +165,14 @@ class ApiResourceTestCaseMixin(ResourceTestCaseMixin, SimpleTestCase):
             copy_file_or_dir(os.path.join(settings.PROJECT_ROOT, TEST_ASSETS_DIR, source_file), target_url)
 
         # start server
-        cls._httpd = TestHTTPServer(('', cls.server_port), TestHTTPRequestHandler)
+        for i in range(100):
+            try:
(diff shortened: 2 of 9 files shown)


rebeccacremona:
Zoom.

PR #1968, opened 2017-04-10
Minor upgrade of Celery in preparation for upgrade to Celery 4

bensteinberg:


bensteinberg:
See http://docs.celeryproject.org/en/latest/whatsnew-4.0.html#step-1-upgrade-to-celery-3-1-25

PR #1993, opened 2017-04-26
Domain special handling

rebeccacremona:
Introduces a mechanism for running special code during the capture process for particular domains. 

Delays captures of forbes.com for up to 24s, the apparent minimum time to bypass the splash page. (Made 10 sequential successful captures locally.)

Only relevant if 1992 is also merged.

### DIFF
--- a/perma_web/perma/tasks.py
+++ b/perma_web/perma/tasks.py
@@ -382,7 +382,8 @@ def proxy_capture(capture_job):
     # (Wrap in a lambda function to delay execution.)
     special_domains = {
         # Wait for splash page to auto redirect
-        "www.forbes.com": lambda: sleep_unless_seconds_passed(24)
+        "www.forbes.com": lambda: sleep_unless_seconds_passed(24),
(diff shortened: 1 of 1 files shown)


perma_web/perma/tasks.py
```diff
@@ -371,6 +371,21 @@ class HaltCaptureException(Exception):
...
+        if delta < seconds:
+            wait = seconds - delta
+            print("Sleeping for {}s".format(wait))
+            time.sleep(wait)
+
+    # By domain, code to run after the target_url's page onload event.
+    # (Wrap in a lambda function to delay execution.)
+    special_domains = {
+        # Wait for splash page to auto redirect
+        "www.forbes.com": lambda: sleep_unless_seconds_passed(24)
```
anastasia: We have just a few archives that are of "http://forbes.com", wondering if we should drop the "www"?

rebeccacremona: Ah, that would need a separate entry, since I'm matching on link.url_details.netloc.  +1 For some regex idea like Jack was proposing. Hmmm.

PR #1999, opened 2017-05-02
first pass at refactoring proxied_capture

rebeccacremona:
What do we think of this experiment? Trying lots of things all at once:

1) moving as much of the capture process up to the top level of tasks.py as possible, so that I can (soon) write unit test for individual functions, and (hopefully) make the story of the code easier to follow

2) if we plan to iterate over a set of elements, use javascript to grab and process them in one go, instead of using selenium to repeatedly access the DOM. Much faster!

3) try doing more of the capture process before the page's onload event, to improve the quality of captures that crash the browser before the page's onload event is fired (happens a lot w/nytimes

4) move the meat of the capture process into the run_next_capture function, so that we can cleanup better when SoftTimeoutException is raised by celery. This may or may not help solve the problem we've been having with truncated warcs (brought to our attention by LOCKSS errors): if the capture process was hanging during file I/O, and that's why it timed out, then this achieves nothing. But I'm not convinced that's what happened. I think the process was taking long for other reasons,  and that the timeout killed the writing of the warc to perma-mirror. But.... who's to say? (Not, so far as I can tell, our logs). In any case, maybe this will help us break the saving part of the job off to another celery process, one with permission to retry on failure.

5) wrap more of the capture process in code that checks to see if the browser is still running

6) more print statements to aid in debugging; an imperfect setting to run synchronously, as much as possible, to aid in debugging

For the pull request review: 
- does this actually make anything more readable/clearer? Can anybody suggest improvements/changes/more things to work on? 
- is there a better way to monitor the status of the browser?
- does it make sense to integrate the capture process into the celery task this way? I think I did the wrapped decorators correctly, but what's a good way to confirm the temp dir is working as expected?
- is my capture progress incrementing function okay? No hard feelings if it's too weird; I just am not convinced It should be a class, since it saves fields in a CaptureJob.

For the pull request or for future work:
- do all our values for how long to wait make sense? 10s seems like a long long time for repeat_while_exception or repeat_until_truthy. And are we using "implicit" wait correctly,
 (http://www.seleniumhq.org/docs/04_webdriver_advanced.jsp), in concert with repeat_while exception and other kinds of waiting we are doing?
- we have some combination of printing status messages and formally incrementing progress. I added the "x robots tag" step months ago, before I understood this code; that was a mistake. Let's spend some time thinking about what makes sense here.
- is there a way we can pull more of the code out into testable helper functions?

### DIFF
--- a/perma_web/perma/tasks.py
+++ b/perma_web/perma/tasks.py
@@ -11,30 +11,31 @@ from urllib2 import URLError
 import os
 import os.path
 import threading
+import Queue as queue
+import time
+from datetime import timedelta
 import urlparse
+import re
+import json
+import robotparser
+import errno
+import tempdir
+from socket import error as socket_error
 from celery import shared_task
 from celery.exceptions import SoftTimeLimitExceeded
-from pyvirtualdisplay import Display
-import re
-from requests.structures import CaseInsensitiveDict
 from selenium import webdriver
 from selenium.common.exceptions import WebDriverException, NoSuchElementException, NoSuchFrameException
 from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
 from selenium.webdriver.common.proxy import ProxyType, Proxy
-from datetime import timedelta
-import logging
-import json
-import robotparser
-import time
-import requests
-import errno
-import tempdir
-from socket import error as socket_error
-import internetarchive
-import Queue as queue
+from pyvirtualdisplay import Display
 from warcprox.controller import WarcproxController
 from warcprox.warcprox import WarcProxyHandler, WarcProxy, ProxyingRecorder
 from warcprox.warcwriter import WarcWriter, WarcWriterThread
+import requests
+from requests.structures import CaseInsensitiveDict
+from requests.packages.urllib3.exceptions import InsecureRequestWarning
+requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
+import internetarchive
 
 from django.core.files.storage import default_storage
 from django.template.defaultfilters import truncatechars
@@ -43,13 +44,11 @@ from django.utils import timezone
 from django.db.models import Q
 from django.http import HttpRequest
 
-
 from perma.models import WeekStats, MinuteStats, Registrar, LinkUser, Link, Organization, CDXLine, Capture, CaptureJob
 from perma.email import sync_cm_list, send_admin_email, registrar_users_plus_stats
-
-
 from perma.utils import run_task, url_in_allowed_ip_range, copy_file_data
 
+import logging
 logger = logging.getLogger(__name__)
 
 ### CONSTANTS ###
@@ -61,7 +60,87 @@ ELEMENT_DISCOVERY_TIMEOUT = 2 # seconds before PhantomJS gives up running a DOM 
 AFTER_LOAD_TIMEOUT = 30 # seconds to allow page to keep loading additional resources after onLoad event fires
 VALID_FAVICON_MIME_TYPES = {'image/png', 'image/gif', 'image/jpg', 'image/jpeg', 'image/x-icon', 'image/vnd.microsoft.icon', 'image/ico'}
 
-### HELPERS ###
+### THREAD HELPERS ###
+
+def run_capture_component(thread_list, func, **kwargs):
+    """
+       Run in a new thread, unless THREADED_CAPTURES is False (for debugging)
+    """
+    if settings.THREADED_CAPTURES:
+        add_thread(thread_list, func, **kwargs)
+    else:
+        func(**kwargs["kwargs"])
+
+def add_thread(thread_list, target, **kwargs):
+    if not isinstance(target, threading.Thread):
+        target = threading.Thread(target=target, **kwargs)
+    target.start()
+    thread_list.append(target)
+    return target
+
+def safe_save_fields(instance, **kwargs):
+    """
+        Update and save the given fields for a model instance.
+        Use update_fields so we won't step on changes to other fields made in another thread.
+    """
+    for key, val in kwargs.items():
+        setattr(instance, key, val)
+    instance.save(update_fields=kwargs.keys())
+
+def get_url(url, thread_list, proxy_address):
+    """
+        Get a url, via proxied python requests.get(), in a way that is interruptable from other threads.
+        Blocks calling thread. Perma assumes only run from sub-threads, and settings.THREADED_CAPTURES = True.
+        Unexpected blocking may result otherwise.
+    """
+    request_thread = add_thread(thread_list, ProxiedRequestThread(proxy_address, url))
+    request_thread.join()
+    return request_thread.response, request_thread.response_exception
+
+class ProxiedRequestThread(threading.Thread):
+    """
+        Run python request.get() in a thread, loading data in chunks.
+        Listen for self.stop to be set, allowing requests to be halted by other threads.
+        While the thread is running, see `self.pending_data` for how much has been downloaded so far.
+        Once the thread is done, see `self.response` and `self.response_exception` for the results.
+    """
+    def __init__(self, proxy_address, url, *args, **kwargs):
+        self.url = url
+        self.proxy_address = proxy_address
+        self.pending_data = 0
+        self.stop = threading.Event()
+        self.response = None
+        self.response_exception = None
+        super(ProxiedRequestThread, self).__init__(*args, **kwargs)
+
+    def run(self):
+        try:
+            self.response = requests.get(self.url,
+                                         headers={'User-Agent': settings.CAPTURE_USER_AGENT},
+                                         proxies={'http': 'http://' + self.proxy_address, 'https': 'http://' + self.proxy_address},
+                                         verify=False,
+                                         stream=True,
+                                         timeout=1)
+            self.response._content = bytes()
+            for chunk in self.response.iter_content(8192):
+                self.pending_data += len(chunk)
+                self.response._content += chunk
+                if self.stop.is_set():
+                    return
+        except requests.RequestException as e:
+            self.response_exception = e
+        finally:
+            self.pending_data = 0
+
+class HaltCaptureException(Exception):
+    """
+        An exception we can trigger to halt capture and release
+        all involved resources.
+    """
+    pass
+
+
+# WARCPROX HELPERS
 
 # monkeypatch ProxyingRecorder to grab headers of proxied response
 _orig_update_payload_digest = ProxyingRecorder._update_payload_digest
@@ -73,6 +152,9 @@ def _update_payload_digest(self, hunk):
         self.headers = re.sub(br'(\r?\n\r?\n).*', r'\1', self.headers)  # remove any part of hunk that came after headers
     return _orig_update_payload_digest(self, hunk)
 ProxyingRecorder._update_payload_digest = _update_payload_digest
+
+
+# BROWSER HELPERS
 
 def get_browser(user_agent, proxy_address, cert_path):
     """ Set up a Selenium browser with given user agent, proxy and SSL cert. """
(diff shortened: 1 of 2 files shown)


perma_web/perma/tasks.py
```diff
@@ -61,7 +60,87 @@
...
+    def run(self):
+        try:
+            self.response = requests.get(self.url,
+                                         headers={'User-Agent': settings.CAPTURE_USER_AGENT},
+                                         proxies={'http': 'http://' + self.proxy_address, 'https': 'http://' + self.proxy_address},
+                                         verify=False,
+                                         stream=True,
+                                         timeout=1)
+            self.response._content = bytes()
+            for chunk in self.response.iter_content(8192):
```
anastasia: nitpick - maybe should be `chunk_size=8192` for clarity?

perma_web/perma/tasks.py
```diff
@@ -265,35 +386,316 @@ def makefile(self, *args, **kwargs):
...
+                    darchive = True
+    return darchive
 
-    # reset
-    browser.implicitly_wait(ELEMENT_DISCOVERY_TIMEOUT)
-    browser.switch_to.default_content()
+# meta tags
 
-    return output_collector
+def meta_thread(browser=None, link=None):
```
anastasia: nitpick/question - is there a reason why browser and link are set to None as default here?

rebeccacremona: Having a late night memory that one can only pass kwargs into the thread-starting function, not positional args: did this so they'd be named. I will verify if that's right, or a relic of late night reasoning.

perma_web/perma/tasks.py
```diff
@@ -302,586 +704,331 @@ def warn_on_exception(message="Exception in block:", exception_type=Exception):
...
+
+        # basic setup
+        link = capture_job.link
+        target_url = link.safe_url
+        browser = warcprox_controller = warcprox_thread = display = None
+        have_content = have_warc = have_html = False
+        thread_list = []
+        successful_favicon_urls = []
+
+        def inc_progress(step_count, step_description):
```
rebeccacremona: Per discussion w/ Anastasia, this should probably be a method on CaptureJob instead.

perma_web/perma/tasks.py
```diff
@@ -302,586 +704,331 @@ def warn_on_exception(message="Exception in block:", exception_type=Exception):
...
-                screenshot_data = browser.get_screenshot_as_png()
-                link.screenshot_capture.write_warc_resource_record(screenshot_data)
-                save_fields(link.screenshot_capture, status='success')
-            else:
-                print "Not saving screenshots! %s" % ("Page size is %s pixels." % pixel_count if pixel_count else "")
-                save_fields(link.screenshot_capture, status='failed')
-        else:
-            # no screenshot if not HTML
-            save_fields(link.screenshot_capture, status='failed')
+        except HaltCaptureException:
```
jcushman: Is this nested try-except useful? Could just combine this except with the following except blocks and unindent a bunch of stuff.

perma_web/perma/tasks.py
```diff
@@ -302,586 +704,331 @@ def warn_on_exception(message="Exception in block:", exception_type=Exception):
...
+            print("HaltCaptureException thrown")
+    except SoftTimeLimitExceeded:
+        capture_job.link.tags.add('timeout-failure')
```
jcushman: Can just be `link.tags.add('timeout-failure')`

perma_web/perma/tasks.py
```diff
@@ -302,586 +704,331 @@ def warn_on_exception(message="Exception in block:", exception_type=Exception):
...
-            ).save()
-            print "Saved favicon at %s" % successful_favicon_urls
+            print "%s capture succeeded." % link.guid
 
-        print "%s capture succeeded." % link.guid
+        except:
+            print "Exception while processing capture job %s:" % capture_job.link_id
+            traceback.print_exc()
+        finally:
+            capture_job.link.captures.filter(status='pending').update(status='failed')
```
jcushman: Can refer to link rather than capture_job.link

perma_web/perma/tasks.py
```diff
@@ -302,586 +704,331 @@ def warn_on_exception(message="Exception in block:", exception_type=Exception):
...
-                save_fields(link.screenshot_capture, status='success')
-            else:
-                print "Not saving screenshots! %s" % ("Page size is %s pixels." % pixel_count if pixel_count else "")
-                save_fields(link.screenshot_capture, status='failed')
-        else:
-            # no screenshot if not HTML
-            save_fields(link.screenshot_capture, status='failed')
+        except HaltCaptureException:
+            print("HaltCaptureException thrown")
+    except SoftTimeLimitExceeded:
```
jcushman: Just a #TODO note that from here on we still have the problem that our current code has throughout, where exceeding the soft time limit needlessly results in a failed capture. I think the right answer is to disable the soft time limit once we reach this point, which is some celery dark magic I can't immediately crack.

rebeccacremona:
Thanks Anastasia and Jack! I will address everything and then reopen afterwards!

perma_web/perma/tasks.py
```diff
@@ -61,7 +60,87 @@
...
+    """
+    for key, val in kwargs.items():
+        setattr(instance, key, val)
+    instance.save(update_fields=kwargs.keys())
+
+def get_url(url, thread_list, proxy_address):
+    """
+        Get a url, via proxied python requests.get(), in a way that is interruptable from other threads.
+        Blocks calling thread. Perma assumes only run from sub-threads, and settings.THREADED_CAPTURES = True.
+        Unexpected blocking may result otherwise.
```
rebeccacremona: Yes.

I was originally hoping to wrap the get requests for favicons and robots.txt so that they could done synchronously, but it got pretty ugly. So I gave up on that, at least temporarily. With THREADED_CAPTURES = False, this function is used for favicons and robots.txt, but the main thread is blocked, so things at least happen one at a time. Given that -that- didn't work out, and given that the media tags requests are still 100% asynchronous regardless of how THREADED_CAPTURES is set.... I think I'll remove all that from the pull request, and save that for another day. 

Re: the comment. The original comment for this function said, "Helper function to get a url, via proxied requests.get(), in a way that is interruptable from other threads. This should only be run from sub-threads." I was trying to alter that comment to explicitly state that, if called from the main thread or if THREADED_CAPTURES = False for debugging, it will block.

perma_web/perma/tasks.py
rebeccacremona: Late night reasoning. I misread the docs ("This constructor should always be called with keyword arguments.") One can pass in args as a named argument, pointing to a tuple. E.g. threading.thread(args=(browser,link))

perma_web/perma/tasks.py
```diff
@@ -265,35 +386,316 @@ def makefile(self, *args, **kwargs):
...
+                var meta_tags = document.getElementsByTagName('meta');
+                var tags = [];
+                for (var i = 0; i < meta_tags.length; i++){
+                    tags.push({"name":meta_tags[i].name, "content":meta_tags[i].content});
+                }
+                return tags
+            """)
+    return repeat_until_truthy(get_meta)
+
+def meta_tags_process(meta_tags, link):
```
rebeccacremona: Sure! So long as we are comfortable collapsing duplicate names. Since duplicates aren't allowed for the meta tags we are currently looking at, seems totally reasonable. Why go with the first instead of the last though? Any reason to think an improperly coded page would intend for us to respect the first, rather than the last, robots declaration, etc.?

rebeccacremona: So much cleaner! How would you feel about leaving out 'reversed'?

jcushman: Sold, going with the last tag instead of first seems fine.

My code would barf on malformed tags that lack a .name, btw -- maybe throw in a check for that.

rebeccacremona: There are lots of meta tags that don't properly have a "name" attr, like `<meta charset="utf-8">`,  `<meta http-equiv="X-UA-Compatible" content="IE=edge">`, or  `<meta property="og:image" content="/assets/images/header-splash-logo-ani@1x.gif">`, but so long as we don't need them, it works:
```
var meta_tags = document.getElementsByTagName('meta');
var tags = {};
for (var i = 0; i < meta_tags.length; i++){
    tags[meta_tags[i].name.toLowerCase()] = meta_tags[i].content
}
```
All tags without a "name" attr get added with a key of "". 
```
{
  "": "NYSE NYT",
  "viewport": "width=device-width, initial-scale=1, maximum-scale=1",
  "sourceapp": "nyt-v5",
  "applicationname": "article",
  "foundation-build-id": "",
  "twitter:app:name:googleplay": "NYTimes",
  "twitter:app:id:googleplay": "com.nytimes.android",
  "twitter:app:url:googleplay": "nytimes://reader/id/100000005072858",
  "robots": "noarchive",
...
}
```
Grotesque, but apparently not disallowed, either in js or in python!

rebeccacremona: Cleaner code. More complex comment lol.
```
def meta_tags_get(browser, link):
    """
        Retrieves meta tags as a dict (e.g. {"robots": "noarchive"}).

        The keys of the dict are the "name" attributes of the meta tags (if
        any) and the values are the corresponding "content" attributes.
        Later-encountered tags overwrite earlier-encountered tags, if a
        "name" attribute is duplicated. Tags without name attributes are 
        thrown away (their "content" attribute is mapped to the key "", 
        the empty string).

        Uses js instead of selenium for speed.
    """
    def get_meta():
        with browser_running(link, browser, lambda: meta_tag_analysis_failed(link)):
            return browser.execute_script("""
                var meta_tags = document.getElementsByTagName('meta');
                var tags = {};
                for (var i = 0; i < meta_tags.length; i++){
                    tags[meta_tags[i].name.toLowerCase()] = meta_tags[i].content
                }
                return tags
            """)
    return repeat_until_truthy(get_meta)

def meta_tags_process(meta_tags, link):
    """Do what we want with the meta tags we find."""

    ## Privacy Related ##
    meta_tag = meta_tags.get('perma')
    if not meta_tag:
        meta_tag = meta_tags.get('robots')
    if meta_tag == 'noarchive':
        safe_save_fields(link, is_private=True, private_reason='policy')
        print "Meta found, darchiving"

    ## Page Description ##
    description_meta_tag = meta_tags.get('description')
    if description_meta_tag:
        safe_save_fields(link, submitted_description=description_meta_tag)
```

perma_web/perma/tasks.py
```diff
@@ -302,586 +704,331 @@ def warn_on_exception(message="Exception in block:", exception_type=Exception):
...
+            print "%s step %s: %s" % (link.guid, inc_progress.step, step_description)
+
+        # By domain, code to run after the target_url's page onload event.
+        # (Wrap in a lambda function to delay execution.)
+        special_domains = {
+            # Wait for splash page to auto redirect
+            "www.forbes.com": lambda: sleep_unless_seconds_passed(24, start_time),
+            "forbes.com": lambda: sleep_unless_seconds_passed(24, start_time)
+        }
+        post_load_function = special_domains.get(link.url_details.netloc)
```
rebeccacremona: Ooh, interesting....... not sure. Maybe it should be both? For instance, if somebody perma'ed a tiny.cc link, content_url would be right. Are there any circumstances where the target_url would be normal, and the content_url would be something weird, like a cdn or....

perma_web/perma/tasks.py
```diff
@@ -302,586 +704,331 @@ def warn_on_exception(message="Exception in block:", exception_type=Exception):
...
+            })
+
+            inc_progress(1, "Checking x-robots-tag directives.")
+            if xrobots_blacklists_perma(robots_directives):
+                safe_save_fields(link, is_private=True, private_reason='policy')
+                print "x-robots-tag found, darchiving"
 
-                successful_favicon_urls.append((favicon_url, mime_type))
+            if have_html:
+                # DO EVERYTHING THAT'S SAFE TO DO BEFORE "ONLOAD" IS FIRED
```
rebeccacremona: (A) Yes.... this division and accompanying comments are somewhat mendacious. Re-examining true motivations: doing meta tags here means we very reliably get NYtimes meta tags, and doing title after onload means we very reliably get accurate titles for forbes articles, rather than the title of their splash screen. But, of course, that means we are at least sometimes analyzing the meta tags on the forbes welcome page, rather than those of the Forbes article. The favicons, on the other hand, again, reanalyzing motivations.... don't matter that much, so it felt "safe" to put that thread here, to see if we could shave a few seconds off the capture process. But that's not necessarily a great decision either.

For title, description, and meta tags, the right thing to do is definitely to grab the html both immediately and after onload, and maybe move the analysis into the "finally" section, since its so zippy, so that it runs regardless of whether the browser has crashed. I'll see what that looks like.

Favicons..... I still think doing it here is fine, but I don't feel strongly about it.

B) Yes: Since we know that we've received a complete, contentful response already at this point, Selenium's implicit wait (set to a generous 2 seconds by browser.implicitly_wait(ELEMENT_DISCOVERY_TIMEOUT)) should more than take care of it. (I just double checked that it waits for execute_script and other commands, not just the explicit element finding commands.) In the case of meta tags, we also have repeat_until_truthy going, just in case.

perma_web/perma/tasks.py
```diff
@@ -302,586 +704,331 @@ def warn_on_exception(message="Exception in block:", exception_type=Exception):
...
+                        "successful_favicon_urls": successful_favicon_urls,
+                        "browser": browser,
+                        "content_url": content_url,
+                        "thread_list": thread_list,
+                        "proxy_address": proxy_address
+                    })
+
+                # WAIT FOR "ONLOAD" FOR EVERYTHING ELSE
+                print("Waiting for onload event before proceding.")
+                page_load_thread.join(ONLOAD_EVENT_TIMEOUT)
```
rebeccacremona: Ah! I intended to do exactly that, and forgot. Thanks so much for catching.

perma_web/perma/tasks.py
rebeccacremona: Whoops.

perma_web/perma/tasks.py
rebeccacremona: Yes.... If the tempfile/memory situation allows, I'm wondering if the right thing to so is to pass the file I/O off to a separate task, one that celery can keep retrying if (for instance) some network situation temporarily makes it impossible to save to s3.

perma_web/perma/tasks.py
rebeccacremona: Thanks!

PR #2005, opened 2017-05-05
Remove tastypie

jcushman:
Here's a pass at removing the tastypie API and moving the DRF api to /v1/.

DEPLOYMENT NOTE: For any systems that have `API_VERSION=2` as an override, remember to remove so they go back to 1.


perma_web/api/tests/test_current_user_resource.py
```diff
@@ -34,7 +31,7 @@ def test_get_archives_json(self):
         self.successful_get(self.detail_url + 'archives/', user=self.regular_user, count=11)
 
     def test_get_folders_json(self):
-        self.successful_get(self.detail_url + 'folders/', user=self.org_user, count=1)
+        self.successful_get(self.detail_url + 'folders/', user=self.org_user, count=2)
```
rebeccacremona: Of all the tiny things to wonder about, when reviewing this..... when/where did org_user pick up a second folder? I'm not seeing it.

jcushman: See point 6 from the previous pull req (#1949): "/archives now includes all archives accessible_to the user, not those created_by the user. /folders now includes all top-level folders for the user, not all folders created_by the user. Rationale: this fixes a logic error in the v1 API. Since links and folders can be moved around and original creator can lose access to them, checking created_by made no sense."

rebeccacremona: Oh of course. Thanks, and apologies.

perma_web/perma/settings/deployments/settings_common.py
```diff
@@ -157,7 +157,6 @@ def _pywb_static_dir_location():
```
rebeccacremona: Do we still need/want this part?

  387: API_VERSION = 1
  388  
  389  TEMPLATE_VISIBLE_SETTINGS = (
  390:     'API_VERSION',

rebeccacremona: (We'll want to make sure that variable is set to v1 on all the various tiers too.)

rebeccacremona: I think it would be prudent to get rid of it, delete the variable from settings on all the servers, and get rid of it in this template:

~perma_upstream/perma_web/perma/templates/js_config.html:
```
    1  <script>
    2    var settings = {
    3:     API_VERSION: {{ API_VERSION }},
    4      STATIC_URL: "{{ STATIC_URL }}",
    5      MEDIA_URL: "{{ MEDIA_URL }}",
    .
    7      DEBUG: {{ DEBUG|lower }}
    8    },
    9:   api_path = "/api/v" + settings.API_VERSION
   10  </script>
```
What do you think?

rebeccacremona:
Thanks so much for doing this! Looks great! With the possible exception of the API_VERSION variable and associated code (see other comments), I can't think of anything else to check or change.

Let's perhaps discuss a prudent way to deploy. Maybe we make a couple users w/ api keys on stage, push to stage and make sure their api keys still work, and then create another new user with a new api, and confirm that works?

perma_web/perma/settings/deployments/settings_common.py
rebeccacremona: (Actually, from your deployment note, I surmise you think it's better to keep it in there, to make future changes easier?)

PR #2011, opened 2017-05-10
Less logging

rebeccacremona:


rebeccacremona:
FYI: I'm seeing something slightly weird. So, if you play back a capture of an NYTimes article, you should get lots of 404s to play with. All the ones whose initiator is us/ the page (rather than wombat.js) should get the perma link not found page as the response. Sometimes, that appears to be getting truncated, at all different parts of the template, but not consistently, and not repeatedly.... and I can't for the life of me figure out why. I'm not too worried about it: could just be the dev server being wonky.

PR #2014, opened 2017-05-11
Improve Chrome support

jcushman:
This improves our support for setting `CAPTURE_BROWSER = 'Chrome'`, primarily by allowing Chrome to take full-page screenshots.

This pull includes some refactoring but should not substantively affect PhantomJS captures, except for one change -- the page size for screenshots is based on measuring the size of `<html>` instead of `<body>`, since the former can be larger than the latter.

This does not yet take advantage of Chrome's new `--headless` flag. It's just using Selenium's longtime support for controlling desktop browsers along with a virtual display. `--headless` with Selenium/ChromeDriver doesn't seem to work reliably yet. (As of Chrome 59/ChromeDriver 2.29 -- next stable release of both might be good.) But I think that's just a two-line switch once `--headless` is ready, and the only impact of that change would be to make everything more efficient by not routing through a virtual display.

The big change here and motivation for most of the refactoring is to let Chrome take full-page screenshots by resizing the browser window to the size of the page before taking the screenshot. This is the recommended technique in https://developers.google.com/web/updates/2017/04/headless-chrome , so although it feels janky it's apparently what's endorsed by the Chrome team.

That makes it more important to accurately measure the size of the page before taking the screenshot -- if the measurement is too small the screenshot cuts off part of the page. For example when we were measuring the size of `<body>` instead of `<html>`, that broke even for example.com. Right now the size is close but still cuts off the footer on nytimes.com. 

To install dependencies and try out in Vagrant:

```
# install Chrome:
sudo apt-get install libxss1 libappindicator1 libindicator7
wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb # change this from stable to beta if you want to try out Chrome 59, which the Chrome team blog post is about -- also requires changing a line in tasks.py to use google-chrome-beta binary
sudo dpkg -i google-chrome*.deb
sudo apt-get install -f # fixes the missing-dependency errors from previous line

# install ChromeDriver:
sudo apt-get install unzip
wget -N http://chromedriver.storage.googleapis.com/2.29/chromedriver_linux64.zip
unzip chromedriver_linux64.zip
chmod +x chromedriver
sudo mv -f chromedriver /usr/local/share/chromedriver
sudo ln -s /usr/local/share/chromedriver /usr/local/bin/chromedriver
```

### DIFF
--- a/perma_web/perma/tasks.py
+++ b/perma_web/perma/tasks.py
@@ -5,7 +5,7 @@ import traceback
 from cStringIO import StringIO
 from contextlib import contextmanager
 
-from httplib import HTTPResponse
+from httplib import HTTPResponse, HTTPException
 from urllib2 import URLError
 
 import os
@@ -59,6 +59,8 @@ RESOURCE_LOAD_TIMEOUT = 45 # seconds to wait for at least one resource to load b
 ELEMENT_DISCOVERY_TIMEOUT = 2 # seconds before PhantomJS gives up running a DOM request (should be instant, assuming page is loaded)
 AFTER_LOAD_TIMEOUT = 30 # seconds to allow page to keep loading additional resources after onLoad event fires
 VALID_FAVICON_MIME_TYPES = {'image/png', 'image/gif', 'image/jpg', 'image/jpeg', 'image/x-icon', 'image/vnd.microsoft.icon', 'image/ico'}
+BROWSER_SIZE = [1024, 800]
+
 
 ### THREAD HELPERS ###
 
@@ -146,8 +148,17 @@ ProxyingRecorder._update_payload_digest = _update_payload_digest
 
 # BROWSER HELPERS
 
+def start_virtual_display():
+    display = Display(visible=0, size=BROWSER_SIZE)
+    display.start()
+    return display
+
 def get_browser(user_agent, proxy_address, cert_path):
     """ Set up a Selenium browser with given user agent, proxy and SSL cert. """
+
+    display = None
+    print "Using browser: %s" % settings.CAPTURE_BROWSER
+
     # PhantomJS
     if settings.CAPTURE_BROWSER == 'PhantomJS':
         desired_capabilities = dict(DesiredCapabilities.PHANTOMJS)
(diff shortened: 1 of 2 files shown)


perma_web/perma/tasks.py
```diff
@@ -440,7 +462,7 @@ def get_meta():
                 }
                 return tags
             """)
-    return repeat_until_truthy(get_meta)
+    return repeat_while_exception(lambda: repeat_until_truthy(get_meta), exception=HTTPException)
```
rebeccacremona: Aha, thanks for including these fixes too. I still haven't tested to see what happens w/o repeat_until_truthy; will do shortly. Won't delay merging this to wait for that!

perma_web/perma/tasks.py
```diff
@@ -59,6 +59,8 @@
 ELEMENT_DISCOVERY_TIMEOUT = 2 # seconds before PhantomJS gives up running a DOM request (should be instant, assuming page is loaded)
 AFTER_LOAD_TIMEOUT = 30 # seconds to allow page to keep loading additional resources after onLoad event fires
 VALID_FAVICON_MIME_TYPES = {'image/png', 'image/gif', 'image/jpg', 'image/jpeg', 'image/x-icon', 'image/vnd.microsoft.icon', 'image/ico'}
+BROWSER_SIZE = [1024, 800]
+
```
rebeccacremona: Just curious: any insights into how these dimensions were selected, whenever they were selected?

rebeccacremona:
Hooray for Chrome!

The capture of nytimes articles and forbes articles is playing out REALLY differently via phantom js and via Chrome. Chrome is timing out a LOT in the repeat_while_exception functions..... and at 10s a pop. I've run a bunch of captures with the new code, but with phantom js, and they seem normal, so I do think it's a difference in Chrome, rather than the difference in how repeat_while_exception functions are called in this request. So! Merging in. But, once the headless version is available, let's dig in and figure out what's going on here.

But, if we do end up moving to the headless version of Chrome shortly, let's definitely

PR #2015, opened 2017-05-12
Extract page contents by parsing DOM.

jcushman:
The primary change here is to extract page contents by fetching the whole DOM and inspecting with PyQuery, rather than using multiple calls to Selenium to inspect individual elements.

The general approach now becomes:

- `dom_tree = get_dom_tree(browser)`, which is a slow function that calls out to `browser.execute_script` and returns a parsed PyQuery object.
- Inspect `dom_tree`, which is basically free.

This doesn't seem to be much faster (or slower) in practice, but it simplifies things in a few different ways:

- Use python instead of javascript to extract data, without loss of speed.
- Lots fewer calls to `browser` makes it easier to reason about performance -- slow steps are isolated to `get_dom_tree`.
- Lots fewer calls to `browser` will make it easier to migrate to raw Chrome connection if we want to do that later.

There are two other bits of refactoring in this pull:

- Instead of adding `link.tags.add('browser-crashed')` in `browser_running`, add it at the end in `teardown`. This avoids passing `link` around to a bunch of places, and catches possible case where the browser crashes after the last time `browser_running` is run.
- Rename `proxied_urls` to `requested_urls`, and have it include urls requested through `ProxiedRequestThread` as well as through warcprox. Avoids some possibility of redundant requests.

### DIFF
--- a/perma_web/perma/tasks.py
+++ b/perma_web/perma/tasks.py
@@ -3,9 +3,11 @@ from __future__ import absolute_import # to avoid importing local .celery instea
 import tempfile
 import traceback
 from cStringIO import StringIO
+from collections import OrderedDict
 from contextlib import contextmanager
+from pyquery import PyQuery
 
-from httplib import HTTPResponse, HTTPException
+from httplib import HTTPResponse
 from urllib2 import URLError
 
 import os
@@ -34,6 +36,7 @@ from warcprox.warcwriter import WarcWriter, WarcWriterThread
 import requests
 from requests.structures import CaseInsensitiveDict
 from requests.packages.urllib3.exceptions import InsecureRequestWarning
+
 requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
 import internetarchive
 
@@ -47,6 +50,7 @@ from django.http import HttpRequest
 from perma.models import WeekStats, MinuteStats, Registrar, LinkUser, Link, Organization, CDXLine, Capture, CaptureJob
 from perma.email import sync_cm_list, send_admin_email, registrar_users_plus_stats
 from perma.utils import run_task, url_in_allowed_ip_range, copy_file_data
+from perma import site_scripts
 
 import logging
 logger = logging.getLogger(__name__)
@@ -80,12 +84,12 @@ def safe_save_fields(instance, **kwargs):
         setattr(instance, key, val)
     instance.save(update_fields=kwargs.keys())
 
-def get_url(url, thread_list, proxy_address):
+def get_url(url, thread_list, proxy_address, requested_urls):
     """
         Get a url, via proxied python requests.get(), in a way that is interruptable from other threads.
         Blocks calling thread. (Recommended: only call in sub-threads.)
     """
-    request_thread = add_thread(thread_list, ProxiedRequestThread(proxy_address, url))
+    request_thread = add_thread(thread_list, ProxiedRequestThread(proxy_address, url, requested_urls))
     request_thread.join()
     return request_thread.response, request_thread.response_exception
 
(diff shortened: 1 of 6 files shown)


perma_web/perma/tasks.py
```diff
@@ -266,32 +270,56 @@ def scroll_browser(browser):
...
+        Return page source as a parsed PyQuery object for querying.
+         
+        PyQuery here works the same as `$(source)` in jQuery. So for example you can do `parsed_source(selector)` 
+        with the returned value to get a list of LXML elements matching the selector.
+    """
+    return PyQuery(source, parser='html')
+
+def get_dom_tree(browser):
+    with browser_running(browser):
+        return parse_page_source(get_page_source(browser))
```
rebeccacremona: Your description of the pull request expresses concern this is slow: on 15 captures of a particular NYTimes, the longest runtime was 0.4s (post-onload), and the shortest was 0.02s (pre-onload). (Very few requests made it to the post-onload event before crashing). On 5 captures of particular Forbes article, the longest runtime was 0.1s (post-onload), and the shortest was 0.02s (pre-onload)) That seems plenty fast for government work ;-)

perma_web/perma/tasks.py
```diff
@@ -940,26 +918,23 @@ def _proxy_request(self):
...
             # Get a copy of the page's metadata immediately, without
             # waiting for the page's onload event (which can take a
             # long time, and might even crash the browser)
-            with browser_running(link, browser):
-                print "Retrieving page metadata (first time)"
-                add_thread(thread_list, get_metadata, args=(
-                    page_metadata,
-                    link,
-                    browser
-                ))
```
rebeccacremona: Super nitpick: may we have a print statement here, something like "Retrieving DOM (pre-onload)" or similar, and a post-load equivalent the next time this runs? I like seeing where in the capture process the browser crashes.

perma_web/perma/tasks.py
```diff
@@ -973,33 +948,29 @@ def _proxy_request(self):
...
             # Get a fresh copy of the page's metadata, if possible
-            with browser_running(link, browser):
-                inc_progress(capture_job, 1, "Retrieving page metadata (after onload)")
-                add_thread(thread_list, get_metadata, args=(
-                    page_metadata,
-                    link,
-                    browser
-                ))
+            dom_tree = get_dom_tree(browser)
+            get_metadata(page_metadata, dom_tree)
```
rebeccacremona: Intriguingly, this is grabbing the DOM of the Forbes welcome page the majority of the time. Maybe after all these changes, the 25s delay needs to be extended by a second or two.....

rebeccacremona: What would you think of waiting until after the scroll for this second load? I wonder if on average that would that make the captured DOM even better, or would just mean the capture process fails out more often before we get to it....

rebeccacremona:
Hmmm, also just saw that the lil site's title is captured as "The Library Innovation Lab at Harvard University | Library Innovation Lab nav-logo". 
```
<title>The Library Innovation Lab at Harvard University | Library Innovation Lab</title>
```

perma_web/requirements.txt
```diff
@@ -68,6 +69,7 @@ jsmin==2.0.9
 jsonpatch==0.4            # via internetarchive
 kombu==3.0.37             # via celery
 LinkHeader==0.4.3
+lxml==3.7.3               # via pyquery
```
jcushman: This rings a bell for me as well, but the lxml package appears to compile and run correctly on perma prod, so I think we're good.

perma_web/perma/tasks.py
jcushman: Thanks for checking that! Good to know it's not long relative to our total capture time.

As background I guess I mean it's slow on some sort of order-of-magnitude programmer intuition scale. Like if we rank things by order of magnitude runtime ...

typical python function < Redis access < typical database query ~ disk access < S3 access ~ typical remote API access

... then you have to fret exponentially more about how frequently each of those is used in a given algorithm as you move to the right -- pretty much no loops allowed after the second `<`, or it creates a human-noticeable delay. `get_dom_tree` is all the way on the right on this scale -- but that's a win, because it replaces multiple functions that were all the way on the right, and now those are over on the left at "typical python function."

rebeccacremona:
Whoops! Now there's a merge conflict. Sorry!

jcushman:
No worries! Will rebase in the next push.

jcushman:
Re: `<title>`, great catch! Fixed the query to grab just 'head > title', and added a test that would catch the mistake.

perma_web/perma/tasks.py
jcushman: Added some print statements.

perma_web/perma/tasks.py
rebeccacremona: Aha! Thanks makes sense.

perma_web/perma/tasks.py
jcushman: I took a pass at a Forbes handler that attempts to click and then waits for the redirect rather than a fixed timeout. Since it's a bit longer, I moved it to a separate file.

It's basically free to add a check after the scroll, so we can do both for now -- I threw in an extra one. In the long run we should just stop the browser from crashing ...

PR #2016, opened 2017-05-12
Django 1.10

jcushman:
I think it's time.

### DIFF
--- a/perma_web/requirements.in
+++ b/perma_web/requirements.in
@@ -7,7 +7,7 @@
 
 # Package:                      # Used for:
 
-Django==1.9.13
+Django==1.10.7
 MySQL-python==1.2.5
 celery==3.1.25
 requests[security]==2.11.1
--- a/perma_web/requirements.txt
+++ b/perma_web/requirements.txt
@@ -39,7 +39,7 @@ django-simple-history==1.8.1
 django-sslserver==0.14
 django-storages==1.1.8
 django-taggit==0.20.2
-Django==1.9.13            # via django-admin-smoke-tests, django-model-utils, django-sslserver
+Django==1.10.7            # via django-admin-smoke-tests, django-model-utils, django-sslserver
 django_webpack_loader==0.3.3
 djangorestframework==3.6.2
 docopt==0.6.2             # via internetarchive


rebeccacremona:
Looks good to me too!! Except for one thing: for logging to work the way it used to, locally, for people like me (who use local log files), we need to add this to settings_dev.py
```
# per https://docs.djangoproject.com/en/1.11/releases/1.10/#runserver-output-goes-through-logging
LOGGING['formatters']['django.server'] = {
    '()': 'django.utils.log.ServerFormatter',
    'format': '[%(server_time)s] %(message)s',
}
LOGGING['handlers']['django.server'] = {
    'level': 'INFO',
    'class': 'logging.StreamHandler',
    'formatter': 'django.server',
}
LOGGING['loggers']['django.server'] = {
    'handlers': ['django.server'],
    'level': 'INFO',
    'propagate': False,
}
```

rebeccacremona:
Should that be part of the repo, or should I just add it to my settings.py? I'm probably the only one using local log files.

jcushman:
What does using local log files mean? Are you adding a setting to settings.py that breaks under 1.10? I'd be curious to see it -- could make sense to pull stuff into settings_dev.py if you're finding it helpful.

rebeccacremona:
With 1.9, I sometimes run with `LOGGING['handlers']['default']['filters'] = []` in settings.py, so that I have a record of stack traces that are happening during capture inside /perma-rlc/services/logs/django.log, so I can more easily look for commonalities, and so I can see a log a little closer to our production log than what prints to the terminal. They are pretty verbose, so you have to clean up every so often. That breaks with 1.10, as per their release notes. I can get the same behavior using:
```
LOGGING['handlers']['default']['filters'] = []
# per https://docs.djangoproject.com/en/1.11/releases/1.10/#runserver-output-goes-through-logging
LOGGING['formatters']['django.server'] = {
    '()': 'django.utils.log.ServerFormatter',
    'format': '[%(server_time)s] %(message)s',
}
LOGGING['handlers']['django.server'] = {
    'level': 'INFO',
    'class': 'logging.StreamHandler',
    'formatter': 'django.server',
}
LOGGING['loggers']['django.server'] = {
    'handlers': ['django.server'],
    'level': 'INFO',
    'propagate': False,
}
```

PR #2017, opened 2017-05-13
Improve run_in_frames_recursive function

jcushman:
This updates the run_in_frames_recursive function to switch to frames by numeric index rather than fetching and using WebElement references. The primary advantage is we don't get expired WebElement references on pages that keep rewriting their HTML. Also saves a couple of find_element calls per recursion.

Other improvements while refactoring:

- Recursion is now limited to depth of 3, and 20 iframes visited overall.
- Check for non-http iframes and skip.
- Avoid 1 redundant tree traversal in each recursion, by putting traversal at the end of the loop instead of the beginning.

On the testing side, this pull consolidates two tests that were hitting the same nested-frame endpoint, and expands that test to look for media elements across a more complicated iframe hierarchy.

Unrelatedly (except it came up while debugging), this pull req goes ahead and swallows warcprox errors in TrackingRequestHandler, since they mess with the log and don't seem to help anything.

### DIFF
--- a/perma_web/api/tests/test_link_resource.py
+++ b/perma_web/api/tests/test_link_resource.py
@@ -252,21 +252,7 @@ class LinkResourceTestCase(ApiResourceTransactionTestCase):
         link = Link.objects.get(guid=obj['guid'])
         self.assertValidCapture(link.primary_capture)
 
-    def test_should_capture_all_srcset_images(self):
-        target_folder = self.org_user.root_folder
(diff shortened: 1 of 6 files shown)


jcushman:
Works for me! Now prints something like: "WarcProx exception: BadStatusLine parsing response from https://beacon.krxd.net/cookie2json?callback=Krux.ns._default.kxjsonp_3pevents"

PR #2043, opened 2017-05-25
fix - 500 error view

anastasia:


perma_web/urls.py
```diff
@@ -1,8 +1,11 @@
 from django.conf.urls import url, include
 from django.contrib import admin
+from django.conf.urls import *
```
rebeccacremona: Thanks so much for working on this!!!!

Curious, why do we need both `from django.conf.urls import url, include` and `from django.conf.urls import *`?

anastasia: because I thought I would be writing more than just a 500 error custom route, and then while working on it changed my mind (404 works fine, as far as I've tested it). I'll change it.

PR #2053, opened 2017-06-02
Collect library physical addresses; add geocoding to admin

jcushman:
This is mostly Becky's work; I just rebased on our current code and made a few tweaks. 

The primary goal here is to make Caitlin's work of adding library partners to our map a lot easier.

Effect of pull req:

- Add address field to registrars.
- Add button on django admin to set lat/lon from address.
- Add address field to library sign-up form. (Jack change: I commented out the logo and "show partner status" fields from Becky's code, because Adam thought that form was getting too heavy on the previous pass. I think adding just the address is pretty lightweight.)
- Format checkboxes in Django forms the bootstrap way. (Jack note: Becky had this as a special case, but I can't find anywhere that's impacted by it, and it seems like a good idea to do to all checkboxes. We can fix up anywhere we find rather than have to special case it.)


perma_web/perma/views/service.py
```diff
@@ -88,4 +92,17 @@ def bookmarklet_create(request):
     '''
     tocapture = request.GET.get('url', '')
     add_url = "{}?url={}".format(reverse('create_link'), tocapture)
-    return redirect(add_url)
\ No newline at end of file
+    return redirect(add_url)
+
+@login_required
+@user_passes_test(lambda user: user.is_staff or user.is_registrar_user())
```
rebeccacremona: Tiniest of suggestions: can we use perma.utils.user_passes_test_or_403 here instead? You can get into some funny redirect situations with the plain user_passes_test decorator.

PR #2060, opened 2017-06-21
Move cloudflare files; autoload into settings.py

jcushman:
Some bikeshedding of the Cloudflare IP system, since we had our first instance where the files actually changed ...

- Move cache files to services/cloudflare.
- Add fab task to update files.
- Add settings helper function to load Cloudflare IPs into settings.py.

With this change you can update settings.py on production to 

```
from .utils.helpers import get_cloudflare_ips
TRUSTED_PROXIES = [get_cloudflare_ips(CLOUDFLARE_DIR), ['<nginx server ip>']]
```

... so settings.py doesn't have to updated after changes.


rebeccacremona:
Slick.

How do you feel about the security of getting this off the web without manual review? I had been thinking of an automated approach when I initially started working on this problem (before you cleverly and successfully implemented it with custom middleware!), but was a bit itchy about the idea of loading app settings straight from some file we download from the web. Hence the manual update + automated test approach. Too big a pain/ too cautious, do you think?

rebeccacremona:
SORRY! Never mind, this still involves manually review! Brain not fully in gear today!

PR #2064, opened 2017-06-24
DB updates - mysql connection update and db reset fabric helper

phillipsm:
MySQL 5.7 (recent upgrade for me. lookout!) appears to require a new name for the storage engine in the config.

Also added a helper for a common task for me -- fabric helper to drop and create the two perma dbs


perma_web/fabfile/dev.py
```diff
@@ -120,6 +123,19 @@ def init_db():
...
+        Drops the perma and perma_cdxline databases and creates and inits them again
+        Let folks run this if they're not in Django's debug mode
+    """
+
+    if settings.DEBUG and not confirm("WARNING! You're about to drop the Perma.cc DBs. Continue anyway?"):
+        abort("No DBs dropped. Aborted.")
+
+    local("python manage.py sqlflush --database default | python manage.py dbshell")
+    local("python manage.py sqlflush --database perma-cdxline | python manage.py dbshell")
+    init_db()
```
jcushman: Chatted about this with Ben and Jay and I think it'll be super handy, but it needs a couple of tweaks:

- The logic for settings.DEBUG is backwards, I think. Can we make it refuse to run unless settings.DEBUG is true?
- Using `manage.py` in the `local()` command feels a little too powerful -- if you changed settings.py in your dev env to point to a real DB to test something out, or changed DEBUG=True on prod, and then forgot about it and ran this task a few days later, it could do unfortunate things without letting you know anything was wrong. Could we change to a low-level `local("mysql -uroot -p -e 'drop database blah; create database blah utf8;'")` instead? That way it'll definitely never be smart enough to delete anything anyone cares about.

perma_web/fabfile/dev.py
```diff
@@ -120,6 +123,19 @@ def init_db():
...
     local("python manage.py loaddata fixtures/sites.json fixtures/users.json fixtures/folders.json")
 
+@task
+def reset_hard_db():
+    """
+        Drops the perma and perma_cdxline databases and creates and inits them again
+        Let folks run this if they're not in Django's debug mode
+    """
+
+    if not settings.DEBUG and not confirm("WARNING! You're about to drop the Perma.cc DBs. Continue anyway?"):
```
rebeccacremona: Hmm. I think this switch is maybe backwards? "If not settings.DEBUG" means "if in production", right?

phillipsm:
Thanks for the notes, @jcushman. def had the debug.settings backward (!). thanks.

Added logic to chat directly with the mysql client (kinda breaks django's db agnostic business?). thanks again!

PR #2076, opened 2017-08-02
Update install.md

PJF16:
Added space between the parameter and the password - would have failed else


bensteinberg:
Thanks for this PR! We're wrangling some subtleties of our testing -- Travis is upgrading its default Linux: https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming -- hence the failed check. Stay tuned.

rebeccacremona:
Thanks again for catching this mistake! We're still wrestling with our automatic testing; your change is obviously perfect, and is not causing the failure. So, we're going to merge this in now despite the failed check. Thanks again, and congrats on your first PR!

PR #2080, opened 2017-08-03
Add robots.txt

bensteinberg:


perma_web/perma/views/common.py
```diff
@@ -383,3 +383,29 @@ def contact_thanks(request):
...
+
+
+def robots_txt(request):
+    """
+    robots.txt
+    """
+    from ..urls import urlpatterns
+
+    # redundant -- from models.py
+    guid_character_set = "23456789ABCDEFGHJKLMNPQRSTUVWXYZ"
```
jcushman: Awesome! My one suggestion is to put this in `Link.GUID_CHARACTER_SET` and refer to it from both places to avoid the redundancy.

PR #2096, opened 2017-08-11
Registrar requested by

### LINKED ISSUE #2072: new registrar acct requests: reply-to email should be requesting user
the email that comes to info@perma.cc from new registrar requests should be from the person who requested the account (requesting user), and not the the general email they give for the library. This will make replying to the email (in our helpdesk service) go to the right person, instead of the general inbox.

anastasia:
fixes #2072

### DIFF
--- a/perma_web/perma/views/user_management.py
+++ b/perma_web/perma/views/user_management.py
@@ -1484,6 +1484,12 @@ def email_registrar_request(request, pending_registrar):
     Send email to Perma.cc admins when a library requests an account
     """
     host = request.get_host()
+    try:
+        email = request.user.email
+    except AttributeError:
+        # User did not have an account
+        email = request.POST.get('a-email')
+
     send_admin_email(
         "Perma.cc new library registrar account request",
         pending_registrar.email,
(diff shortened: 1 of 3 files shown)


anastasia:
Sorry I think this is my mistake. It looks like, upon second reading, we want the "reply-to" field to be changed. This pull request simply adds more context (by including the email of the person that signed up). @bjohnson747 can you please tell me again what an optimal result would look like?

perma_web/perma/views/user_management.py
```diff
@@ -1484,6 +1484,12 @@ def email_registrar_request(request, pending_registrar):
...
     host = request.get_host()
+    try:
+        email = request.user.email
+    except AttributeError:
+        # User did not have an account
+        email = request.POST.get('a-email')
+
     send_admin_email(
         "Perma.cc new library registrar account request",
         pending_registrar.email,
```
rebeccacremona: I think if you switch this to "email" from pending_registrar.email, people will get all the benefit of the nice text you added to the email, AND the reply-to will be the user's email address.

PR #2110, opened 2017-08-25
Prepare to use Chrome

bensteinberg:
This PR allows capture using Chrome and chromedriver. Next up is to get them installed on staging and production, and of course to produce a new Vagrant dev box, but if you want to try this out locally, I used https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb (version 60.0.3112.113) and https://chromedriver.storage.googleapis.com/2.31/chromedriver_linux64.zip. Unzip the latter and put `chromedriver` in `/usr/local/bin`; then I did this, though there's probably a shorter critical path:

```
sudo apt-get install gconf-service
sudo apt-get -f install
sudo dpkg -i google-chrome-stable_current_amd64.deb
```
Finally, set `CAPTURE_BROWSER = 'Chrome'` in `settings.py`.


perma_web/perma/tasks.py
```diff
@@ -636,6 +634,9 @@ def get_srcset_image_urls(dom_tree):
             src = src.strip().split()[0]
             if src:
                 urls.append(src)
+    # get src, too: Chrome doesn't do this automatically, although phantomjs does
```
rebeccacremona: Should this behind a flag, ` if settings.CAPTURE_BROWSER == 'Chrome':` or ` if settings.CAPTURE_BROWSER != 'PhantomJS':` or similar? Maybe not sufficiently expensive to care, but maybe?

rebeccacremona:
Should https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/settings/deployments/settings_common.py#L406 also get changed, either the comment or the chrome version number or....?

bensteinberg:
I think I'd like to leave `CAPTURE_USER_AGENT` as is for now, until we change `CAPTURE_BROWSER`'s default value.

perma_web/perma/tasks.py
```diff
@@ -694,18 +686,25 @@ def get_screenshot(link, browser):
...
         safe_save_fields(link.screenshot_capture, status='failed')
 
+
 def get_page_size(browser):
     try:
-        root_element = browser.find_element_by_tag_name('html')
-    except (NoSuchElementException, URLError):
+        return browser.execute_script("""
+            return {'height': document.body.scrollHeight, 'width': document.body.scrollWidth}
+        """)
```
rebeccacremona: I'm going to spend a few minutes seeing how this js works with phantom running, and see how it works for a few different domains with Chrome running. I didn't really give it a good try Friday.

I'm thinking it might not be that reliable.... maybe something like this instead... https://stackoverflow.com/a/1147768

perma_web/perma/tasks.py
bensteinberg: Yes, commit for this is in the pipeline. Thx!

bensteinberg:
The current status of this PR is that capture with Chrome works in our Vagrant dev environment (although I have not prepared a new Vagrant box) and in Travis CI, but not completely in Heroku. I believe this is a memory issue; for example, a capture where screenshot fails with a hobby dyno works when the dyno is 1x. I think this is probably not acceptable for our current dev/deployment workflow, and in any case won't allow us to do useful manual testing at Heroku, so I'm going to close the PR and revisit it later.

PR #2127, opened 2017-10-04
Remove secondary storage mechanism and SFTP storage

bensteinberg:


perma_web/fabfile/dev.py
```diff
@@ -624,8 +624,6 @@ def check_storage(start_date=None):
 
     # this can be generalized later to an arbitrary number of storages
     storages = {'primary': {'storage': default_storage, 'lookup': {}}}
```
rebeccacremona: Does it make sense at this point to simplify this data structure, and take out the loops (for key in storages)? Or is that too violent?

rebeccacremona: If not worth it, perhaps updating the comment to indicate there's no plan to generalize?

bensteinberg: I had a vague sense that we might someday change to multiple storages again, and then again I think this task is somewhat historical in any case.

PR #2132, opened 2017-10-11
Upgrade requests and other packages, tweak teardown for functional tests

### LINKED ISSUE #2130: Upgrade cryptography to 1.9+ to allow build on stretch
See https://github.com/pyca/cryptography/issues/3605 and https://github.com/pyca/cryptography/pull/3603 -- our current cryptography, 1.8.1, won't build with openssl 1.1.0f, which is what's in Debian stretch.

bensteinberg:
In order to upgrade `cryptography`, I started by upgrading `requests`, and then had a small chain of additional package upgrades in order to get pip-compile to complete. In the end, cryptography was still set at 1.8.1, so I specified the version by hand in `requirements.in`, and left the other upgrades in place. Then, on testing, I saw an occasional error (`selenium OSError: [Errno 9] Bad file descriptor`) on teardown of phantomjs, so I adjusted the teardown functions according to https://stackoverflow.com/a/45786385/4074877.

Fixes #2130.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,9 @@ addons:
     paths:
       - perma_web/failed_test_files
 
+  # a modern firefox
+  firefox: latest-esr
+
 cache:
   pip: true
   directories:
--- a/perma_web/functional_tests/tests.py
+++ b/perma_web/functional_tests/tests.py
@@ -189,7 +189,10 @@ class FunctionalTest(BaseTestCase):
     def quitDriver(self):
         # to get past transient error (selenium OSError: [Errno 9] Bad file descriptor),
         # send signal according to https://stackoverflow.com/a/45786385/4074877
-        self.driver.service.process.send_signal(signal.SIGTERM)
+        try:
+            self.driver.service.process.send_signal(signal.SIGTERM)
+        except AttributeError:
+            pass
         self.driver.quit()
 
     def tearDownLocal(self):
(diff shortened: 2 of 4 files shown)


bensteinberg:
Fixing the teardown problem exposed a NoSuchElementException at Travis, which looked like it could be a timeout, but turned out to be the old (38.4.0esr) Firefox installed by default in Travis' Precise image.

bensteinberg:
I've taken a look at the relevant changelogs/release notes. I think the changes to cryptography will allow us to remove some dev dependencies, FWIW.

- chardet 2.3 -> 3.0.4: https://github.com/chardet/chardet/releases
- CherryPy 3.8.2 -> 11.0.0: http://docs.cherrypy.org/en/latest/history.html
- cryptography 1.8.1 -> 2.1: https://cryptography.io/en/latest/changelog/
  2.0: We now ship manylinux1 wheels linked against OpenSSL 1.1.0f. These wheels will be automatically used with most Linux distributions if you are running the latest pip.
  2.0.1: Altered the manylinux1 wheels to statically link OpenSSL instead of dynamically linking and bundling the shared object. This should resolve crashes seen when using uwsgi or other binaries that link against OpenSSL independently.
  2.0.2: Marked all symbols as hidden in the manylinux1 wheel to avoid a bug with symbol resolution in certain scenarios.
- jasmine, jasmine-core 2.5.2 -> 2.8.0: https://github.com/jasmine/jasmine/tree/master/release_notes
- Jinja2 2.9.6 -> 2.8.1! http://jinja.pocoo.org/docs/2.9/changelog/
  A downgrade -- pip-compile barked when I attempted to put the version back to 2.9.6
- pywb 0.33 -> 0.33.2: https://github.com/ikreymer/pywb/blob/master/CHANGES.rst
  0.33.2: Better handling of exceptions from in wsgi_wrapper
requests[security] 2.11.1 -> 2.18.4: http://docs.python-requests.org/en/master/community/updates/#release-history
- selenium 2.47.3 -> 3.6.0: https://github.com/SeleniumHQ/selenium/blob/master/py/CHANGES

PR #2136, opened 2017-10-13
Send signal to phantomjs, not just to ffox in tests

bensteinberg:


bensteinberg:
It might be safer, but less Pythonic, not to specify any error where I have `AttributeError` here.

Issue #2139, opened 2017-10-17
`fab test` creates persistent warcs, via append


rebeccacremona:
Empty `services/django/generated_assets/warcs/` then run `fab test`.

```
MECC02R72AUFVH3:perma-rlc rcremona$ ls -l services/django/generated_assets/warcs/
total 0
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:54 2J
drwxr-xr-x  4 rcremona  staff  136 Oct 16 19:54 3S
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:55 42
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:54 6X
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:54 7V
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:54 8T
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:55 9W
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:55 C2
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:55 CU
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:55 ME
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:55 NG
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:55 U4
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:55 UE
drwxr-xr-x  4 rcremona  staff  136 Oct 16 19:53 X3
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:55 XM
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:53 Y2
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:54 Y7
drwxr-xr-x  3 rcremona  staff  102 Oct 16 19:55 Z6
drwxr-xr-x  4 rcremona  staff  136 Oct 16 19:43 Z8
```
Tests will pass. 

Run `fab test` again. 

Tests will fail: instead of 1 CDXLine, 2 will be returned. 

[...]

PR #2142, opened 2017-10-17
Use a tmp dir for MEDIA_ROOT when running tests; clean up afterwards.

rebeccacremona:
Doing this in fabric for 2 reasons:
1) it covers the whole test suite, including code that is imported/invoked outside of any particular test
2) you can see the tmp file created, used, and deleted all in one spot, instead of across multiple files (for instance, if the tmp dir was created in settings_testing

But, I'm totally open to other potential solutions. This was just the only thing I could think of.

See https://github.com/harvard-lil/perma/issues/2139

### DIFF
--- a/perma_web/api/tests/utils.py
+++ b/perma_web/api/tests/utils.py
@@ -135,19 +135,12 @@ class ApiResourceTestCaseMixin(SimpleTestCase):
         super(ApiResourceTestCaseMixin, self).setUp()
 
         self.api_client = LoggingAPIClient()
-
-        self._media_org = settings.MEDIA_ROOT
-        self._media_tmp = settings.MEDIA_ROOT = tempfile.mkdtemp()
-
         try:
             self.list_url = "{0}{1}".format(self.url_base, self.resource_url)
         except:
             pass
 
     def tearDown(self):
-        settings.MEDIA_ROOT = self._media_org
-        shutil.rmtree(self._media_tmp)
-
         # wipe cache -- see https://niwinz.github.io/django-redis/latest/#_testing_with_django_redis
         from django_redis import get_redis_connection
         get_redis_connection("default").flushall()
--- a/perma_web/fabfile/dev.py
+++ b/perma_web/fabfile/dev.py
@@ -1,7 +1,9 @@
 import os
+import shutil
 import subprocess
 import sys
 import signal
+import tempfile
 
 from django.conf import settings
 from fabric.context_managers import shell_env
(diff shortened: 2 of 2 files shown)


perma_web/api/tests/utils.py
```diff
@@ -135,19 +135,12 @@ def setUp(self):
         super(ApiResourceTestCaseMixin, self).setUp()
 
         self.api_client = LoggingAPIClient()
-
```
jcushman: I think we still want to run individual tests in individual temp dirs, right? Ideally no state gets preserved from one test to another. The thing in `fab` would be in addition to this instead of replace it.

perma_web/fabfile/dev.py
```diff
@@ -74,7 +76,16 @@ def test_python(apps=_default_tests):
     if "functional_tests" in apps and not os.environ.get('SERVER_URL'):
         local("DJANGO__STATICFILES_STORAGE=django.contrib.staticfiles.storage.StaticFilesStorage python manage.py collectstatic --noinput")
 
-    local("coverage run manage.py test --settings perma.settings.deployments.settings_testing %s" % (apps))
+    # temporarily set MEDIA_ROOT to a tmp directory, in a way that lets us clean up after ourselves
```
jcushman: You can use the `tempdir` package (which we also use in tasks.py) to simplify this a bit:

```
import tempdir

with tempdir.in_tempdir() as tmp:
    shell_envs = {
        'DJANGO__MEDIA_ROOT': tmp.name
    }
    local(...)
```

This takes care of wrapping in a try-finally, which otherwise you'd want to do here.

rebeccacremona: Aha! That's much nicer, thanks.

perma_web/api/tests/utils.py
rebeccacremona: "Ideally no state gets preserved from one test to another." Totally! Unfortunately, that isn't reliably achieved by the setup/teardown code that was here.

It's super interesting to watch go by in pdb: the temp directories are indeed created, settings.MEDIA_ROOT is reset appropriately at the start and finish of each test, and the tmp directory and is deleted. But, unless it's the first test that has been run that attempts to access default_storage, these steps have no effect on where the files are actually saved. 

On develop: if the perma app tests are run first, files are saved in whatever directory MEDIA_ROOT was, when storage was first accessed there. If the api tests are run first, files are saved in whatever tmp directory was created by the first api test. To watch this happen, put a breakpoint as the first line of the api tests tearDown.
```
> fab test:apps='api.tests.test_link_resource'

[localhost] local: coverage run manage.py test --settings perma.settings.deployments.settings_testing api.tests.test_link_resource

Creating test database for alias 'default'...
Creating test database for alias 'perma-cdxline'...

<truncating! press c until you see a test run that creates a capture>

ET5Z-6VYD step 0: Starting capture
....
ET5Z-6VYD capture succeeded.
> /vagrant/perma_web/api/tests/utils.py(151)tearDown()
-> settings.MEDIA_ROOT = self._media_org
(Pdb) import os
(Pdb) self._media_tmp
'/tmp/tmpOWPyIu'
(Pdb) os.listdir(self._media_tmp)
['warcs']
(Pdb) n
> /vagrant/perma_web/api/tests/utils.py(152)tearDown()
-> shutil.rmtree(self._media_tmp)
(Pdb) c

<truncating again! press c until you see another capture created>

28ZC-B2GF step 0: Starting capture
...
28ZC-B2GF capture succeeded.
> /vagrant/perma_web/api/tests/utils.py(151)tearDown()
-> settings.MEDIA_ROOT = self._media_org
(Pdb) self._media_tmp
'/tmp/tmpYIeYlg'
(Pdb) os.listdir(self._media_tmp)
[]
(Pdb) os.listdir('/tmp/tmpOWPyIu')
['warcs']
```

So, I thought it was best to remove these setUp and tearDown methods, until we can figure out a way to get them to work as advertised. Googling has not suggested a solution: the best anybody seems to have come up with is writing a customer test runner that, when setting up the testing environment, does the same thing I do here via fab.

Maybe we can manipulate the location attribute of the FileStorage object directly (if we can figure out how to get a hold of the instance in use...) https://docs.djangoproject.com/en/1.11/_modules/django/core/files/storage/#FileSystemStorage.

PR #2148, opened 2017-10-23
Run tests with pytest by default.

rebeccacremona:
There's a memory problem with python unittest in 2.7. Short version of the story: data initialized during self.setUp isn't freed at the end of the test. This behavior is changed in 3.4. See https://chrisbailey.blogs.ilrt.org/2013/05/19/pythons-leaky-testcase-aka-hidden-gotchas-using-self/ and https://bugs.python.org/issue11798

As a result, as our test suite has grown, we've needed more and more memory to run the tests. We've gotten to the point where over 2GB of RAM is required.... more than @anastasia and @rebeccacremona have available in our dev boxes. So, when running the complete suite of tests locally, we start getting failures midway, of the form `OSError: [Errno 12] Cannot allocate memory` or `raise WebDriverException("Unable to start phantomjs with ghostdriver: %s" % e) WebDriverException: Message: Unable to start phantomjs with ghostdriver: [Errno 12] Cannot allocate memory` or similar. 

Using pytest completely avoids the issue.

If anybody wants to run with manage.py (or if we want to continue using that in travis, for example), fab test now takes a second optional argument:
`fab test:pytest=False` or `fab test:apps='api perma',pytest=False` etc. to run with manage.py

### DIFF
--- a/perma_web/fabfile/dev.py
+++ b/perma_web/fabfile/dev.py
@@ -60,15 +60,15 @@ def run_ssl(port="0.0.0.0:8000"):
 _default_tests = "perma api functional_tests lockss"
 
 @task
-def test(apps=_default_tests):
+def test(apps=_default_tests, pytest=True):
     """ Run perma tests. (For coverage, run `coverage report` after tests pass.) """
     reset_failed_test_files_folder()
-    test_python(apps)
+    test_python(apps, pytest)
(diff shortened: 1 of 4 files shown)


rebeccacremona:
Weird: this doesn't catch the functional tests. Investigating....

PR #2150, opened 2017-10-25
Enforce password standards

rebeccacremona:
For PCI compliance:
```
(a) Are user password parameters configured to
require passwords/passphrases meet the
following?
 A minimum password length of at least
seven characters
 Contain both numeric and alphabetic
characters
Alternatively, the passwords/passphrases must have
complexity and strength at least equivalent to the
parameters specified above.
```
This PR does two things: 
1) immediately imposes stricter password policies for new passwords (which covers new users and any current users who reset or change their passwords)
2) adds a configuration variable VALIDATE_ALL_PASSWORDS, False by default, which if True, checks a user's password against the current validation rules when they log in. If it doesn't meet the requirements, they are forced to update it.

To force all users to update their passwords, we can set VALIDATE_ALL_PASSWORDS = True and then expire all sessions (http://blog.sashalaundy.com/blog/2014/08/26/howto-force-all-django-users-to-log-out-with-the-django-orm/)

### DIFF
--- a/perma_web/perma/models.py
+++ b/perma_web/perma/models.py
@@ -12,6 +12,7 @@ import socket
 from urllib import urlencode
 from urlparse import urlparse
 import simple_history
+import string
 import requests
 import itertools
 import time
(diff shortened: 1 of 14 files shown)


rebeccacremona:
Tests are passing locally but failing on Travis. Investigating why.

PR #2177, opened 2017-12-05
Private archive markers

anastasia:
An attempt at light-touch changes to make private records more visibly different.

screen shot 2017-12-05 at 4 09 36 pm

screen shot 2017-12-05 at 4 09 02 pm


rebeccacremona:
Looks good!

One recommendation: the orange text looked a little low contrast to me, so I checked the color contrast using Chrome Accessibility Developer Tools. Sounds like $color-orange isn't contrasty enough to meet [WCAG](https://www.w3.org/TR/UNDERSTANDING-WCAG20/intro.html) requirements.

image

A switch to their recommended color for AA compliant (the less stringent),  #c25700, to my eye makes a big difference, and seems to fit in well with the theme. (You can double click on the little color swatch in the panel, and it makes the change for you).

What would you think of adding that color to our list?

anastasia:
I don't know if this is the pull request to do it, but at some point in time we can just switch all our colors to ones compliant with the AA level. Then we could remove the $color-somecolor-a variables I added in this pull request

rebeccacremona:
Sweet, thanks!

Recording our in-person conversation for posterity: the new SCSS variables are for colors similar to those in our theme, but, which, against white background (or possibly tan background, in certain places in Perma) have a contrast ratio that meets the AA spec. Only the orange, though, is currently in use: we can phase in the others later.

PR #2200, opened 2017-12-12
Updated: language re paid subscriptions on firm signup page

abziegler:


perma_web/perma/templates/registration/sign-up-firms.html
```diff
@@ -53,8 +53,8 @@ <h2 class="body-ah">Why should lawyers and law firms use Perma.cc?</h2>
...
 
-      <h3 class="body-bh">Paid option coming soon</h3>
-      <p class="body-text">Right now, individual lawyers can use Perma.cc on a limited basis, for free, to protect up to 10 links per month. Soon we'll be launching a paid service for law firms and lawyers needing more links. If you’d like to learn more about how Perma.cc can help your firm, please fill out this form and we’ll follow up soon.<p class="body-text">
+      <h3 class="body-bh">Paid option now available</h3>
+      <p class="body-text">Right now, individual lawyers can use Perma.cc on a limited basis, for free, to protect up to 10 links per month. For those who need to protect more links, we now offer unlimited usage through paid subscriptions. Paid subscriptions are $500 per month for firms with at least 100 lawyers and $300 per month for firms with less than 100 lawyers. If you’d like to learn more about how Perma.cc can help your firm, please fill out this form and we’ll follow up soon.<p class="body-text">
```
bensteinberg: I think I'd say "fewer than" here instead of "less than".

bensteinberg: Another thing -- this isn't from your change, but I noticed it here: "we’ll" instead of "we'll", curly single quote for apostrophe rather than straight single quote. Do we have a policy or style guide for this?

PR #2202, opened 2017-12-19
Serve at perma.test instead of .dev because Google ruined everything

rebeccacremona:
https://ma.ttias.be/chrome-force-dev-domains-https-via-preloaded-hsts/


rebeccacremona:
If this is merged in, devs will need to update both their local `/etc/hosts` and the `/etc/hosts` in their vagrant box

PR #2205, opened 2018-01-02
Fix error emails from pywb

jcushman:
Provide full request and traceback info to the logger.error call in pywb_config, so we get reasonable error emails from Django's email-admins handler.


jcushman:
Happy New Year.

Issue #1976, opened 2017-04-13
Send perma an email digest of frontend errors


anastasia:
Could be a good thing to do? Otherwise we just forget /manage/errors exists

rebeccacremona:
Let's investigate why we are getting so many of these, and let's look into text encoding (e.g. " [{&quot;function&quot;: &quot;&quot;, &quot;column&quot;: 0, &quot;line&quot;: 9591, &quot;file&quot;: &quot;https://172.69.60.83/static/bundles/stats.d1d1971406e7.js&quot;}]"

rebeccacremona:
Looks like sometimes, we don't have a stack trace:
```
  File "/srv/www/perma/perma_web/perma/tasks.py", line 1351, in send_js_errors
    formatted_errors = map(lambda err: err.format_for_reading(), errors)
  File "/srv/www/perma/perma_web/perma/tasks.py", line 1351, in <lambda>
    formatted_errors = map(lambda err: err.format_for_reading(), errors)
  File "/srv/www/perma/perma_web/perma/models.py", line 1497, in format_for_reading
    formatted['stack'] = json.loads(self.stack)[0]
IndexError: list index out of range
```

rebeccacremona:
Yeah, evidently occasionally the stacktrace is a unicode cast of an empty list.

```
print(problem.stack)
u'[]'
```

PR #2207, opened 2018-01-03
add - sending email digest of last week's uncaught JS errors to perma admins

anastasia:
finishes #1976

### DIFF
--- a/perma_web/perma/tasks.py
+++ b/perma_web/perma/tasks.py
@@ -15,7 +15,7 @@ import os.path
 import threading
 import Queue as queue
 import time
-from datetime import timedelta
+from datetime import datetime, timedelta
 import urlparse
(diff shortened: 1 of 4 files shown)


perma_web/perma/tasks.py
```diff
@@ -1325,3 +1325,20 @@ def cm_sync():
...
+
+@shared_task()
+def send_js_errors():
+    """
+    finds all uncaught JS errors recorded in the last week, sends a report if errors exist
+    """
+    errors = UncaughtError.objects.filter(
+        created_at__gte=datetime.now() - timedelta(days=7),
+        resolved=False)
+    if errors.count() > 0:
```
rebeccacremona: I think this can just be `if errors:`

Issue #2095, opened 2017-08-10
Write docs for internet archive


anastasia:

rebeccacremona:
This is fantastic, thanks so much for putting it together!!

PR #2212, opened 2018-01-09
Docs

anastasia:
- finishes #2095 
- fixed API key image (updated it)
- fixed hover color for doc links


rebeccacremona:
This is fantastic, thanks so much for putting it together! 

One quick follow up question: if I wanted to read more about what happens when we delete stuff, do you think https://internetarchive.readthedocs.io/en/latest/index.html is the right place to explore? (For instance, if I wanted to see which fields we can't delete, or if I wanted to try and find out whether IA purges their servers, etc.) Is that the best source of info on interacting with Internet Archive, or are there docs that you know about?

Merging anyway; we can stick a link in there later if that seems appropriate.

PR #2214, opened 2018-01-10
Change git:// to git+http:// in dev_requirements.txt to fix Pip Vagrant error

leonidg:
pip -r dev_requirements.txt throws an error inside of Vagrant because
of the git:// requirement.  Github does not allow a clone of a git://
URL without credentials.  git+http:// allows a fallback to HTTP for
public repositories (which the Django one of course is).


leonidg:
Closing due to inability to reproduce.  Current suspect is transient Github error.

PR #2215, opened 2018-01-22
Only take link title from DOM if not already supplied by the user.

rebeccacremona:
Makes "Yowza" stick as title:
```
curl -H 'Content-Type: application/json' -X POST -d '{"url": "http://example.com", "title": "Yowza"}' http://api.perma.test:8000/v1/archives/?api_key=<api key>
```


rebeccacremona:
The tests are failing because of this: https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/models.py#L912

Rethinking my strategy.

PR #2220, opened 2018-01-24
jquery3 update

anastasia:


anastasia:
found another jquery related change needing to be made

Issue #681, opened 2014-10-20
Store folder status in url fragment


jcushman:
As you click around the folders, the url fragment should update (like `#My Links/foo/bar`) so you can link directly to a given folder.

PR #2223, opened 2018-01-24
Link straight to folders on manage/create page. 

rebeccacremona:
Addresses https://github.com/harvard-lil/perma/issues/681.

Folder specified in the URL takes precedence over that in local storage and causes local storage to update.

I decided not to use the fragment, because I think that might have unexpected effects: browsers will attempt to scroll to and set focus to an element with id equal to the fragment. If no such element exists, keyboard focus is usually dropped and set to the top-level html element, and the screen scrolls back to the top. Altering the query string should avoid those problems.

### DIFF
--- a/perma_web/static/bundles/create.js
+++ b/perma_web/static/bundles/create.js
@@ -1985,7 +1985,9 @@ webpackJsonp([1],[
 	    var selectedFolders = ls.getAll();
 	    selectedFolders[current_user.id] = { 'folderIds': folderIds, 'orgId': orgId };
 	
-	    history.pushState(null, null, "?folder=" + folderIds.join('-'));
+	    if (folderIds && folderIds.length) {
+	      history.pushState(null, null, "?folder=" + folderIds.join('-'));
+	    }
 	
 	    Helpers.jsonLocalStorage.setItem(localStorageKey, selectedFolders);
 	    Helpers.triggerOnWindow("CreateLinkModule.updateLinker");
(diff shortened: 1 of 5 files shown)


perma_web/static/bundles/create.js
```diff
@@ -1982,15 +1984,47 @@ webpackJsonp([1],[
...
+	      var folder_list = queryDict['folder'].split('-');
+	      folder_list = folder_list.map(function (s) {
+	        return parseInt(s);
+	      });
+	      folder_list.forEach(function (i) {
+	        if (isNaN(i)) throw "Invalid folder list";
+	      });
+	      return folder_list;
+	    } catch (err) {
+	      console.error(err);
```
anastasia: Is this something we want to debug later on? Might be more useful to send the error to our error handler
```
var ErrorHandler = require('error-handler.js');
ErrorHandler.airbrake.notify(err);
```

anastasia: same comment here. Thinking we can punt on this until I fix up the ErrorHandler to be a bit more useful, then I can add to it a sweep for more js errors.

perma_web/static/js/folder-tree.module.js
```diff
@@ -37,16 +37,44 @@ export var ls = {
...
+  var queryDict = Helpers.getQueryStringDict();
+  if (queryDict['folder']){
+    try {
+      var folder_list = queryDict['folder'].split('-');
+      folder_list = folder_list.map(s => parseInt(s));
+      folder_list.forEach(i => {if(isNaN(i)) throw "Invalid folder list"});
+      return folder_list
+    } catch (err) {
+      console.error(err);
+    }
```
anastasia: same comment here. Thinking we can punt on this until I fix up the ErrorHandler to be a bit more useful, then I can add to it a sweep for more js errors.

rebeccacremona: Good thinking. Switching!

Issue #1642, opened 2016-08-11
There should be an API way to download an archive


anastasia:
also bulk downloads would be nice

jcushman:
API doesn't currently include the URL to download the warc. Add something like "warc_download_url": "https://perma.cc/GUID-GUID?type=warc_download"

rebeccacremona:
That url works with wget or curl for public warcs, but we should probably also have a mechanism for link owners and admins to download private warcs.

Hmmm.

PR #2224, opened 2018-01-26
Api routes for downloading warcs

rebeccacremona:
https://github.com/harvard-lil/perma/issues/1642

Public archives can be downloaded at `/api/v1/public/archives/:guid/download` without an api_key.

All archives can be downloaded by qualified users (admins, private link owners) at `/v1/archives/:guid/download?api_key=<appropriate api key>` .

The URL used to download warcs via the GUI, `/:guid/?type=warc_download`, still works exactly like it did before: public archives may be downloaded by anyone there, and private archives may be downloaded by qualified users if they have an authenticated session (i.e. are logged in, via the log in form). This URL (still) cannot be authenticated using an API key.

### DIFF
--- a/perma_web/api/serializers.py
+++ b/perma_web/api/serializers.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.core.exceptions import ValidationError as DjangoValidationError
+from rest_framework.reverse import reverse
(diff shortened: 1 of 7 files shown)


anastasia:
This looks fantastic!

PR #2231, opened 2018-02-01
Handle requests to both api and api mirror.

rebeccacremona:
This fixes a significant bug @abziegler noticed on stage: 500 on requests that need warc_download_url that come in to api.perma.cc rather than perma.cc/api

Our tests all use /api/v1/<route> to work around a similar issue: https://github.com/harvard-lil/perma/blob/develop/perma_web/api/tests/utils.py#L110

We should address that TODO soon: optimally, we should run all tests at both urls, but with the spped of the current tests, that's not feasible.


rebeccacremona:
Done!

PR #2232, opened 2018-02-01
Count orgs on the python side to avoid counting deleted orgs.

rebeccacremona:
This means we'll hit the the database once per registrar displayed on the page, but since this view is paginated, and since this is an admin-only view, I'm not worried. 

(This well might be more efficient than annotating with the counts in the first place, since that had to be done for ALL registrars, not just the ones on the desired page.)

### DIFF
--- a/perma_web/perma/templates/user_management/manage_registrars.html
+++ b/perma_web/perma/templates/user_management/manage_registrars.html
@@ -32,7 +32,7 @@
   <div class="row row-no-bleed admin-data">
     <div class="col col-xs-6 col-no-gutter admin-data-point">
       <p class="count-label">Organizations</p>
-      <p class="count-number">{{ orgs_count.count }}</p>
+      <p class="count-number">{{ orgs_count }}</p>
     </div>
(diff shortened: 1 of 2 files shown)


jcushman:
Looks good to me -- worth a try since it's nice and simple. Not sure how fast it will be in prod -- I think it's actually running two queries per registrar to print the count twice?

If the form is too slow once deployed, the next step could be to change the `orgs_count = Organization.objects.filter(registrar__in=registrars).count()` line to fetch counts per registrar instead of counts total, and then annotate `registrars` with that info in python. (Something like https://stackoverflow.com/questions/19101665/how-to-do-select-count-group-by-and-order-by-in-django .) That could get back down to just a single additional SQL query for a page of results, while keeping the deleted-org filtering from the `Organization` query.

(Also obligatory mention that it's always worth considering just deleting the admin-only frontend pages rather than fixing them ... :)

rebeccacremona:
Ah, that's a good idea, I didn't think of that.

Issue #2098, opened 2017-08-14
"Last active" date weirdness


rebeccacremona:
<img width="913" alt="last-active" src="https://user-images.githubusercontent.com/11020492/29284139-801b376c-80f7-11e7-9a01-bdbe1b2a4d3f.png">

rebeccacremona:
A registrar's "last active" date is the date a registrar user last logged in. An organization's "last active" date is the date an organization user last logged in.

Confirming: you'd rather have both report the date an organization last logged in?

rebeccacremona:
Nope: the union of the two. Most recent login from any registrar user or organization user.

PR #2233, opened 2018-02-01
Show most recent login by reg user or associated org users.

rebeccacremona:
Addresses https://github.com/harvard-lil/perma/issues/2098


jcushman:
That's wacky but I believe you -- cool stuff. :)

PR #2234, opened 2018-02-05
UncaughtJS errors

anastasia:
- gave cron job a minute value so it doesn't run 60 times an hour
- formatted errors for emails, sending only very last value for error stack (added sending error id for easier tracking down of errors)

### DIFF
--- a/perma_web/perma/settings/utils/post_processing.py
+++ b/perma_web/perma/settings/utils/post_processing.py
@@ -39,7 +39,7 @@ def post_process_settings(settings):
         },
         'send-js-errors': {
             'task': 'perma.tasks.send_js_errors',
-            'schedule': crontab(hour='10', day_of_week=1)
+            'schedule': crontab(hour='10', minute='0', day_of_week=1)
         }
 
     }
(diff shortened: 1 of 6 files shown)


perma_web/perma/settings/utils/post_processing.py
```diff
@@ -39,7 +39,7 @@ def post_process_settings(settings):
         },
         'send-js-errors': {
             'task': 'perma.tasks.send_js_errors',
-            'schedule': crontab(hour='10', day_of_week=1)
+            'schedule': crontab(hour='10', minute='0', day_of_week=1)
```
rebeccacremona: Aha!!!! So that is how this works!!!

perma_web/perma/tasks.py
```diff
@@ -1348,9 +1348,10 @@ def send_js_errors():
         created_at__gte=datetime.now() - timedelta(days=7),
         resolved=False)
     if errors:
+        formatted_errors = map(lambda err: err.format_for_reading(), errors)
         send_admin_email("Uncaught Javascript errors",
```
rebeccacremona: Lambdas and map bring me joy.

PR #2235, opened 2018-02-05
sending users to mementoweb's Time Travel instead of Internet Archive

anastasia:
screen shot 2018-02-05 at 2 06 24 pm

falling back on today's date when wbrequest's wb_url object doesn't include timestamp field

### DIFF
--- a/perma_web/perma/templates/archive/archive-error.html
+++ b/perma_web/perma/templates/archive/archive-error.html
@@ -9,7 +9,10 @@
   <div class="record-message">
     {% if status == '404 Not Found' %}
       <p class="record-message-primary">Link not found</p>
-      <p class="record-message-secondary">You are probably seeing this page because you clicked a link in one of our records. We don’t capture secondary links, but you could check <a href="https://web.archive.org/web/*/{{ err_url }}" target="_top">the Internet Archive</a> or <a href="{{ err_url }}" target="_top">view the link as it exists on the live site</a>.</p>
+      <p class="record-message-secondary">
+        You are probably seeing this page because you clicked a link in one of our records. We don’t capture secondary links,
+        but you could check <a href="http://timetravel.mementoweb.org/list/{{ timestamp }}/{{ err_url }}" target="_top">Time Travel</a> or
+        <a href="{{ err_url }}" target="_top">view the link as it exists on the live site</a>.</p>
     {% else %}
       <p class="record-message-primary">Playback Unavailable</p>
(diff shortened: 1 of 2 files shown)


perma_web/perma/templates/archive/archive-error.html
```diff
@@ -9,7 +9,10 @@
   <div class="record-message">
     {% if status == '404 Not Found' %}
       <p class="record-message-primary">Link not found</p>
-      <p class="record-message-secondary">You are probably seeing this page because you clicked a link in one of our records. We don’t capture secondary links, but you could check <a href="https://web.archive.org/web/*/{{ err_url }}" target="_top">the Internet Archive</a> or <a href="{{ err_url }}" target="_top">view the link as it exists on the live site</a>.</p>
+      <p class="record-message-secondary">
+        You are probably seeing this page because you clicked a link in one of our records. We don’t capture secondary links,
+        but you could check <a href="http://timetravel.mementoweb.org/list/{{ timestamp }}/{{ err_url }}" target="_top">Time Travel</a> or
```
rebeccacremona: This is a great idea.

perma_web/warc_server/pywb_config.py
```diff
@@ -74,14 +76,24 @@ def get_archive_path():
...
+def raise_not_found(url, timestamp=None):
     if not settings.LOG_PLAYBACK_404:
         # use a custom error to skip pywb printing of exceptions
+        if not timestamp:
+            # if timestamp is not available, fall back on today's date
+            now = datetime.now()
+            # month and day need to be two digits
+            month = str(now.month) if now.month > 9 else "0%s" % now.month
+            day = str(now.day) if now.day > 9 else "0%s" % now.day
+            timestamp = "%s%s%s" % (now.year, month, day)
```
rebeccacremona: Nit: I think you can actually do this all in one bang using the string .format() method, which lets you do fancy stuff like:
```
>>> import datetime
>>> d = datetime.datetime(2010, 7, 4, 12, 15, 58)
>>> '{:%Y-%m-%d %H:%M:%S}'.format(d)
'2010-07-04 12:15:58'
```
Worth it to try  something like that?

rebeccacremona:
Whoops, we're all working at the same time :-)

perma_web/warc_server/pywb_config.py
anastasia: Ha! Yes! Awesome. I hate that code.

Issue #2249, opened 2018-02-08
API docs TOC


rebeccacremona:
We need some aria for the expanding/collapsing sections. Keyboard navigation seems under control.

rebeccacremona:
Aha!! Actually, we don't: the menu doesn't expand/collapse. The markup simply includes subsections for the current page, while you are on it, and omits the markup for other pages' subsections. I'll tweak this in a different way.

rebeccacremona:
(Though, since it APPEARS to expand/collapse, as a JS menu would.... This is essentially a server side implementation of functionality we are presently more used to seeing on the front end. I take back my former: aria attributes WOULD be helpful here.)

rebeccacremona:
OMG, I'm so confused. Okay, actually: this is all just one giant page, and there are SEVERAL table of contents navs on the page, one per main section. And, each of those contains subsection headings pertaining to the main section it is located visually next to.

This requires some thinking....

Issue #2247, opened 2018-02-08
Tables of contents


rebeccacremona:
put tables of contents (about, guide, terms of service, privacy policy, etc.) in labeled `<nav>` wrappers

rebeccacremona:
And probably add headers to that effect as well.

PR #2273, opened 2018-02-16
Accessible Tables of Contents

rebeccacremona:
Addresses remainder of https://github.com/harvard-lil/perma/issues/2247 and https://github.com/harvard-lil/perma/issues/2249

The solution for the API docs might be overly verbose, but I can't think of anything more succinct. This is a rather uncommon design, in my admittedly limited experience: I'm not sure there's a great solution. If we felt like it, we could consider breaking this into multiple pages, or making a single, accordion-style menu that positions itself via css.... but that sounds like much too much work. This verbose technique is probably better bang for our buck.


rebeccacremona:
Whoops, css problem.

rebeccacremona:
(never mind, it's fine... and I think I'm done with git for the day....)

Issue #2276, opened 2018-02-20
Skip to main, almost everywhere


rebeccacremona:
Placeholder issue for adding skip-to-main links on other pages, not covered by https://github.com/harvard-lil/perma/issues/2241 or https://github.com/harvard-lil/perma/issues/2246

Issue #2246, opened 2018-02-08
Skip links, about page


rebeccacremona:
Skip to main
Skip over carousel
Skip over partners list

Issue #2241, opened 2018-02-08
Skip links, landing page


rebeccacremona:
- skip over the carousel (even once we make it accessible)
- skip over the partner list (OMG they are ALL LINKS)
- skip to footer nav
- skip to main (and mark up main as main: currently no landmark)

PR #2277, opened 2018-02-21
Skip links

rebeccacremona:
Addresses:
https://github.com/harvard-lil/perma/issues/2276
https://github.com/harvard-lil/perma/issues/2241
https://github.com/harvard-lil/perma/issues/2246

I'm using a Wordpress styling for skip links (we thought it looked a little more idiomatic than bootstrap-style links, and it avoided introducing jarring motion to the page).

image

Some of the labels are pretty verbose... a hint that perhaps the structure of these pages is a little too complex. But simplifying our content/design is a job for another day.


rebeccacremona:
You should be able to navigate to these links (and all the other focusable items in the page) by pressing tab to go forwards, shift + tab to go backwards. If not, check out https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Accessibility#Using_native_keyboard_accessibility for info on tweaking your browser's settings. Chrome should work by default.

PR #2285, opened 2018-03-01
Install boto in order for Kombu to use SQS

bensteinberg:
This fixes a problem obscured in one case on Heroku by the uninvited presence of boto 2.42.0. (Probably see https://github.com/heroku/heroku-buildpack-python/issues/379.) Kombu requires boto, not boto3, to talk to SQS.


bensteinberg:
(I think this would not be necessary if we moved off the `cedar-14` stack.)

Issue #2293, opened 2018-03-19
Batch link project: track TODOs here


rebeccacremona:
## Nits

- [x] Typo: "expost" button should read "export"
- [x] Missing setting: 'BULK_UPLOADS' needs to be added to TEMPLATE_VISIBLE_SETTINGS
- [x] Importing settings: api/view.py imports settings from perma.settings. Django docs say we should import settings like this `from django.conf import settings` to make sure that everything works as expected (including giving you the flexibility to change your directory structure).https://docs.djangoproject.com/en/2.0/topics/settings/#using-settings-in-python-code
- [x] Let's call this model LinkBatches (discussed, perhaps not communicated!)
- [x] Let's squash these migrations into one
- [x] Batch/LinkBatch model needs a `__unicode__(self)` method so that it prints/displays nicely (instead of the default ` <Batch: Batch object>`)
- [x] Let's double check to make sure we use the word "batch" vs "bulk" consistently across the code and UI
- [x] If RUN_TASKS_ASYNC = False, only the first link in a batch will capture successfully. Let's make sure developers get notified of that when reading the code. Comment in settings_common? Some consistency check that prints a warning to the console on server start?
- [x] Batch model's "accessible_to" method tweak: I think this can be condensed to `return user.is_staff or self.created_by == user`
- [x] Total total nit: would "saved_folder" be better "target_folder"?
- [x] Let's catch a specific exception in the safe_get method, rather than the blanket Exception

## JS Investigations
- [x] We want the progress bar template (and any other new templates )to be inside the `{% templates %}` block... but this turns out to open a big can of worms... the progress bar template doesn't work if put there! To be investigated...

## Error Handling
- [x] A CaptureJob now needs to be created for every attempted creation of a PermaLink. The current draft of the batch link creation code handles the case where we get so far as:
    ```
    serializer = self.serializer_class(data=data, context={'request': request})
    if serializer.is_valid(): 
    ```
    But we can actually return in a number of places above that, wherever `raise_validation_error` is called, or if an unexpected exception gets raised somewhere. Let's tweak this code so that a CaptureJob is always created. Strategy might include creating a CaptureJob object at the very top of the function and changing relevant fields (status, batch, link, etc.) further down in the function as appropriate. We can figure out how to save the object most efficiently/robustly/dry-ly.
- [x] A CaptureJob should store error messages, if any.
[...]

PR #2296, opened 2018-03-20
Batch link creation, part 1

rebeccacremona:
This PR represents all the excellent work by @leonidg, with a tweak so the tests pass, and with a few things renamed per recent design discussions. 

Everything is behind a feature flag, so this shouldn't introduce any changes into the production app. We're keeping track of known TODOs here. https://github.com/harvard-lil/perma/issues/2293

If someone wouldn't mind: before merging, would you take a very quick skim to see if you can find any functionality that should be behind the feature flag that isn't?


jcushman:
LGTM!

Issue #1218, opened 2015-10-26
Your Perma Links: Should infinite scroll be triggered by a button press instead?


thisisdano:
Automatic infinite scroll means you can "never" access the footer on YPL (if you have a lot of records). Perhaps this isn't much of an issue, but maybe instead of the infinite scroll we can have a "Show more records" button?

jcushman:
Good question -- not being able to reach the footer is annoying. OTOH I think for a big folder having to click "show more" a lot would be annoying as well.

What do you think of doing a scrollable div for the record list, so it's infinite scroll inside the div but you can still scroll past the whole div to get to the footer? If we let you scroll your folder list separately from the records list that could be a usability win anyway -- e.g. if you were trying to drag and drop a bunch of links to a new folder.

thisisdano:
Sorry so late on this — I think I like the implementation on Instagram for the web: one push to trigger infinite scroll. That way, you can access the footer (if needed) but trigger infinite scroll if desired. Obvs not a first-class issue right now!

jcushman:
First pass low hanging fruit: let's try making both the folder list and link list overflow-scroll.

rebeccacremona:
Or, controversially: how about we use Django pagination instead?

rebeccacremona:
Just tested this: I'm finding that it doesn't actually work that reliably. Depending on where your cursor is, how fast you manage to scroll..... even if you get to the bottom, the links tend to catch up with you pretty quickly, and bump the footer out of sight just as soon as you get ready to click on it lol. The scrollbars are also pretty ugly. I could easily be doing this wrong, but not this technique is going to work as a first pass.

rebeccacremona:
That's the case whether you make each column scroll independently, or the whole row as a unit.

rebeccacremona:
I still think pagination is better, always and everywhere, unless a person is idly scrolling for content (as in a twitter feed), but we've had that argument already, and I lost lol. 

I'm going to implement Dano's suggestion: "I think I like the implementation on Instagram for the web: one push to trigger infinite scroll. That way, you can access the footer (if needed) but trigger infinite scroll if desired."

And then we can keep discussing this, until we find something we all like. But in the meantime, the footer will be available. 😉

rebeccacremona:
overflow=scroll works when you actually set a static height for the div... which is of course what Jack was recommending above.

So we'll do that, using js to pick some height that makes sense for the viewport.

[...]

PR #2299, opened 2018-03-23
Infinite scroll and footer access

rebeccacremona:
This is a first pass at https://github.com/harvard-lil/perma/issues/1218.

The styles look good on my browsers, but zoom settings are interfering somewhat with testing: I _think_ it's legitimately fine (at any standard zoom), but am not positive.

Scroll bars should appear when the content is too long, and not appear when the content fits. Folders and Links scroll separately. The Folders section heading with its important button, does not scroll. A subtle shadow should keep the Links list from looking peculiar during scrolling. Scroll bars should not bump other content out of the way. The Folders list should be horizontally scrollable, but the horizontal scroll bar should not appear. 

No scrolling:
image

Scrollable links, at the top:
image

Scrolling through links:
image

Scrolling folders and links at the same time:
image

### DIFF
--- a/perma_web/static/bundles/admin-stats.js
+++ b/perma_web/static/bundles/admin-stats.js
@@ -52,6 +52,9 @@ webpackJsonp([0],[
 	exports.showElement = showElement;
 	exports.hideElement = hideElement;
 	exports.addCSS = addCSS;
+	exports.scrollIfTallerThanFractionOfViewport = scrollIfTallerThanFractionOfViewport;
+	exports.viewportHeight = viewportHeight;
+	exports.markIfScrolled = markIfScrolled;
 	function setInputValue(domSelector, value) {
 	  $(domSelector).val(value);
 	}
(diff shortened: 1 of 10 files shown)


rebeccacremona:
WHOA! And, evidently, those scroll bars only show up all the time, at least on Apple devices, if you are a weirdo like me and have your preferences configured accordingly. For other people, the scroll bars show up only while you are scrolling.

How do you people know when there is more content to see, if you don't have a scroll bar?

image

anastasia:
This looks really great, both code-wise and locally in browser. I'm so happy it's happening, it makes Perma much more friendly to users. 
One question and also the tiniest of nitpicks that you're welcome to ignore:
- You treat the Folders header and the Your Perma Links header differently — one is excluded from scroll and one is included in it. I'm assuming that you do that because the Folders header has important buttons in it, like you said in the description, but wondering if you think it might be worthwhile to treat the headers the same way for clarity? 
- The nitpick is there are function-scoped `var`s that can be declared using `let`s (`scrollIfTallerThanFractionOfViewport`, `markIfScrolled`).

rebeccacremona:
Thx! I will change the `var` declarations in the folder-tree to `let`s.

Re: the headers, yeah, I was wondering about that and tried it both ways. Since the search bar is so big, and since there aren't any buttons in the Link section, I personally liked it better this way: more links visible at a time seemed like a win. I also realized at one point that very few people are going to be scrolling folders anyway: I think with very few exceptions, only admins will run into so long a list of folders. Do you feel strongly about it (fine if you do!)? 

FWIW I hate this whole thing and think it should be paginated instead lol.

anastasia:
(I'm pro pagination as well) No I don't feel strongly enough about any of it to fight for it. Let's keep it!

rebeccacremona:
Done!

Issue #2312, opened 2018-04-03
cookies with colons in the key


rebeccacremona:
A third-party script used by a number of news organizations is setting cookies with colons in the key, like `ffx:ffxDeviceID`. There's a heated argument about how python should handle such cookies here: https://bugs.python.org/issue2193. They decide in the end to accept colons, but only in python 3.3+

 Do we want to try to work around? I think yes.....

PR #2313, opened 2018-04-03
Don't explode if cookies with colons have been set during playback

rebeccacremona:
Note: this isn't a full patch of the Cookie module.

Our only interest here is to see if Perma's special cookie that grants access to a private link is set. We only need to be able to create a cookie object and check for our key. This patches the module sufficiently to let us do that. It does NOT attempt to patch the module such that it can fully handle cookies with colons as keys (parsing, etc.), because that involves too much code, and we wouldn't use any of it. For a more complete patch, see http://pythonfiddle.com/cookie-accepting-a-colon-as-a-key/ (although I note that fiddle leaves `_Translator` untouched.)

### DIFF
--- a/perma_web/warc_server/pywb_config.py
+++ b/perma_web/warc_server/pywb_config.py
@@ -7,6 +7,7 @@ import threading
 import re
 from urlparse import urljoin
 import requests
+import string
 import sys
 from datetime import datetime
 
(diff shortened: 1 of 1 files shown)


rebeccacremona:
Fixes https://github.com/harvard-lil/perma/issues/2312

Issue #2069, opened 2017-07-05
Encoding problem (gasp)


rebeccacremona:
Internal Server Error: /service/bookmarklet-create/

UnicodeEncodeError at /service/bookmarklet-create/
'ascii' codec can't encode character u'\u0308' in position 60: ordinal not in range(128)

Request Method: GET
Request URL: https://perma.cc/service/bookmarklet-create/?v=1&url=https%3A%2F%2Fmedlineplus.gov%2Fency%2Farticle%2F000500.htm%23AlternativeN%CC%88ames

File "...service.py" in bookmarklet_create
  93.     add_url = "{}?url={}".format(reverse('create_link'), tocapture)

Exception Type: UnicodeEncodeError at /service/bookmarklet-create/
Exception Value: 'ascii' codec can't encode character u'\u0308' in position 60: ordinal not in range(128)

GET:
url = u'https://medlineplus.gov/ency/article/000500.htm#AlternativeN\u0308ames'
v = u'1'

HTTP_X_FORWARDED_PROTO = 'https'
PATH_INFO = u'/service/bookmarklet-create/'
QUERY_STRING = 'v=1&url=https%3A%2F%2Fmedlineplus.gov%2Fency%2Farticle%2F000500.htm%23AlternativeN%CC%88ames'
RAW_URI = '/service/bookmarklet-create/?v=1&url=https%3A%2F%2Fmedlineplus.gov%2Fency%2Farticle%2F000500.htm%23AlternativeN%CC%88ames'

Issue #1801, opened 2016-11-09
Unicode in URLS and the bookmarklet


rebeccacremona:
(this description is not correct, it turns out. see below)
Visit a subpage of http://gaemmagami.com/ (Persian) and try to use the bookmarklet.

UnicodeEncodeError at /service/bookmarklet-create/
'ascii' codec can't encode characters in position........

Might as well look for other unicode problems, while in the zone.

bensteinberg:
Is this a job for https://docs.djangoproject.com/en/1.10/ref/unicode/#uri-and-iri-handling ?

rebeccacremona:
I can't reproduce this. If you click the link in the Django error email, indeed, you get the same error (which I accidentally triggered several times while researching this).

But, using Chrome or Firefox, if you visit http://gaemmagami.com/قوانین-مربوط-به-زنان-خانواده-کودکان/, 

or 

http://gaemmagami.com/%D9%82%D9%88%D8%A7%D9%86%DB%8C%D9%86-%D9%85%D8%B1%D8%A8%D9%88%D8%B7-%D8%A8%D9%87-%D8%B2%D9%86%D8%A7%D9%86-%D8%AE%D8%A7%D9%86%D9%88%D8%A7%D8%AF%D9%87-%DA%A9%D9%88%D8%AF%DA%A9%D8%A7%D9%86/

then everything appears to work fine, via the bookmarket or the browser extensions.

rebeccacremona:
This is definitely browser-specific behavior. All my browsers percent-encode unicode for me. But clearly, the browser of the user who ran into trouble does not:

Request information:
GET:
url = u'http://gaemmagami.com/\u0642\u0648\u0627\u0646\u06cc\u0646-\u0645\u0631\u0628\u0648\u0637-\u0628\u0647-\u0632\u0646\u0627\u0646-\u062e\u0627\u0646\u0648\u0627\u062f\u0647-\u06a9\u0648\u062f\u06a9\u0627\u0646/'
v = u'1'

Next goal: find a browser that does not.

jcushman:
Original error looks like it came from Windows 7, IE11, if I'm looking at the right thing. Could give that a shot using Sauce.

PR #2324, opened 2018-04-09
don't implicitly cast to ascii

rebeccacremona:
Closes https://github.com/harvard-lil/perma/issues/2069 and https://github.com/harvard-lil/perma/issues/1801


rebeccacremona:
If we add tests for services.py, check a bookmarklet route like http://perma.test:8000/service/bookmarklet-create/?v=1&url=http%3A%2F%2Fgaemmagami.com%2Fقوانین-مربوط-به-زنان-خانواده-کودکان%2F

PR #2328, opened 2018-04-20
First cut at exposing application version

bensteinberg:
This is a first try at exposing Perma's version number. The idea is that the deployment process will populate `version.txt` in `STATIC_ROOT` (with an arbitrary string, but probably the current git tag) and update the load balancer's healthcheck to look for the version number in question, so that application servers in the pool will only be used if they've been updated.

This doesn't need to be exposed on all pages -- happy to discuss whether a version page makes sense. It could be exposed instead in the API, but it seems a little outside the realm of what the API is for, and shouldn't be confused with API version.


bensteinberg:
The output could be an HTML comment instead of a meta tag, too.

rebeccacremona:
Note for posterity, and one second thought.

The note: to test locally, add version.txt to `/services/django/static_assets`, where collectstatic puts things, not `/static/`, where we normally put static files.

The second thought: this calls the template tag twice if version.txt is present, once when assigning to the variable pv, and once when putting the value into data-version. Total nit, but I'd consider renaming the file to `perma_version.py`, so that the load statement is identical to the tag that is loaded, and would consider `{% if pv %} data-version="{{ pv }}`, which prints the value of pv, instead of `{% if pv %} data-version="{% perma_version %}`

jcushman:
Let's do this.

PR #2335, opened 2018-05-01
First pass: batch link error handling

rebeccacremona:
- creates a CaptureJob for every attempted Permalink
- records validation errors on the CaptureJob model
- exposes that information via the admin and the api
- displays error messages in the LinkBatch admin with styling similar to the way failed captures are displayed in the Link List.

No JS error handling for this batch link excitement yet; I have to learn how first, and would rather do that after the js has stabilized a bit more.

image


rebeccacremona:
(FWIW "nil" above isn't an error: I submitted the world "nil" instead of a valid URL)

rebeccacremona:
(I think the border css needs some tweaking, but will get to that when addressing the css of the modals in general.)

PR #2336, opened 2018-05-02
Upgrade Django to 1.11.13

bensteinberg:
See https://docs.djangoproject.com/en/2.0/releases/1.11.13/


rebeccacremona:
Cool. Shall we do perma-payments too?

PR #2338, opened 2018-05-03
Handle batch-related functionality in a single modal

rebeccacremona:
instead of one with the input form and another displaying links.


rebeccacremona:
This js is still due for some refactoring, pending changes to the API. A few draft styles are added here, but I've still put essentially no effort into the markup or the appearance.

PR #2341, opened 2018-05-07
Refactor for fun

rebeccacremona:
I had a js refactoring party on Friday... and I was having so much fun, I forgot to make incremental commits. Apologies for this humongous undifferentiated blob. While refactoring the JS, I found a couple unhandled python edge cases.

Changes to utils.py: handle the case when the wrapped view function simply borks, rather than returning a well-formatted DRF error response.

Changes to tasks.py: if a capture job fails early on, there is no `link` var we can refer to in the error handling. Access link as `capture_job.link` so the the error handling doesn't 500.

The main project: pull all manual element construction out of link-batch.module.js and create-link.module.js and put it into Handlebars templates, and get Webpack to pre-compile those templates for better performance. We still have a couple other `<script>` Handlebars templates used by other modules (admin stats page, search page) which I have not refactored yet, but once those are converted to the js format, we can forgo including the large, complete Handlebars library in our bundled js, and instead can just load the much smaller runtime library`<script src="/libs/handlebars.runtime.js"></script>`. 

To simplify our interactions with the DOM, I retrieve all the necessary elements once, on module init, and store them as local variables in the module, rather than retrieving them each time we need them via jquery.

Then, I renamed a bunch of functions in create-link.module.js and organized them into sections: code related to uploading, code related to normal captures, etc.

The diff (` 2,987 additions and 2,783 deletions`) is a little misleading: the actual source code is shorter now, but the pre-compiled templates are now included in the create.js bundle.


rebeccacremona:
This should also stop some of the mysterious JS errors we are seeing reported:

```
Message: Parse error on line 48:
...>Delete</a>{{/if}}
```

```
Message: You must pass a string or Handlebars AST to Handlebars.compile. You passed undefined
```

anastasia:
Looks _excellent_. So readable, so friendly.

PR #2349, opened 2018-05-11
Store created_by on CaptureJob...

rebeccacremona:
... and don't assume capture jobs have links when assigning order.

This should fix a bug I introduced in https://github.com/harvard-lil/perma/commit/d1a01052a3653baa125af4a388bc9c002de8728b, where I arranged for every attempt to make a Perma Link to create a Capture Job. If a user with a pending capture job tried to create another one, it would have caused an Internal Server Error. Since we didn't get any emails, I think this never happened. It was exposed when testing the new Link Batch feature, which creates Perma Link requests much faster than people do.

This PR also tweaks the utility function for dispatching multiple requests so that it doesn't swallow errors as completely. 

I haven't figured out yet how we want to expose Internal Server errors to users if they occur during the Link Batch creation process.

This PR includes a peculiar data migration: I'd super appreciate a careful look at it.


bensteinberg:
Is `perma_web/perma/migrations/0033_auto_20180510_1637.py` a no-op? Doesn't the end of migration 0032 do the same thing?

rebeccacremona:
Fixed and rebased.

jcushman:
Looks good to me!

PR #2350, opened 2018-05-14
Update entire create page during batch creation

rebeccacremona:
- Refresh the link list when new links are created
- Refresh the count of links remaining in limited folders when new links are created
- Clear out more things between viewings of the batch modal, so that you don't see relics of previous modals for an instant when opening several instances in a row.
- Propagate error messages to the window's error message handler.


rebeccacremona:
@bensteinberg, this should resolve the remainder of the problems you noticed when testing the batch link feature

perma_web/api/views.py
```diff
@@ -582,7 +582,13 @@ def post(self, request, format=None):
                     'verb': 'GET'
                 }]
                 response = dispatch_multiple_requests(request.user, call_for_fresh_serializer_data)
-                return Response(response[0]['data'], status=status.HTTP_201_CREATED)
+                data = response[0]['data'].copy()
+
+                # include remaining links in response
+                links_remaining = request.user.get_links_remaining()
+                data['links_remaining'] = links_remaining
```
bensteinberg: Total nit, probably not worth picking: is there a reason not to combine lines 588-589 into `links_remaining = data['links_remaining'] = request.user.get_links_remaining()`?

rebeccacremona: Ah, right. Fixed in eebba1855151781d90fd3a4320450c70ff7c8525

Issue #2116, opened 2017-09-05
API documentation for archive view does not match result


leonidg:
The response for `https://api.perma.cc/v1/user/archives/` does not match the [documentation](https://perma.cc/docs/developer#view-all-archives-of-one-user).  Here is an example for user `law.annualsurvey@nyu.edu` (I called with `limit=`):

<pre>
{   u'meta': {   u'limit': 1,
                 u'next': u'https://api.perma.cc/v1/user/archives/?api_key=bdce593fff5dd3d7c4a97cd1142ac1f8c17090ec&limit=1&offset=1',
                 u'offset': 0,
                 u'previous': None,
                 u'total_count': 107},
    u'objects': [   {   <b>u'archive_timestamp': u'2017-09-01T19:15:02Z',</b>
                        <b>u'capture_time': 15,</b>
                        <b>u'captures': [   {   u'content_type': u'text/html; charset=utf-8',
                                             u'playback_url': u'//perma-archives.org/warc/9PHF-BA7P/http://www.law.nyu.edu/news/Ndidi-Oriji-NBCUniversal-advertising-standards-and-practices-comedy',
                                             u'record_type': u'response',
                                             u'role': u'primary',
                                             u'status': u'success',
                                             u'url': u'http://www.law.nyu.edu/news/Ndidi-Oriji-NBCUniversal-advertising-standards-and-practices-comedy',
                                             u'user_upload': False},
                                         {   u'content_type': u'image/png',
                                             u'playback_url': u'//perma-archives.org/warc/9PHF-BA7P/id_/file:///9PHF-BA7P/cap.png',
                                             u'record_type': u'resource',
                                             u'role': u'screenshot',
                                             u'status': u'success',
                                             u'url': u'file:///9PHF-BA7P/cap.png',
                                             u'user_upload': False},
                                         {   u'content_type': u'image/x-icon',
                                             u'playback_url': u'//perma-archives.org/warc/9PHF-BA7P/http://www.law.nyu.edu/sites/default/files/favicon.ico',
                                             u'record_type': u'response',
                                             u'role': u'favicon',
                                             u'status': u'success',
                                             u'url': u'http://www.law.nyu.edu/sites/default/files/favicon.ico',
                                             u'user_upload': False}],</b>
                        u'created_by': {   u'first_name': u'Leonid',
                                           u'full_name': u'Leonid Grinberg',
                                           u'id': 14558,
                                           u'last_name': u'Grinberg',
[...]

Issue #2337, opened 2018-05-02
API Docs should discuss creating archive in subfolder


dbateyko:
Similar to [this filed issue](https://github.com/harvard-lil/perma/issues/2117), but I didn't understand their fix.  Using the API documentation, I'm trying and failing to create an archive in an organizational subfolder. 

`
curl -H 'Content-Type: application/json' -X POST -d '{"url": "http://google.com", "title": "This is an example site", "folder": 'xx-xx-xx'}' https://api.perma.cc/v1/archives/?api_key=xx
`
Is there a good fix for this?

rebeccacremona:
Yes! The above is almost right: just use `"folder": 54` rather than `"folder": 2-9-54`, where the numbers match your case. Or, POST to https://api.perma.cc/v1/folders/54/archives (with, then, no need to include the folder ID in the posted data).

Sounds like our API docs aren't doing the job. We'll aim to revise and polish soon! If you have particular suggestions, we'd love to hear them!

dbateyko:
Thanks Rebecca! That's so straightforward, I can see how I overlooked that folders are integers. If you do plan to update the docs though, a brief note on how subfolders work in the folders section (in addition to the archive section) would clarify this.

rebeccacremona:
Instructions are now included here: https://perma-stage.org/docs/developer#create-an-archive

Issue #2117, opened 2017-09-05
No way to get IDs for subfolders?


leonidg:
I may be wrong on this, but I think the combination of #2115 and #2116 make it impossible to get any IDs, and therefore info for, subfolders.

* The `https://api.perma.cc/v1/folders/<folder id>/` endpoint will note whether the folder has children, but not say anything about them (consistent with the documentation).
* As #2115 explains, the `https://api.perma.cc/v1/user/organizations/<organization ID>/folders/` endpoint appears to be broken; I am not sure, but I assume that that response would have included subfolders.
* As #2116 explains, the `folders` list is missing for both endpoints that reveal details of archives.
* This isn't really an answer for the API, but as far as I can tell, the UI does not reveal the folder ID either.

leonidg:
Ah, ok, having perused the source, I have discovered `/folders/<folder id>/folders/`.  So never mind, although I would add that to the documentation. :)

rebeccacremona:
Fixed in 2356

Issue #2115, opened 2017-09-04
500 response on "shared folders inside an organization" call


leonidg:
I am getting a 500 response on the "Get a user's shared folders inside an organization" call (`https://api.perma.cc/v1/user/organizations/1/folders/?api_key=xxx`).  I tried with both a personal account that was added to an organization and the account that was created specifically for the organization (not that there's an actual difference in the underlying object, I don't think).  In both cases, the response is the web server's HTML 500 response (the one with `<h1><p>Internal Server Error</p></h1>`), so even if I'm calling it wrong, this seems like a bug.

If it helps, the accounts I tried this with are law.annualsurvey@nyu.edu and lg2651@nyu.edu.

leonidg:
I just realized that the "1" in the URL is probably the organization ID (if so, that should probably be documented), but it doesn't seem to help; I get the same result if I call `https://api.perma.cc/v1/user/organizations/390/folders/` (390 is the ID for "Annual Survey of American Law," an organization of which the user is a member).

rebeccacremona:
Investigating this:

Part 1: the 500 error was, I believe, due to a missing "return" statement, fixed in https://github.com/harvard-lil/perma/commit/b7eb44292fe907e43dd1ee5160cc8d98c4224c97.

Part 2: `/organizations/:id/folders` is indeed not a route in the current perma api, even though it is documented! If you check the routes mention in the issue above, you will now get a 404. Let's definitely figure out what happened here 😄

rebeccacremona:
The docs have been overhauled! Info on getting an organization's folders is now explained more thoroughly.

https://perma-stage.org/docs/developer#get-user-account-details
https://perma-stage.org/docs/developer#get-one-organization
https://perma-stage.org/docs/developer#view-folder-subfolders

PR #2356, opened 2018-05-16
API Docs overhaul

rebeccacremona:
The original goal was to document the new batch feature.

But in the meantime, updated the examples to reflect all the new fields we've added and the DRF response format (which differs slightly from what Tastypie returned), and updated the structure to reflect the new URL structure, and removed sections on routes we don't expose any more (!).

Since everything was getting rewritten anyway, took a swing at clarifying https://github.com/harvard-lil/perma/issues/2115, https://github.com/harvard-lil/perma/issues/2117, and https://github.com/harvard-lil/perma/issues/2337.

Feel free to merge without reviewing: this is definitely easier to read in the browser.


rebeccacremona:
Closes https://github.com/harvard-lil/perma/issues/2116 too

Issue #1668, opened 2016-08-31
investigate deleted and pending capture jobs on stats page


bensteinberg:

rebeccacremona:
We've made a lot of progress identifying the cause for the "pending" jobs. 
1) https://github.com/harvard-lil/perma/commit/b0873c79ee7110ee22fd5c852d3be5c3bb1c7b81
2) adding soft celery timeouts, and raising HaltCaptureException when soft timeout is triggered
3) locating unqualified try/accept loops that caused the soft celery timeout to be swallowed. 

At this point, we should probably review all the pending capture jobs, and see if any of them have warcs that can be played back, and adjust accordingly.

PR #2374, opened 2018-05-17
CaptureJob failure messages and cleanup of old jobs

jcushman:
- Make sure we set CaptureJob.message when setting status='failed', in the same format that we sometimes do now (`'{"errors":["reason"]}'`).
- Whenever `run_next_capture` is run (which should be at least once a minute due to celerybeat), mark any jobs as failed that have been in_progress longer than the celery hard timeout.

Note that this will update all the jobs we currently have on prod that managed to stick around as in_progress in the past. I think it's safe to say we're not planning to do anything with those at this point anyway. But it will also set `capture_job.link.tags.add('hard-timeout-failure')` on those jobs, so we could find them if necessary.

### DIFF
--- a/perma_web/perma/models.py
+++ b/perma_web/perma/models.py
@@ -36,6 +36,7 @@ from django.core.cache import cache
 from django.core.files.storage import default_storage
 from django.db import models
 from django.db.models import Q, Max, Count
+from django.db.models.functions import Now
 from django.db.models.query import QuerySet
 from django.utils import timezone
 from django.utils.crypto import get_random_string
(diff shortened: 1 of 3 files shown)


bensteinberg:
It's not clear to me that `run_next_capture` runs every minute as a result of celerybeat, in any of our environments. Should `update_stats` be calling `run_next_capture`, or should we add `update_stats` to `CELERYBEAT_JOB_NAMES`? Or am I missing something?

rebeccacremona:
We'll still want to do https://github.com/harvard-lil/perma/issues/1668, but we'll need to filter using the tag instead.

rebeccacremona:
Hmm now that you mention it, I see that run_next_capture is not in fact a celerybeat job: https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/settings/utils/post_processing.py#L29

jcushman:
Huh, yeah. I thought that was happening but can't find any evidence it ever was. It's a nice fallback in case a capture_job gets created but the queue isn't writeable right then. I'll add that as a separate bug.

PR #2376, opened 2018-05-17
Run run_next_capture every minute with celerybeat

### LINKED ISSUE #2375: Add minute-ly run_next_capture
As a fallback in case the queue is unreachable, have a celerybeat job the queues a `run_next_capture` job once a minute to check for pending capture_jobs.

bensteinberg:
Closes #2375?


rebeccacremona:
Looks good! Works locally!

PR #2377, opened 2018-05-21
Update error display code to handle null values

### LINKED ISSUE #2360: Batch modal error handling
If `link status="failed", message=null`, `stripDataStructure()` gets called with null, which results in an infinite loop.

jcushman:
Fixes #2360


rebeccacremona:
So much safer. Thanks!

Issue #2363, opened 2018-05-17
Batch queue information


rebeccacremona:
When testing this at scale, we quickly filled up the Celery queue. This made it clear that, when you have to wait a long time for links in a batch (which you WILL have to do, if there are like 60 captures in the queue), the modal just looks like it's frozen. 

We should display info about each link's position in the queue, when they are pending.

And, we can use some fuzzy arithmetic to predict how long you'll have to wait for your whole batch to complete. Add the estimated wait time to the top of the modal.... and reveal the hidden status information currently only available to screen reader users.

rebeccacremona:
We reviewed this at Perma meeting, and consensus was:
- don't estimate the completion time for the whole batch
- don't provide specific estimates: round to the closest minute
- add some styling for the error report up at the top
- check with Ben about alternatives to hard coding estimates

rebeccacremona:
First draft done.

PR #2387, opened 2018-06-13
Queue/wait times

rebeccacremona:
Give people rough estimates of how long the batch will take. 

image

image

image

### DIFF
--- a/perma_web/perma/templates/archive/includes/link_batch_modal.html
+++ b/perma_web/perma/templates/archive/includes/link_batch_modal.html
@@ -26,8 +26,8 @@
           </div>
         </div>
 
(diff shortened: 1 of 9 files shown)


rebeccacremona:
During screen reader testing, I found that only some of the message gets read for in-progress batches because we refresh so quickly. We could consider removing the "log" role and doing something else, but I think this is probably fine. Would love to get the opinion of some real users at some point.

rebeccacremona:
Closes https://github.com/harvard-lil/perma/issues/2363

rebeccacremona:
This now calculates the expected wait time instead of hard coding guesses, and looks like the below:

image

image

PR #2426, opened 2018-07-24
Mechanism for blocking the playback of arbitrary urls.

rebeccacremona:
Override REFUSE_PLAYBACK in settings.py, so that we don't publicly expose blacklisted urls.


perma_web/perma/settings/deployments/settings_common.py
```diff
@@ -318,6 +318,9 @@ def _pywb_static_dir_location():
 }
 LOG_PLAYBACK_404 = False
 
+# URLS we should not allow to plackback
```
bensteinberg: fix typo

PR #2436, opened 2018-08-07
Add a release script

bensteinberg:


rebeccacremona:
nice! is it worth documenting what version of python you are running this with?

bensteinberg:
Good idea. This has to be Python 3, which I was thinking (!?) we could take as read now, but is *not* necessarily true in the host machine. Maybe I'll rewrite this to accommodate 2 or 3.

Issue #2444, opened 2018-08-09
Redirect issue with timestamp urls


rebeccacremona:
```
[09/Aug/2018 19:38:58] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate/ HTTP/1.1" 301 1433
[09/Aug/2018 19:38:58] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate HTTP/1.1" 301 1437
[09/Aug/2018 19:38:58] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate/ HTTP/1.1" 301 1433
[09/Aug/2018 19:38:59] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate HTTP/1.1" 301 1437
[09/Aug/2018 19:38:59] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate/ HTTP/1.1" 301 1433
[09/Aug/2018 19:38:59] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate HTTP/1.1" 301 1437
[09/Aug/2018 19:38:59] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate/ HTTP/1.1" 301 1433
[09/Aug/2018 19:38:59] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate HTTP/1.1" 301 1437
[09/Aug/2018 19:38:59] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate/ HTTP/1.1" 301 1433
[09/Aug/2018 19:38:59] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate HTTP/1.1" 301 1437
[09/Aug/2018 19:38:59] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate/ HTTP/1.1" 301 1433
[09/Aug/2018 19:38:59] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate HTTP/1.1" 301 1437
[09/Aug/2018 19:38:59] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate/ HTTP/1.1" 301 1433
[09/Aug/2018 19:38:59] "GET /warc/20180809193208/https://lil.law.harvard.edu/collaborate HTTP/1.1" 301 1437
[...]

Issue #2442, opened 2018-08-09
Permissions issue w/ timestamp urls


rebeccacremona:
https://perma-archives.org/warc/HT38-Z2LT/http://lil.law.harvard.edu/ works but https://perma-archives.org/warc/20161215124949/http://lil.law.harvard.edu/ (timestamp version of same capture) does not

`Error message: warcs/HT/38/Z2/HT38-Z2LT.warc.gz: 403 Client Error: Forbidden for url: https://perma.s3.amazonaws.com/generated/warcs/HT/38/Z2/HT38-Z2LT.warc.gz`

rebeccacremona:
Probably I need to switch the way pywb gets the warc here, like in https://github.com/harvard-lil/perma/commit/49004c0d1ccf33a134419e2e404cd76654dc5e64

rebeccacremona:
I think this is because we are patching in our custom record_loader here (https://github.com/harvard-lil/perma/blob/develop/perma_web/warc_server/pywb_config.py#L552), but we are not patching in a custom resolving_loader https://github.com/webrecorder/pywb/blob/0.33.2/pywb/webapp/handlers.py#L135

https://github.com/webrecorder/pywb/blob/0.33.2/pywb/webapp/handlers.py#L135

rebeccacremona:
nope, that's not it

rebeccacremona:
The first difference in a normal playback and a timestamp playback:

timestamp:
```
-> def handle_replay(self, wbrequest, cdx_lines):
(Pdb) self
<warc_server.pywb_config.PermaHandler object at 0x7f254cfd27f0>
```

regular:
```
-> def handle_replay(self, wbrequest, cdx_lines):
(Pdb) self
<warc_server.pywb_config.PermaGUIDHandler object at 0x7f254cf74860>
```
playback then proceeds identically until we get to here:

timestamp
```
> /root/.local/share/virtualenvs/perma_web-kkDLCj8f/lib/python3.5/site-packages/pywb/webapp/replay_views.py(62)render_content()
-> def render_content(self, wbrequest, cdx_lines, cdx_loader):
(Pdb) self
<pywb.webapp.replay_views.ReplayView object at 0x7f254cfd2860>
(Pdb) self.content_loader.record_loader.loader
<pywb.utils.loaders.BlockLoader object at 0x7f254cfd27b8>
```

regular:
```
[...]

PR #2445, opened 2018-08-13
Timemap fixes

rebeccacremona:
Closes https://github.com/harvard-lil/perma/issues/2444 and https://github.com/harvard-lil/perma/issues/2442.

### DIFF
--- a/perma_web/warc_server/pywb_config.py
+++ b/perma_web/warc_server/pywb_config.py
@@ -294,6 +294,7 @@ class PermaHandler(WBHandler):
     def _init_replay_view(self, config):
         replay_view = super(PermaHandler, self)._init_replay_view(config)
         replay_view.response_class = PermaMementoResponse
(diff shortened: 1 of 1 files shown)


rebeccacremona:
I don't think this is redundant with the redirect-loop protections applied to warc routes that include GUIDs via https://github.com/harvard-lil/perma/blob/develop/perma_web/warc_server/pywb_config.py#L188 via https://github.com/webrecorder/pywb/blob/0.33.2/pywb/framework/archivalrouter.py#L84

rebeccacremona:
And, per discussion, though we are still not caching any CDXLines in Redis for timestamp-format URLs, by introducing the CachedLoader, this PR does start caching the WARC file itself, which I'd imagine somewhat addresses the performance bottleneck we sometimes see with these routes.

jcushman:
Cool!

Issue #2425, opened 2018-07-24
Expose link count in batch link admin


bensteinberg:
@rebeccacremona says, "It should probably be a @cached_property on the model".

PR #2449, opened 2018-08-15
Display capture job count in link batch admin

rebeccacremona:
Closes https://github.com/harvard-lil/perma/issues/2425

### DIFF
--- a/perma_web/perma/admin.py
+++ b/perma_web/perma/admin.py
@@ -191,8 +191,9 @@ class CaptureJobAdmin(admin.ModelAdmin):
 
 
 class LinkBatchAdmin(admin.ModelAdmin):
-    list_display = ['id', 'created_by', 'started_on', 'target_folder']
+    list_display = ['id', 'created_by', 'started_on', 'target_folder', 'capture_job_count']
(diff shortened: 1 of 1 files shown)


perma_web/perma/models.py
```diff
@@ -1273,7 +1273,7 @@ class CaptureJob(models.Model):
     human = models.BooleanField(default=False)
     order = models.FloatField(db_index=True)
     submitted_url = models.CharField(max_length=2100, blank=True, null=False)
-    created_by = models.ForeignKey(LinkUser, blank=False, null=False, related_name='capture_jobs')
+    created_by = models.ForeignKey(LinkUser, blank=False, null=False)
```
jcushman: Why are you dropping the related name here? It seems better to have `LinkUser.capture_jobs` than `LinkUser.capture_job_set`.

rebeccacremona: For no good reason. Fixing.

rebeccacremona: (I saw "related_name='capture_jobs'" below for link_batch and was momentarily confused, thought this was a typo)

Issue #2428, opened 2018-08-01
Investigate pywb and tmp files


rebeccacremona:
Occasionally, we are seeing playback minions' disks fill up!

rebeccacremona:
e.g. `1.1G    /tmp/_pywbcachetJAW5v`

rebeccacremona:
```
4096 Jul 19 13:25 _pywbcache603MlU
4096 Jul 19 21:06 _pywbcacheBQFbGm
12288 Aug  1 16:31 _pywbcacheKgDd18
12288 Aug  1 16:31 _pywbcacheNY6qnQ
4096 Jul 19 20:12 _pywbcacheNqwXC0
12288 Aug  1 16:31 _pywbcacheOZXtnB
4096 Jul 19 21:06 _pywbcacheTtJNgP
4096 Jul 19 21:05 _pywbcacheWd1MNE
12288 Aug  1 16:31 _pywbcachel0Iwe0
20480 Aug  1 16:31 _pywbcachetJAW5v
```

rebeccacremona:
Here's where those tmp files are created: https://github.com/webrecorder/pywb/blob/0.32.0/pywb/webapp/rangecache.py#L40

rebeccacremona:
`with NamedTemporaryFile(delete=False, dir=self.temp_dir) as fh:`
> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later). If delete is true (the default), the file is deleted as soon as it is closed.
https://docs.python.org/3/library/tempfile.html
So, the files in this directory have to be explicitly deleted.

rebeccacremona:
The cleanup function is registered with atexit to run at "interpreter termination time" https://docs.python.org/3.5/library/atexit.html. So in our context, what does that mean? From what I can tell, we run uwsgi with 5 processes, 1 thread each, and don't have any config that tells uwsgi to reload/kill processes at any point, so I think these directories only get cleaned up when perma is redeployed or if uwsgi is manually restarted. There's uwsgi config for reloading/killing processes when various conditions are met; we could look into that.

But first, let's figure out when these files are being written. Caching to disk, and cleaning up on process restart, doesn't sound like something we want to do. I don't see any temp files being created locally during normal playbacks.

rebeccacremona:
[...]

PR #2450, opened 2018-08-15
Don't cache response bodies on disk when clients request http ranges.

rebeccacremona:
Proposed solution to https://github.com/harvard-lil/perma/issues/2428. I think this is at least as fast as the original, given that we are caching warcs in redis.

`curl http://perma.test:8000/warc/9ZF9-X6DB/http://lil.law.harvard.edu/collaborate -i -H "Range: bytes=0-"` gets the whole response

`curl http://perma.test:8000/warc/9ZF9-X6DB/http://lil.law.harvard.edu/collaborate -i -H "Range: bytes=0-1023"` gets `Content-Range: bytes 0-1023/8545`, which ends like this: `<script src='/static/pywb/wb.js'> </sc`

`curl http://perma.test:8000/warc/9ZF9-X6DB/http://lil.law.harvard.edu/collaborate -i -H "Range: bytes=1024-8545"` picks up with `ript>`, correctly. There's a weird off-by-one error with the content length (`Content-Range: bytes 1024-8544/8545`), but I have confirmed that's the case with or without my patch.

### DIFF
--- a/perma_web/warc_server/pywb_config.py
+++ b/perma_web/warc_server/pywb_config.py
@@ -15,6 +15,7 @@ from django.db import close_old_connections
 from django.template import loader
 from django.test import RequestFactory
 from surt import surt
+import types
 
 # configure Django
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "perma.settings")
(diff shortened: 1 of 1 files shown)


rebeccacremona:
Thanks @jcushman for your help figuring out how to get at this method! For the future record: https://filippo.io/instance-monkey-patching-in-python/

jcushman:
Sweet! Monkey-patching level up.

PR #2460, opened 2018-08-20
Timemap html view

rebeccacremona:
Makes links absolute and forces the use of the correct urlconf in pywb. The forcing of the urlconf might be redundant, if we can figure out what's going on with our staging tier, but in the meantime, this ought to do it.


rebeccacremona:
tests are failing because, during the functional tests, request.host = localhost, not perma.test

PR #2462, opened 2018-08-22
Faster tests

rebeccacremona:
This PR makes a few changes to the way we run our tests, without actually changing any of our tests. I THINK it includes meaningful changes; a couple more commits experiments that didn't seem fruitful are available here: https://github.com/rebeccacremona/perma/commits/faster-tests.

There is still plenty of room for refactoring. Another day! Hopefully another day soon.


rebeccacremona:
Um, hold one. I messed up my pipenv move.

PR #2465, opened 2018-08-24
Individual users may request paid accounts

rebeccacremona:
We did not design our infrastructure such that individual users can subscribe to unlimited accounts, but we can reuse the Registrar -> Org -> user setup to make do.

This PR offers "premium,  unlimited" accounts to individual users during sign-up, on their /manage/create page, and in the docs. It sends them to a souped-up version of our contact form to "upgrade".

I put in some copy; I don't imagine it's what we'll want to deploy.

This PR does not address the second half of this project: adapting the code to make the Registrar -> Org setup smoother (and harder to misuse), when it is used to shoehorn individual users into our paid account setup.

### DIFF
--- a/perma_web/perma/views/user_management.py
+++ b/perma_web/perma/views/user_management.py
@@ -1318,6 +1318,10 @@ def sign_up(request):
         form = UserForm(request.POST)
         if form.is_valid():
             new_user = form.save()
+            upgrade = request.POST.get('upgrade', None)
+            if upgrade:
+                new_user.requested_account_type = 'premium'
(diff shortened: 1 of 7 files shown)


perma_web/perma/views/user_management.py
```diff
@@ -1318,6 +1318,10 @@ def sign_up(request):
         form = UserForm(request.POST)
         if form.is_valid():
             new_user = form.save()
+            upgrade = request.POST.get('upgrade', None)
+            if upgrade:
+                new_user.requested_account_type = 'premium'
+                messages.add_message(request, messages.INFO, "We will be in touch shortly with more information about upgrading to premium, unlimited service.")
```
abziegler: "to premium" => "to our premium"

PR #2466, opened 2018-08-27
Better handing of email we send to ourselves

rebeccacremona:
We send ourselves some administrative emails. Most of those should only go to whoever is currently in settings.ADMINS. Occasionally, we might want to email the main Perma email address and open a ticket. This PR adds a utility function for sending ourselves mail and switches existing code to use it, always with the default (=no ticket).

Perma Payments is going to need the same change.


rebeccacremona:
Why isn't this getting the updated commit? Closing and reopening.

PR #2468, opened 2018-08-30
Individual subscriptions

rebeccacremona:
- behavior remains the same for everybody (admins, registrar users both paying and nonpaying, organization users both paying and nonpaying) except vanilla 10-link-a-month-free accounts.
not_toggled

- if people do express interest in upgrading (via one of the avenues added in https://github.com/harvard-lil/perma/pull/2465), a Perma.cc admin can uncheck the "nonpaying" field and supply a monthly rate in the Django admin for the user. No need go through all the rigamarole of creating a placeholder registrar and org, as before.

- until they actually do subscribe, they'll continue to get 10 links a month
toggled_not_activated

- when then do, they'll then get unlimited links
active

- if their credit card expires or something else goes wrong with payment, they'll get up to 10 links a month again
expired_free_links
expired_no_links

- if they cancel, it will be like they never subscribed in the first place
toggled_not_activated

If they aren't entitled to unlimited links and they do something interesting on the UI side (make lots of  links in one tab, switch back to a tab that still displays that they have links, hit the create button), they'll get the below error message, instead of the usual "You've already reached your limit."
api_error_message

This PR is dependent on https://github.com/harvard-lil/perma-payments/pull/83, which is not backwards compatible.


rebeccacremona:
Adam and Jack both gave informal thumbs up offline.

Issue #2477, opened 2018-09-10
Teardown very slow


rebeccacremona:
Capture https://fox13now.com/2017/08/24/federal-appeals-court-upholds-injunction-against-vidangels-streaming-service/. Observe that we get so far as 
![image](https://user-images.githubusercontent.com/11020492/45299079-70034c00-b4d8-11e8-9f32-1ec58f86ec45.png).... and then stays there for a long, long time.

I think probably...
```
2EV9-H5RE step 4.972267039616902: Waiting for post-load requests
Waiting for 1 pending requests
2EV9-H5RE step 4.989350485801697: Waiting for post-load requests
Waiting for 1 pending requests
Waited 30 seconds to finish post-load requests -- giving up.
```
...we aren't actually giving up here.

rebeccacremona:
`warcprox_controller.proxy.pool.shutdown()` is where we are getting stuck. If I comment out that line, we end up with a gorgeous speedy capture. Have we left orphaned threads, though? Checking....

rebeccacremona:
Right, if I comment that out, or if I call it with `wait=False`, predictably, we hang on to threads, one per stalled url or stream. After 3 successive captures of a page with a single stream:
```
MitmProxyHandler(tid=244,started=2018-09-11T15:38:14.087101,client=127.0.0.1:38358) True True
MitmProxyHandler(tid=104,started=2018-09-11T15:37:04.750628,client=127.0.0.1:37996) True True
MitmProxyHandler(tid=187,started=2018-09-11T15:37:31.815138,client=127.0.0.1:38178) True True
```
where `True True` => is alive, is daemonic

rebeccacremona:
Yeah not fixed

rebeccacremona:
Test with:
https://fox13now.com/2017/08/24/federal-appeals-court-upholds-injunction-against-vidangels-streaming-service/
http://bangordailynews.com/2011/11/16/news/hancock/blue-hill-farmer-cited-for-violating-state-law/
http://time.com/money/5304493/us-exports-tariffs-list/
https://www.dailyprogress.com/news/local/when-it-comes-to-college-sports-revenue-even-a-powerhouse/article_29eb808c-edd8-11e3-9f88-0017a43b2370.html
and
http://partytronic.com

Probably in batches. Probably with easy captures interspersed. Try stressing out a single Celery worker, to see if there are problems.

rebeccacremona:
[...]

Issue #2019, opened 2017-05-15
Warcprox threads don't always die at capture end


rebeccacremona:
Local behavior: with the dev server up (fab run), I have two python processes running. The first, I think, is fabric. The second, I think, is the server, and it has two threads. E.g.
```
(perma)vagrant@perma:/vagrant/perma_web$ pgrep python
26622
26997
(perma)vagrant@perma:/vagrant/perma_web$ ls /proc/26997/task
26997  27003
```
As expected, during a capture, you can see new threads being created, and eventually see them get torn down (you have to wait a while for the proc files to get cleaned out, even after the thread is done).

However, when I capture partytronic.com, the very last thread in the list sticks around forever. It's repeatable: for each capture of patrytronic, one additional thread, always the last one.

e.g.
```
(perma)vagrant@perma:/vagrant/perma_web$ ls /proc/26997/task
26997  27003
(perma)vagrant@perma:/vagrant/perma_web$ ls /proc/26997/task
26997  27003  27280  27281  27282  27283  27307  27317  27318
(perma)vagrant@perma:/vagrant/perma_web$ ls /proc/26997/task
26997  27003  27318
```
Doing this even once means you can't quit out of the dev server in the normal way: ^c only quits node and the first python process, and the server keeps running in the background (I haven't found a way to bring it to the foreground, but I'm a super newbie in this area). Auto-reload on code changes also stops working. You have to manually kill the process instead.

rebeccacremona:
See this check after teardown():
```
def teardown(thread_list, browser, display, warcprox_controller, warcprox_thread):
    print("Shutting down browser and proxies.")
    for thread in thread_list:
        # wait until threads are done (have to do this before closing phantomjs)
        if hasattr(thread, 'stop'):
            thread.stop.set()
        thread.join()
    if browser:
        browser.quit()  # shut down phantomjs
    if display:
        display.stop()  # shut down virtual display
    if warcprox_controller:
        warcprox_controller.stop.set()  # send signal to shut down warc thread
    if warcprox_thread:
        warcprox_thread.join()  # wait until warcprox thread is done writing out warc
    for thread in threading.enumerate():
        if thread.is_alive():
            print "THREAD IS ALIVE!!! {}\n{}".format(thread, thread._Thread__target)
```
Results:
```
THREAD IS ALIVE!!! <_MainThread(MainThread, started 139729509562112)>
None
[...]

PR #2480, opened 2018-09-11
Patch warcprox's proxy function to be stoppable.

rebeccacremona:
Not sure about this: making a PR to facilitate discussion. Would close https://github.com/harvard-lil/perma/issues/2477 and the underlying issue, https://github.com/harvard-lil/perma/issues/2019.

But I think this technique is pretty ridiculous.

TBD....


rebeccacremona:
If I were going to try and contribute something like this upstream to warcprox, what would it look like? Not a global variable called stop....

Issue #2491, opened 2018-10-01
Repeatable DB errors while creating CDXLines!!!


rebeccacremona:
ADL8-966Y and 8UZZ-W9LH at least. Usually, "gone away", but once, (1153, "Got a packet bigger than 'max_allowed_packet' bytes")"!!

PR #2492, opened 2018-10-01
Only create 1000 cdxlines at a time.

rebeccacremona:
Should prevent https://github.com/harvard-lil/perma/issues/2491


rebeccacremona:
This limit is chosen arbitrarily, based on https://docs.djangoproject.com/en/1.11/ref/models/querysets/#bulk-create, which says they similarly limit themselves for SQLite.

HOW BONKERS, OVER 1000 NETWORK REQUESTS FOR ONE PAGE?!?!

PR #2493, opened 2018-10-01
Noarchive behavior

rebeccacremona:
This PR adds a setting, `PRIVATE_LINKS_IF_GENERIC_NOARCHIVE`. 

Perma's current behavior is True: though we only respect robots.txt directives if Perma is mentioned by name, we respect meta tags and x-robots headers directed at generic "robots" (http://noarchive.net/).

This PR also toggles the default setting to False.


rebeccacremona:
something is messed; this PR isn't picking up any changes.

PR #2501, opened 2018-10-15
Display an existing Perma Link in image.

rebeccacremona:
Thanks @ChefAndy


bensteinberg:
Is it worth it pr possible to delete  `websites-change-square.svg`?

PR #2507, opened 2018-10-30
Upgrades

rebeccacremona:
This upgrades requests...... and then upgrades everything else we don't have pinned, partly to solve some dependency issues, partly for style, partly because that's what Pipenv wants us to do lol.


rebeccacremona:
Oh, and the new pycodestyle was upset by some formatting, so fixed that too.

PR #2513, opened 2018-12-06
Individual subscription tiers

rebeccacremona:
This changes how link limits work for individuals.

## for review

This is huge and hard to read and hard to understand..... I'm not sure reviewing the code is worth it. We'll test the living daylights out of it on stage.

But, I would love a second look at the migrations..... I don't want to mess up any production data...

I can rebase if there are too many.....

### nonpaying users

- all users who aren't explicitly marked `nonpaying` have the option to purchase a personal subscription to Perma, even if they are registrar users, org users, or admins
- all current users are migrated so that `self.nonpaying == False`
- all current users who don't already have a price set get a default one

### link limits

- personal link limits can be unlimited, or particular integers associated with subscription tiers set up in `settings.TIERS`. Link limits are associated with a link limit period: once, monthly, or annually. Once means once ever. Right now, we don't have the ability to sell `once` tiers; we can only offer them manually, for free, to particular users via the admin. Link limits that refresh `monthly` and `annually` can be sold
- current "premium" users keep their status
- all other users are migrated to get 10 personal links `once`. NOT 10 links a month, like it is on production today

### ui

- the UI advertises that people can upgrade to more personal links, if they have their personal links folder selected and if they haven't purchased unlimited links

### upgrading/downgrading

- there's a page for upgrading and downgrading between tiers. once you buy a subscription, you can't downgrade back to the trial `once` 10 links tier
- upgrading and downgrading should work all the time..... with some weirdness if you upgrade or downgrade on the very day you are scheduled to be charged for an existing subscription. to be discussed and to be further worked on

### code quality

This is hopefully fine; I don't pretend it's gorgeous. There are some reasonably thorough tests. They aren't comprehensive.


rebeccacremona:
Danger zone for deployment: this makes some significant changes to the user table and to some of the data therein.

This needs to be deployed along with the latest Perma Payments code: they work together.

rebeccacremona:
No changes to docs are included in this PR.

rebeccacremona:
image
Sample subscription page for a new registrar user at a Firm, before they pay

rebeccacremona:
Sample link whereby an org user or registrar user can purchase more personal links 
image

rebeccacremona:
image

PR #2515, opened 2018-12-11
Tier tweaks

rebeccacremona:
Fixes bugs found by Adam and Clare during testing and makes requested UI changes:

## UI 

- instead of "updating" subscriptions, you modify them
- display of subscription page altered for registrar users
- headings don't change depending on subscription status (helpful for registrars, who may have multiple subscriptions
- "upgrade for more links" not displayed if no folder selected (like when logging in fresh as an org user)
- if you have requested a downgrade, you can cancel that request, but cannot otherwise switch tiers

## Bugs

- after requesting a downgrade, the stats on your current subscription were incorrect, and you were subsequently given improper upgrade/downgrade options, potentially resulting in overcharges
- error in the logic for deciding if any tiers were available for you to upgrade/downgrade to


rebeccacremona:
Oh, and added an "effective" date to the cancellation request page, to potentially help clear up the confusion about what cancellation means.

PR #2516, opened 2018-12-11
Forgot to add field to LinkUser admin.

rebeccacremona:


rebeccacremona:
Necessary so I can manually fix the records of test LinkUsers and test Registrars with pending downgrade requests, which won't look correct without a cached_subscription_rate, which cannot be populated from Perma Payments automatically.

PR #2518, opened 2018-12-14
Upgrades

rebeccacremona:
Cautiously, but boldly, start upgrading packages that are more central to our app.


rebeccacremona:
If we don't want to deploy this at the same time as tiers, then let's mark the PR "changes requested" so that nudgebot doesn't hassle us every day until January.

PR #2520, opened 2019-01-02
Webrecorder Integration - Support for Webrecorder Replay

ikreymer:
Integration of refactored `wr-api` branch to enable Webrecorder-based replay.
Enabled if `ENABLE_WR_PLAYBACK` is set to true in the settings.


rebeccacremona:
What a thrill!

My clever attempt to resolve the merge conflicts was not so clever. I will fix it when merging.... which we will do as soon as is prudent, given the number of other things we have up on staging right now.

ikreymer:
@rebeccacremona should be resolved now! Looks like its running the CI

rebeccacremona:
Documentation for our future selves: to play back Perma Links via Webrecorder:

Configure Webrecorder with a `wr.env` of:
```
APP_HOST=perma.test:8000
CONTENT_HOST=perma-archives.test:8092
```
and spin it up.

Then, comment in the Webrecorder network in Perma's `docker-compose.yaml` and spin it up.

Tweak your Perma settings.py to something like....
```
from .deployments.settings_dev import *

SECRET_KEY = 'a super secret key'
ENABLE_WR_PLAYBACK = True
if ENABLE_WR_PLAYBACK:
    PLAYBACK_HOST = 'perma-archives.test:8092'
    ALLOWED_HOSTS = ['perma.test', 'api.perma.test']
```
and then freely toggle between ENABLE_WR_PLAYBACK = True and ENABLE_WR_PLAYBACK = False

perma_web/perma/utils.py
```diff
@@ -608,6 +612,21 @@ def clear_wr_session(request):
...
+
+
+def get_wr_uploaded(request, link):
+    """
+    Mark a link id/collection slug as having been uploaded
+    """
+    return request.session.get('wr_uploaded:' + link)
+
+
+
```
rebeccacremona: Great idea to store this info in the user's session!

I think the logic of these functions would fit quite naturally as methods on the `Link` model, rather than independent utility functions. E.g.
```
# models.py

class Link(DeletableModel):
    ...
    
    def set_uploaded_during_wr_session(self, request):
        """
        Record in user's session that this link has been uploaded to WR for playback
        """
        request.session['wr_uploaded:' + self.wr_collection_slug] = True

    def get_uploaded_during_wr_session(self, request):
        """
        Retrieve from user's session whether this link has been uploaded to WR for playback
        """
        return request.session.get('wr_uploaded:' + self.wr_collection_slug)
```
No strong opinions on method names 😉

ikreymer: Sure, could move it there, but note that currently all WR related functions are in one place in utils, and also the `request.session` manipulation happens there. The cleanup `wr_uploaded` keys for example is in `clear_wr_session()`. Perhaps there should be a separate WebrecorderUtils class or something like that. Don't feel strongly about it either way :)

rebeccacremona: We do actually have a number of link methods related to WR, for logic that is specific to a given link:
```
link.wr_collection_slug
link.wr_iframe_prefix
link.init_replay_for_user
link.upload_to_wr
```
To me these seem of a kind. I'm sure we haven't managed to be totally consistent, but generally we try to put as much logic as possible in the models, and only create util functions when the logic doesn't match nicely with any particular Perma model, like  `get_wr_session` and `clear_wr_session`. 

I do see what you are saying about the desirability of keeping the `request.session` manipulation all in one spot to assist in proper cleanup. IMO keeping link-related logic within the link model is weightier.... Hmmmm. Maybe just popping any key that is prefixed with `wr_` in `clear_wr_session` would be sufficiently elegant?

perma_web/perma/models.py
```diff
@@ -1230,6 +1230,76 @@ def clear_cache(self):
...
+                    valid_if=lambda code, data: code == 200 and data.get('done'))
+
+                # successfully uploaded if reached here
+                return
+
+            except WebrecorderException as e:
+                time.sleep(0.5)
+                # if all uploaded (size == total_size), but not yet done after
+                # timeout, likely something is wrong, so raise exception
+                if ((upload_data.get('total_size') == upload_data.get('size')) and
```
rebeccacremona: A question about this... `upload_data` is not updated during this `while True` loop, but rather is set once at line 1275 by the return value of the `PUT` call. Does that mean there are two separate error conditions here: one where PUT returns `upload_data.get('total_size') != upload_data.get('size')`, and one where `upload_data.get('total_size') == upload_data.get('size')`, but the `GET` never succeeds?

ikreymer: Good catch! The `upload_data` should be getting updated in the loop, now fixed.
The api returns a json with several fields, including `size` / `total_size` and `done` boolean.
The size/total_size is useful if displaying a graphical progress bar.
However, the `done` field indicates that the upload finished and succeeded.

Occasionally, we have seen an error case where the WARC has uploaded, (size == total_size), but `done` is not set to true. This indicates some kind of error (which could be reported in a better as well).
For now, if the size == total_size but done is still not true after some time, treat this as an upload failure.

ikreymer: Or, to put another way, if the WARC has been fully uploaded/sent, but WR has not returned that its done after the timeout, then it really has failed to fully upload.

PR #2521, opened 2019-01-02
No "upgrades" for individuals on contact form or sign up form

rebeccacremona:
There should be no conflicts with https://github.com/harvard-lil/perma/pull/2519/files, which doesn't alter the actual forms.

Currently, when an organization clicks the links to request more information about a paid organizational account, they see the below:
image

This PR switches it to:
image


clare-stanton:
Would suggest: 

"Organizations may purchase unlimited Perma.cc service for their members that includes collaboration tools and administrative controls at a group rate."

PR #2527, opened 2019-01-07
Bonus links

rebeccacremona:
To give a user a batch of extra Personal Links, put an integer in their "bonus_links_remaining" field, via the Django admin. No time limit applies to bonuses: subsequent creation of Personal Links will count first against your bonus_links_remaining, and then, after they run out, against your normal link limit.

This can be used for customer support issues.

With this deployment, we are using it to ease the transition to subscription Perma tiers: all users are being awarded a 1-time bonus of up to 10 Personal Perma Links. 
- If this deployment would have resulted in them no longer being able to create links, they get 10 bonus Links.
- If they've never created any before, they get 0 bonus Links: their regular, unused allocation of 10 does the job.
- If they've created between 1 and 9 Links, they get enough bonus Links to top them off, so that between their regular allocation + the bonus Links results in them being able to create 10 more.


rebeccacremona:
This isn't quite right.

rebeccacremona:
Okay, I think this is now right.

rebeccacremona:
No still not quite right.

PR #2530, opened 2019-01-09
Give all trial users exactly 10 more links

rebeccacremona:
This PR arranges for every trial user (every user w/o a paid subscription) to get exactly 10 more links, going forward.

It starts tracking whether a user is "in_trial", default True, in the DB as well, to make a number of things easier.... and make this possible.

CAVEAT:
We evidently have some special users with special link limits right now, in prod. I don't want to deal with that programmatically.

Before we deploy tiers, let's get a list of all our special users, and make a note if we want to reinstate their special limits after the fact. We can do so manually; shouldn't take long.


rebeccacremona:
Data migration ignores Registrar customers.

PR #2538, opened 2019-01-22
Some utils for running Perma with WR

rebeccacremona:
This is a first pass at running WR locally using our pre-built Docker image. 

Includes an alternate command for `web` in `docker-compose.yml`, if you want Perma to start up automatically when you run `docker-compose up`

If doing simultaneous WR and Perma development, see the instructions at the bottom of `docker-compose.yml` to adjust the network.

Includes two utility functions for clearing WR-related data from Perma's session store, useful when debugging/ restarting services.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -117,6 +117,7 @@ services:
       - redis
     volumes:
       - ./data:/data
+      - ./services/docker/webrecorder/wr-custom.yaml:/code/webrecorder/config/wr-custom.yaml
     networks:
       - webrecorder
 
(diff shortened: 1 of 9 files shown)


bensteinberg:
Is it possible to leave the keys out of `wr.env`?

rebeccacremona:
You do need some dummy keys to run, kind of like out Django app has dummy secret keys set to `I'm a super secret key` or similar. We could require the running dev to have environment variables set in their shell, and then inject those into the container's environment using an `env` stanza for the all the services that need them. We could... hmmm... will think about other options. 

Does this strike you as unlike the Django dummy secret key situation?

rebeccacremona:
Actually, to be honest, I haven't tried running WR without any keys at all: the docs just say we need them. I should try it and see what happens.

Issue #2407, opened 2018-06-27
Upgrade django-ratelimit


rebeccacremona:
When a version containing https://github.com/jsocol/django-ratelimit/commit/b2841923a1affd33c0c3403bc23110a763ed5ee2 is released.

rebeccacremona:
Done.

PR #2541, opened 2019-01-24
Upgrades

rebeccacremona:
This PR upgrades Django, and also upgrades any other packages permitted by our constraints specified Pipfile. See Pipfile.lock for a complete list. Of note:

- [This commit](https://github.com/harvard-lil/perma/issues/2407) has landed in `django-ratelimit`. So, that plus the new setting I added to `settings_common.py` means we should no longer run into the situation where perma-dev.org goes out of commission due to overstepping our redis instance's limits.

- Upgrades to `pytest` and `pytest-django`  went too far for `pytest-django-ordering`; I have a [PR out](https://github.com/rlucioni/pytest-django-ordering/pull/5) to upgrade the package, but in the meantime have pinned us to use my fork.

- The most recent version of `fakeredis` is a release candidate, which is too bleeding edge for our code, so I've pinned us to only accept minor and point releases for now.

Everything else seems to have upgraded without problems.

NOTE FOR DEPLOYMENT: I ran into some trouble with `pipenv` hanging when generating this new lockfile, possibly due to user error re: the package we are installing from my fork. There's no reason there should be trouble, but if anything funny happens, I recommend deleting any pip cache that's on the minion and trying again, and/or throwing out the virtualenv and starting fresh.


rebeccacremona:
Yeah, this failed on Travis in the way I initially saw locally....... I think it's a virtualenv problem, investigating.....

rebeccacremona:
The failure was due to the wrong version of Django being installed. 

The wrong version of Django was being installed because of the combination of `pip 9.0.1`, `pipenv` installing the dependencies in synchronous mode, and installing my forked version of `pytest-django-ordering` from github using setup.py, which mentions that it requires django but correctly does not specify what version.

Upgrading `pip>=9.0.2` was sufficient to solve the problem; running with `pip==9.0.1` and `pipenv --sequential` also solves the problem.

PR #2554, opened 2019-02-12
Tests w/ WR playback (local only)

rebeccacremona:
This PR fixes a number of issues that were uncovered by running tests locally with `ENABLE_WR_PLACKBACK=True`, and adds infrastructure for running functional tests using headless Chrome in a Docker container whose image is managed by SeleniumHQ.

It does not change how tests run by default: `ENABLE_WR_PLAYBACK` and `REMOTE_SELENIUM` are both False.

To-dos and remarks:
- It does not appear that Webrecorder playback works in PhantomJS; TBD.
- Perma's JS error reporting does not appear to work with the dockerized Chrome; TBD
- This includes two workarounds for WR auth difficulties; TBD with WR team.
- This does not include a mechanism for easily switching between WR "dev" and "testing" settings (toggle comments, restart the containers in between)
- Next: Travis config for running with `REMOTE_SELENIUM` True..... Docker!

### DIFF
--- a/perma_web/functional_tests/tests.py
+++ b/perma_web/functional_tests/tests.py
@@ -9,26 +9,37 @@ from urllib.parse import urlparse
 import requests
 from pyvirtualdisplay import Display
 from selenium import webdriver
-from selenium.common.exceptions import ElementNotVisibleException, NoSuchElementException
+from selenium.common.exceptions import ElementNotVisibleException, NoSuchElementException, StaleElementReferenceException
 import time
 import signal
 
+from django.conf import settings
 from django.contrib.staticfiles.testing import StaticLiveServerTestCase
 from django.core.urlresolvers import reverse
 
 from perma.urls import urlpatterns
 from perma.wsgi import application as wsgi_app
 from perma.models import UncaughtError
-from perma.settings import SAUCE_USERNAME, SAUCE_ACCESS_KEY, USE_SAUCE, TIMEGATE_WARC_ROUTE, WARC_ROUTE
 from perma.tests.utils import failed_test_files_path
 
 
 # In this file we point a browser at a test server and navigate the site.
-# There are two separate choices to make:
-# (1) is the server remote, or a local Django test server we start for the occasion?
-# (2) is the browser a local PhantomJS browser, or a set of remote Sauce browsers?
+# There are several choices to make:
+# (1) is the Perma application server:
+#    (a) remote
+#    (b) a local Django test server we start for the occasion on 127.0.0.1
+#    (c) a local Django test server we start for the occasion, but broadcast to 0.0.0.0, to allow other Docker containers on the same network to browse the site.
 #
-# Examples:
+# (2) is the browser:
+#    (a) a set of remote Sauce browsers
+#    (b) a local PhantomJS browser
+#    (c) a selenium-driven browser running in its own Docker container
+#
+#
+# N.B. Some of these combinations may not work with the current code base.
+# TODO: document, or fix, or remove...
+#
+# Example invocations:
 #
 # Local browser, local server:
 # $ fab test:functional_tests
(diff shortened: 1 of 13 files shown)


perma_web/functional_tests/tests.py
```diff
@@ -44,26 +55,22 @@
...
 
-# (2) Configure Sauce vs. local PhantomJS browsers:
-
-if USE_SAUCE:
+if settings.SAUCE_USERNAME:
     from sauceclient import SauceClient
-    assert SAUCE_USERNAME and SAUCE_ACCESS_KEY, "Please make sure that SAUCE_USERNAME and SAUCE_ACCESS_KEY are set."
-    sauce = SauceClient(SAUCE_USERNAME, SAUCE_ACCESS_KEY)
+    assert settings.SAUCE_USERNAME and settings.SAUCE_USERNAME, "Please make sure that settings.SAUCE_USERNAME and settings.SAUCE_USERNAME are set."
+    sauce = SauceClient(settings.SAUCE_USERNAME, settings.SAUCE_USERNAME)
```
rebeccacremona: Mmm.... find/replace gone awry in this stanza.

perma_web/functional_tests/tests.py
```diff
@@ -137,77 +144,112 @@ class FunctionalTest(BaseTestCase):
...
-            self.setUpSauce()
+            if settings.REMOTE_SELENIUM:
+                cls.server_url = 'http://{}'.format(settings.HOST)
+            else:
+                cls.server_thread.host = LOCAL_SERVER_DOMAIN
+                cls.server_url = cls.live_server_url
+
+    @classmethod
+    def setUpBrowser(cls):
+        if settings.SAUCE_USERNAME:
```
rebeccacremona: I think this too was a find/replace error

perma_web/functional_tests/tests.py
```diff
@@ -137,77 +144,112 @@ class FunctionalTest(BaseTestCase):
...
-    def tearDown(self):
-        if USE_SAUCE:
-            self.tearDownSauce()
+            cls.setUpLocal()
+        cls.driver.implicitly_wait(5)
+        print("Using %s for integration tests." % (type(cls.driver)))
+
+    @classmethod
+    def tearDownClass(cls):
+        if settings.SAUCE_USERNAME:
```
rebeccacremona: And here.

perma_web/functional_tests/tests.py
```diff
@@ -137,77 +144,112 @@ class FunctionalTest(BaseTestCase):
...
+    @classmethod
+    def setUpSauce(cls):
+        desired_capabilities = dict(cls.base_desired_capabilities, **cls.desired_capabilities)
+        desired_capabilities['name'] = cls.id()
         sauce_url = "http://%s:%s@ondemand.saucelabs.com:80/wd/hub"
-        self.driver = webdriver.Remote(
+        cls.driver = webdriver.Remote(
             desired_capabilities=desired_capabilities,
-            command_executor=sauce_url % (SAUCE_USERNAME, SAUCE_ACCESS_KEY)
+            command_executor=sauce_url % (settings.SAUCE_USERNAME, settings.SAUCE_USERNAME)
```
rebeccacremona: And here.

perma_web/functional_tests/tests.py
```diff
@@ -247,7 +289,7 @@ def type_to_element(element, text):
             element.send_keys(text)
 
         def info(*args):
-            if USE_SAUCE:
+            if settings.SAUCE_USERNAME:
```
rebeccacremona: And here.

perma_web/functional_tests/tests.py
```diff
@@ -283,7 +325,7 @@ def repeat_while_false(func, timeout=10, sleep_time=.1):
                 time.sleep(sleep_time)
 
         def fix_host(url):
```
rebeccacremona: Is this method needed any more?

rebeccacremona: Actually.... this could easily be why the functional tests are failing on travis if I inspect the contents of the iframe directly. Is the iframe's src pointed at the test server, for pywb playback? How could it be? (Why does it work locally.....)

rebeccacremona: Putting back the iframeless loading of the warc causes tests to pass, which is good enough. But, experimenting w/ a tweaked `/etc/hosts` on travis and direct inspection of the iframe, because that would be better.

rebeccacremona: YES! It was just `/etc/hosts`

PR #2556, opened 2019-02-15
Optionally run tests with Selenium/Chrome and Webrecorder on Travis

rebeccacremona:
- upgraded to Trusty to get a recent-enough Docker/Docker-Compose
- upgrading to Trusty required losing the option to run functional tests w/ the Travis pre-installed Firefox; our Selenium is too old to handle the slightly-more-modern Firefox that comes with Trusty. To run tests with Firefox, we should get the dockerized Selenium/Firefox working.

This PR does *not* toggle on Selenium/Chrome or Webrecorder playbacks. You can test with Selenium/Chrome + pywb playback, if you want, but Webrecorder playback cannot be tested with PhantomJS; best I can tell, it requires a more modern browser.

If you want Travis to test with Webrecorder playbacks, toggle `REMOTE_SELENIUM` and `ENABLE_WR_PLAYBACK` from `False` to `True` in `settings_testing.py` (....necessary so that the related settings also get toggled appropriately).

If you want to run tests locally, do as above, but also toggle `wr.env` to use the special local settings mentioned there, and re-run `docker-compose up` before running tests. Switch back and restart again if you want your own web browser to work again. (Sub-optimal, but I can't work out a better alternative at the moment.)

(toggling on to demo seems to have uncovered a bug that didn't show up locally... dunno, maybe I never ran it, somehow.... but fixed in this PR in any case)

### DIFF
--- a/perma_web/fabfile/dev.py
+++ b/perma_web/fabfile/dev.py
@@ -65,6 +65,10 @@ def test(apps=_default_tests):
 @task
 def test_python(apps=_default_tests, travis=False):
     """ Run Python tests. """
+    if os.environ.get('DOCKERIZED'):
+        print("\n\n\n!!!!!!\n!!!!!!\nWarning! Webrecorder requires test-specific settings in this context.\n" +
+              "Be sure to services/docker/webrecorder/wr.env appropriately and\n" +
+              "reload the WR containers by re-running `docker-compose up` before running tests.\n!!!!!!\n!!!!!!\n\n\n")
 
     # .pyc files can contain filepaths; this permits easy switching
     # between a Vagrant- and Docker-based dev environment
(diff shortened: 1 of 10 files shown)


perma_web/fabfile/dev.py
```diff
@@ -65,6 +65,10 @@ def test(apps=_default_tests):
 @task
 def test_python(apps=_default_tests, travis=False):
     """ Run Python tests. """
+    if os.environ.get('DOCKERIZED'):
+        print("\n\n\n!!!!!!\n!!!!!!\nWarning! Webrecorder requires test-specific settings in this context.\n" +
+              "Be sure to services/docker/webrecorder/wr.env appropriately and\n" +
```
bensteinberg: Probably should be "be sure to edit"

PR #2558, opened 2019-02-20
SSL in docker

bensteinberg:
This defaults to `runserver` unless you're using SSL; I can add the option to use `runserver_plus` without SSL if it seems like a good idea.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,8 @@ npm-debug.log
 .hypothesis/
 # except for the examples directory
 !.hypothesis/examples/
+
+# ignore certs/keys/confs from serving via SSL with runserver_plus
+perma_web/*.crt
+perma_web/*.key
+perma_web/*.conf
--- a/developer.md
+++ b/developer.md
@@ -51,7 +51,8 @@ something like
 self-signed certificate is created.)
 
 Remember to set `SECURE_SSL_REDIRECT = False` when you go back to
-running without SSL.
+running without SSL. Note also that if you run with SSL against a
+local (non-SSL) webrecorder, playback will fail silently.
 
 ### Run all the tests
 
--- a/developer.md
+++ b/developer.md
@@ -34,6 +34,25 @@ The server will automatically reload any time you made a change to the
 
 Press `CONTROL-C` to stop the server.
 
+To run with SSL (for instance, to test locally against a remote
+instance of webrecorder that is behind SSL), set `SECURE_SSL_REDIRECT
+= True`, try
+
+`d fab run:use_ssl=True`
+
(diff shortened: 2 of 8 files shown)


rebeccacremona:
This looks great! Two questions: 
- When might a person want to supply their own certificate, rather than using the automatically-created one?
- Do you know how this behaves if Perma is configured to interact with your local Webrecorder? I assume it fails, since the local WR's nginx is not configured for SSL, but it might be good to document the manner in which it fails.

bensteinberg:
- I can imagine a situation where a developer's browser absolutely won't put up with the self-signed certificate created here, and it makes more sense to generate a cert locally and add it to your OS or browser keychain.
- With a local webrecorder, at the moment, playback fails silently; the iframe is empty. I see a `ReferenceError: ContentFrame is not defined` in the console. I'll add a note in `developer.md`.

bensteinberg:
Whoops, stay tuned, one more test against local WR....

bensteinberg:
(Decided: leave `runserver` the default for non-SSL.)

bensteinberg:
For future reference, and this might go in the code later -- the reason for generating our own self-signed cert is that the one our current version of werkzeug creates on behalf of `runserver_plus` is not strong enough. The stacktrace looks like
```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.5/threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File "/root/.local/share/virtualenvs/perma_web-kkDLCj8f/lib/python3.5/site-packages/werkzeug/serving.py", line 646, in inner
    fd=fd).serve_forever()
  File "/root/.local/share/virtualenvs/perma_web-kkDLCj8f/lib/python3.5/site-packages/werkzeug/serving.py", line 544, in make_server
    passthrough_errors, ssl_context, fd=fd)
  File "/root/.local/share/virtualenvs/perma_web-kkDLCj8f/lib/python3.5/site-packages/werkzeug/serving.py", line 479, in __init__
    server_side=True)
  File "/root/.local/share/virtualenvs/perma_web-kkDLCj8f/lib/python3.5/site-packages/werkzeug/serving.py", line 406, in wrap_socket
    ssl_version=self._protocol, **kwargs)
  File "/usr/lib/python3.5/ssl.py", line 1077, in wrap_socket
    ciphers=ciphers)
  File "/usr/lib/python3.5/ssl.py", line 699, in __init__
    self._context.load_cert_chain(certfile, keyfile)
ssl.SSLError: [SSL: CA_MD_TOO_WEAK] ca md too weak (_ssl.c:2943)
```

PR #2564, opened 2019-02-27
Monitor WR api for availability.

rebeccacremona:
Try it out locally by setting `RUN_TASKS_ASYNC = False` and `CELERYBEAT_JOB_NAMES = [ 'verify_webrecorder_api_available']` and re-running `fab run`.

Enable on a Perma server by appending this new celerybeat job to the existing setting array `CELERYBEAT_JOB_NAMES`

### DIFF
--- a/perma_web/perma/settings/deployments/settings_testing.py
+++ b/perma_web/perma/settings/deployments/settings_testing.py
@@ -140,7 +140,7 @@ TIERS = {
     ]
 }
 
-REMOTE_SELENIUM = False
+REMOTE_SELENIUM = True
 if REMOTE_SELENIUM:
     if os.environ.get('DOCKERIZED'):
         HOST = 'web:8000'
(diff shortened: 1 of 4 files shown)


rebeccacremona:
I think we can use UptimeRobot itself for the WR content host, which IS (must be) available from the public web, and which Perma itself never contacts.

bensteinberg:
What do you think of adding the task to `CELERYBEAT_JOB_NAMES` in `settings_prod.py` and conditioning it on `ENABLE_WR_PLAYBACK`?

rebeccacremona:
@bensteinberg I would quite like that idea, except that toggling `ENABLE_WR_PLAYBACK` in `settings.py` would not make the desired change; when `settings_prod.py` is loaded, `ENABLE_WR_PLAYBACK` will have whatever value is in `settings_common.py`. I think that's a little confusing.

rebeccacremona:
Hmm, test failed for real.

PR #2579, opened 2019-04-09
Less aggressive handling of 404s during WR temp user deletion attempts.

rebeccacremona:
404s are expected here: if the WR session has timed out, WR will delete the temp user. Perma will still have a cookie stored in the user's session (Django has no mechanism for automatically expiring individual keys stored in sessions, only for expiring the whole session). So, Perma will send the DELETE request just to make sure. Don't count that as an error, unless we're doing something fancy; we can call with `error_if_wr_user_not_found=True` if we need to for some reason.

### DIFF
--- a/perma_web/perma/utils.py
+++ b/perma_web/perma/utils.py
@@ -594,7 +594,7 @@ def get_wr_session(request):
     return wr_username, wr_session_cookie, is_new_session
 
 
-def clear_wr_session(request):
+def clear_wr_session(request, error_if_wr_user_not_found=False):
     """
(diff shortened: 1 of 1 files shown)


rebeccacremona:
The error emails say `WebrecorderException 404: {'error': 'no_such_user'}`, but locally I'm seeing {'error': 'not_found'}..... hold one......

rebeccacremona:
The error message is 'not_found' if you attempt after the FIRST `tempchecker.py` step; `no_such_user` thereafter.

PR #2585, opened 2019-05-20
wr auth refactor

rebeccacremona:
This is a version of https://github.com/harvard-lil/perma/pull/2569, with merge conflicts resolved. I did it several weeks ago, so it's not as fresh in my mind as it should be, but I'm pretty comfortable with the correctness of the code.
> -  for public guids, create public collection in shared 'perma' user
> -  for private guids, continue creating temp user/collection
> -  store public and private/temp session cookies, delete private/temp user to logout
> -  check if collection actually exists each time before uploading instead of relying on internal cache

### Deploying
To deploy (relevant to stage only, since we are still doing pywb-based playbacks in production), we'll need to run a command similar to the below on ONE of the Webrecorder minions:
```
docker-compose exec app python -m webrecorder.admin \
    -c info@perma.cc public Test123Test123 archivist 'Info at Perma.cc'
```
- The command creates a permanent Webrecorder user, username "public", associated with the email address "info@perma.cc" with the password Test132Test123, with the permission-level "archivist" and the full name "Info at Perma.cc". 
- The username should be "public" on all tiers: this ensures that playback of public permalinks are available at a predictable, cachable url like `http://wr.perma-archives.test/public/acbd-efgh/20180911212718mp_/http://lil.law.harvard.edu` (the "public" in the url is the username).
- The permission level "archivist" is required; everything else (email, password, full name) can be changed to whatever. 
- The password needs to be present in Perma's config: see https://github.com/harvard-lil/perma/compare/develop...rebeccacremona:ikreymer-wr-perma-user?expand=1#diff-d3aaf937ef61d55c6a72ddebda8bc733R486: we'll need to override the setting in tier-specific `settings.py` via Salt.
- This creation command will need to be repeated if Webrecorder's redis databases are ever deleted/refreshed.

### Testing
The code in this PR works best paired with a newer version of Webrecorder. I'm working on a new Docker image. We should check that this is working in stage by testing a public and a private Permalink or two, but no more than that: let's wait until I finish the new image before we put this through its paces.


rebeccacremona:
(re-running the tests locally, with WR playback turned on, which is not what Travis is currently checking)

rebeccacremona:
Okay good: tests passed locally, as expected.

PR #2589, opened 2019-05-22
Hack to refuse playback of particular urls or domains.

rebeccacremona:
This patches Webrecorder to add a mechanism for blocking the playback of blacklisted urls, similar to https://github.com/harvard-lil/perma/pull/2426/files.

We should add `contentcontroller.py` to the list of files we watch in the WR repo, after this is merged.

Let's discuss the best way to set the env var `REFUSE_PLAYBACK` so that we don't expose the list in the repo. Probably, the easiest/best way is to interpolate it into `wr.env` via Salt, along with other confidential config. Probably should NOT mimic the dev setup I'm using here, where docker-compose sets an env var in the container by interpolating the value of `REFUSE_PLAYBACK` in my shell (my bash shell on my mac). 

`REFUSE_PLAYBACK` should be a comma separated list, like `REFUSE_PLAYBACK=http://example.com,http://foo.com/bar/baz.html`

### DIFF
--- a/docker-compose-travis.yml
+++ b/docker-compose-travis.yml
@@ -30,6 +30,7 @@ services:
     volumes:
       - wr_warcs:/data/warcs:ro
       - ./services/docker/webrecorder/wr-custom.yaml:/code/webrecorder/config/wr-custom.yaml:ro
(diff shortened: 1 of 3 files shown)


rebeccacremona:
(closed and reopened to see if webhook -> Slack triggers)

Issue #2596, opened 2019-05-28
wr and replacement links


rebeccacremona:
This mechanism for playing back the correct warc during "replacement" was broken during the recent re-architecture, for public Perma Links: https://github.com/harvard-lil/perma/blob/develop/perma_web/api/views.py#L445

Add a call to the WR api to delete the public collection

PR #2599, opened 2019-05-29
Handle link patch/replace with new WR integration.

rebeccacremona:
Observe the "TODO" about Cloudflare. 

There is probably some refactoring to be done here (keep the keys "wr_temp_username" and similar in Perma config? make this method more concise?), but I'm not sure.

Will close https://github.com/harvard-lil/perma/issues/2596


rebeccacremona:
Tests were run locally with `REMOTE_SELENIUM = True` and `ENABLE_WR_PLAYBACK = True` and passed.

Issue #2478, opened 2018-09-10
Cookies with slashes in key cause 500 on playback


rebeccacremona:
```
Traceback (most recent call last):
  File "/root/.local/share/virtualenvs/perma_web-kkDLCj8f/lib/python3.5/site-packages/pywb/rewrite/cookie_rewriter.py", line 19, in rewrite
    cookie = SimpleCookie(cookie_str)
[10/Sep/2018 13:12:38] "GET /warc/2EV9-H5RE/https://s2.wp.com/_static/??/wp-content/themes/vip/localtv-2/js/plugins/throttle-debounce.js,/wp-content/themes/vip/plugins/safe-report-comments/js/ajax.js?m=1519230344j HTTP/1.1" 200 958
  File "/usr/lib/python3.5/http/cookies.py", line 507, in __init__
    self.load(input)
  File "/usr/lib/python3.5/http/cookies.py", line 556, in load
    self.__parse_string(rawdata)
  File "/usr/lib/python3.5/http/cookies.py", line 620, in __parse_string
    self.__set(key, rval, cval)
  File "/usr/lib/python3.5/http/cookies.py", line 512, in __set
    M.set(key, real_value, coded_value)
  File "/usr/lib/python3.5/http/cookies.py", line 380, in set
    raise CookieError('Illegal key %r' % (key,))
http.cookies.CookieError: Illegal key 'ballotbox/respondent_id'
```
try https://fox13now.com/2017/08/24/federal-appeals-court-upholds-injunction-against-vidangels-streaming-service/ to reproduce

rebeccacremona:
Confirmed, this persists w/WR playback.

rebeccacremona:
Line + traceback from modern pywb:
https://github.com/webrecorder/pywb/blob/master/pywb/rewrite/cookie_rewriter.py#L19
```
app_1             | Traceback (most recent call last):
app_1             |   File "/usr/local/lib/python3.7/site-packages/pywb-2.3.0.dev0-py3.7.egg/pywb/rewrite/cookie_rewriter.py", line 19, in rewrite
app_1             |     cookie = SimpleCookie(cookie_str)
app_1             |   File "/usr/local/lib/python3.7/http/cookies.py", line 480, in __init__
app_1             |     self.load(input)
app_1             |   File "/usr/local/lib/python3.7/http/cookies.py", line 529, in load
app_1             |     self.__parse_string(rawdata)
[...]

PR #2612, opened 2019-06-04
Silence cookie errors

rebeccacremona:
Addresses https://github.com/harvard-lil/perma/issues/2478

Eventually, this will be part of a PR to pywb, but not that soon: I think they'd prefer a larger logging-related PR, rather than a bunch of little ones.


bensteinberg:
I was tripped up for a minute when I saw gevent -- hadn't we just removed that in #2611? -- but then realized that this file belongs with WR.

PR #2623, opened 2019-06-06
Correct testing setup

bensteinberg:


bensteinberg:
Closing to get PR to test at Circle...

bensteinberg:
Reopening....

bensteinberg:
OK, this is now passing at both Circle and Travis. I've squashed all my experiments into one commit. I don't love the two versions of `wr.env`. I also am not sure this is the best way to set up the database before testing. To be discussed.

services/docker/webrecorder/wr.env
```diff
@@ -38,10 +38,10 @@ DISABLE_SSR=false
 # ============================
 ALLOW_EXTERNAL=1
```
rebeccacremona: I think the changes to this file can be reverted.

PR #2668, opened 2019-08-02
Upgrade Django to 1.11.23

bensteinberg:


bensteinberg:
This lock file was produced with `docker-compose exec web pipenv lock --keep-outdated`

PR #2679, opened 2019-10-16
New CA bundle

bensteinberg:
See https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL-certificate-rotation.html and https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html


bensteinberg:
The stage tier looks fine with the new CA bundle; next step is to update the cert on the RDS instance in question and check it again.

bensteinberg:
With cert updated, the application appears to be working fine.

PR #2680, opened 2019-10-23
Update Pillow

bensteinberg:


bensteinberg:
Thanks, @rebeccacremona, for noticing the change in example.com!

bensteinberg:
For what it's worth, an attempt to remove the dependency on example.com by testing the about page in the web container itself didn't work:

failed_functional_test

PR #2683, opened 2019-12-18
Update Django

bensteinberg:
I altered `Pipfile.lock` with `docker-compose exec web pipenv lock`, so it's a somewhat bigger change than it would have been with `--keep-outdated`. I can do that instead if it would be better.

I had to pin `django-model-utils` because the latest version, 4.0.0, requires Django to be at least 2.0.1.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,7 +36,7 @@ services:
       - default
   web:
     build: ./perma_web
-    image: perma3:0.29
+    image: perma3:0.30
     tty: true
     command: bash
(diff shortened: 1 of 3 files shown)


bensteinberg:
Oh -- I don't understand why `python-dateutil` is getting downgraded from 2.8.1 to 2.8.0.

rebeccacremona:
Thanks! Looks great. I don't see anything alarming in the changelogs of any of the various packages... though of note, DRF's changelog does not yet describe 3.11. Weird about the date package decrementing.... whatever.

Issue #2696, opened 2020-01-22
SSL error from warcprox


rebeccacremona:
```
2020-01-22 17:16:28,399 [ERROR] mitmproxy.py 380: problem processing request 'GET /en HTTP/1.1': SSLError(1, '[SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:720)')
Traceback (most recent call last):
  File "/root/.local/share/virtualenvs/perma_web-kkDLCj8f/lib/python3.5/site-packages/warcprox/mitmproxy.py", line 372, in do_COMMAND
    self._connect_to_remote_server()
  File "/root/.local/share/virtualenvs/perma_web-kkDLCj8f/lib/python3.5/site-packages/warcprox/warcproxy.py", line 184, in _connect_to_remote_server
    return warcprox.mitmproxy.MitmProxyHandler._connect_to_remote_server(self)
  File "/root/.local/share/virtualenvs/perma_web-kkDLCj8f/lib/python3.5/site-packages/warcprox/mitmproxy.py", line 279, in _connect_to_remote_server
    self._remote_server_conn.sock, server_hostname=self.hostname)
  File "/usr/lib/python3.5/ssl.py", line 385, in wrap_socket
    _context=self)
  File "/usr/lib/python3.5/ssl.py", line 760, in __init__
    self.do_handshake()
  File "/usr/lib/python3.5/ssl.py", line 996, in do_handshake
    self._sslobj.do_handshake()
  File "/usr/lib/python3.5/ssl.py", line 641, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLError: [SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:720)
2020-01-22 17:16:28,404 [WARNING] mitmproxy.py 499: code 500, message [SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:720)
2020-01-22 17:16:32,530 [INFO] __init__.py 161: <ListenerPostfetchProcessor(RunningStats, started daemon 140606953219840)> shutting down
2020-01-22 17:16:32,533 [INFO] __init__.py 161: <WarcWriterProcessor(WarcWriterProcessor, started daemon 140606961612544)> shutting down
2020-01-22 17:16:32,626 [INFO] warcproxy.py 500: shutting down
```
[...]

PR #2697, opened 2020-01-23
Proof of concept: running on Buster.

rebeccacremona:
Running on Buster instead of Stretch would solve https://github.com/harvard-lil/perma/issues/2696.

This, if merged, would be a no-op for prod, except insofar as some background python packages are upgraded, due to unavoidable bits of the Pipenv process.

Tests are passing locally; there may still be some fussing for the CI tests.... Will address if so.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: python
 python:
-  - "3.5"
+  - "3.7"
 env: CFLAGS="-O0"
 
-dist: trusty
+dist: xenial
+sudo: required
 
 services:
   - docker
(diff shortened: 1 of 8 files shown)


rebeccacremona:
The DB-related error we're seeing with that one test is because of a long-standing "warning" getting escalated to an "error" when changing versions of MySQL. With our Docker setup, it's still just a warning, which is why tests are passing there and at Cicle-CI.

Eventually worked out that the problem is that `character_set_server | latin1`.
```
MySQL [perma]> show variables like 'char%';
+--------------------------+----------------------------+
| Variable_name            | Value                      |
+--------------------------+----------------------------+
| character_set_client     | utf8mb4                    |
| character_set_connection | utf8mb4                    |
| character_set_database   | utf8                       |
| character_set_filesystem | binary                     |
| character_set_results    | utf8mb4                    |
| character_set_server     | latin1                     |
| character_set_system     | utf8                       |
| character_sets_dir       | /usr/share/mysql/charsets/ |
```
I can fix easily in Docker; will figure out how to fix in Travis next. We should probably check prod & stage to see how that's set up.... but since those are using RDS instances, seems less likely there is a problem.....

Issue #2682, opened 2019-11-11
inoperable URI-G and URI-T values in HTTP response


phonedude:
I noticed this a while ago but I'm just now getting to opening an issue.  Taking the URLs for the TimeMap (URI-T) and TimeGate (URI-G) from this response:

$ curl -I https://perma.cc/7TVJ-PMF7
HTTP/1.1 200 OK
Date: Mon, 11 Nov 2019 23:00:45 GMT
Content-Type: text/html; charset=utf-8
Connection: keep-alive
Set-Cookie: __cfduid=dba4cf8d2137ef05dc6f8f762ebf1debd1573513245; expires=Tue, 10-Nov-20 23:00:45 GMT; path=/; domain=.perma.cc; HttpOnly; Secure
Memento-Datetime: Sun, 29 Apr 2018 18:24:10 GMT
Vary: Cookie, Host
Link: <https://twitter.com/zittrain/status/990648288918626304>; rel="original"; datetime="Sun, 29 Apr 2018 18:24:10 GMT",<https://perma.cc/7TVJ-PMF7>; rel="memento"; datetime="Sun, 29 Apr 2018 18:24:10 GMT",<https://wr.perma-archives.org/warc/timegate/https://twitter.com/zittrain/status/990648288918626304>; rel="timegate",<https://wr.perma-archives.org/warc/timemap/*/https://twitter.com/zittrain/status/990648288918626304>; rel="timemap"; type="application/link-format"
Cache-Control: public, max-age=3600
Strict-Transport-Security: max-age=15768000
CF-Cache-Status: MISS
Expires: Tue, 12 Nov 2019 00:00:45 GMT
Accept-Ranges: bytes
Set-Cookie: sessionid=f8n10zmqehphoeat8zir2jw4rej12nbw; expires=Mon, 25-Nov-2019 23:00:45 GMT; HttpOnly; Max-Age=1209600; Path=/; Secure
Set-Cookie: __cflb=1582820817; Secure; path=/; expires=Tue, 12-Nov-19 22:00:45 GMT; HttpOnly
Expect-CT: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
Server: cloudflare
CF-RAY: 5343e255afbec084-ORF

I get 303'd to an error page:

$ curl -I https://wr.perma-archives.org/warc/timemap/*/https://twitter.com/zittrain/status/990648288918626304
HTTP/1.1 303 See Other
[...]

Issue #2471, opened 2018-09-02
Include styled and non-styled timemap urls + MIME type in memento "link" headers


phonedude:
I uncovered this in the previous issue, and it might still be related to https://github.com/harvard-lil/perma/issues/2442 and https://github.com/harvard-lil/perma/issues/2458

but when TimeMap links for URLs of the form:

https://perma-archives.org/warc/89ZB-BWVX/http://www.library.umass.edu/

don't return TimeMaps, they return HTML pages:

% curl -I https://perma-archives.org/warc/89ZB-BWVX/http://www.library.umass.edu/
HTTP/1.1 200 OK
Date: Sun, 02 Sep 2018 19:46:13 GMT
Content-Type: text/html; charset=utf-8
Content-Location: https://perma-archives.org/warc/89ZB-BWVX/http://www.library.umass.edu/
Memento-Datetime: Fri, 03 Nov 2017 21:11:50 GMT
Link: <https://perma-archives.org/warc/89ZB-BWVX/http://www.library.umass.edu/>; rel="memento"; datetime="Fri, 03 Nov 2017 21:11:50 GMT", <http://www.library.umass.edu/>; rel="original", <https://perma-archives.org/warc/*/http://www.library.umass.edu/>; rel="timemap"; type="application/link-format", <https://perma-archives.org/warc/timegate/http://www.library.umass.edu/>; rel="timegate"
[deletia]

% curl -I https://perma-archives.org/warc/*/http://www.library.umass.edu/
HTTP/1.1 200 OK
Date: Sun, 02 Sep 2018 19:48:31 GMT
Content-Type: text/html; charset=utf-8
[deletia]

perhaps I'm not supposed to see URIs like https://perma-archives.org/warc/89ZB-BWVX/http://www.library.umass.edu/ but the TimeMap link is wrong (though the TimeGate link works fine).

rebeccacremona:
Thanks! I'm afraid I don't understand.

In Perma's current setup, links of the form https://perma-archives.org/warc/89ZB-BWVX/http://www.library.umass.edu/ are for a specific memento within a single warc (89ZB-BWVX). 

[...]

PR #2699, opened 2020-02-03
Remove pywb playback; reimplement memento support

rebeccacremona:
This PR removes the option to play Perma warcs back via pywb, fully migrating to the Webrecorder-based playback that went live last summer. 

Since the CDXLine database is not required for Webrecorder-based playback, I prepare in this PR to drop that DB, refactoring the tests that rely on cdx entries so that they to calculate them on the fly, and dropping the model from `models.py`. We may, in a later PR, be able to remove the associate DB adapter and config; we cannot, however, until this PR's migrations are run.

Since our current, almost entirely broken (https://github.com/harvard-lil/perma/issues/2682, https://github.com/harvard-lil/perma/issues/2471) support for the memento protocol was implemented via pywb and the CDXLines db, this PR also reimplements memento support, making timemaps available at
```
/timemap/link/<url>
/timemap/json/<url>
/timemap/html<url>
```
and timegates available at
```
/timegate/<url>
```
Per offline discussion, we'll want to redirect at the Cloudflare level from the old routes to the new ones, since the old ones were at perma-archives.org, and the new ones will be at perma.cc.

Memento support required the introduction of a new field to the Link model, to store the surt (Sort-friendly URI Reordering Transform) of each submitted URL. I have included a data migration to populate the field for existing links, but am concerned about the efficiency of the migration; a "batch update" feature was added to Django recently, but Perma is using an old version. I'm not sure whether I am correct to be concerned or not...  to be discussed.


rebeccacremona:
image

rebeccacremona:
To be discussed before merging:
- the data migration
- caching behavior

rebeccacremona:
Once on stage, we should test against http://mementoweb.org/tools/validator

PR #2724, opened 2020-02-20
Handle content_type=None during capture.

rebeccacremona:
Sometimes, `warcprox` now returns `None` for `content_type` (https://github.com/internetarchive/warcprox/blob/932001c92115c4881a32d8b8f4d05c00a6efdaa7/warcprox/warcproxy.py#L212). Handle that.


rebeccacremona:
https://www.problemgambling.ca/gambling-help/gambling-information/what-is-problem-gambling.aspx does not return a `content-type` header. So, this will let us capture it... but, since we aren't recording that it is html, capture and playback experiences are suboptimal. Should we default to a content-type of `text/html; charset=utf-8` given that this is the web instead of `None`?

PR #2762, opened 2020-04-02
Wait longer for WR upload timeout, to account for large warcs.

rebeccacremona:
I am investigating how it can be so big (is limiting mechanism broken?), but, we have a 219MB Perma Link created yesterday. The first attempt to play back always fails, due to WR upload timeout. The second (automatic) attempt always works. So, let's increase the WR upload timeout, see if we can avoid the retry, even for our biggest warcs. (Why is upload happening more quickly on the retry?)


rebeccacremona:
Said warc is still timing out on the first try lol. Hmmm.

rebeccacremona:
re: warc size, see https://github.com/harvard-lil/perma/issues/2775

Issue #2779, opened 2020-04-20
URL hangs when loaded via requests; curl is fine.


rebeccacremona:
https://www.seattletimes.com/seattle-news/education/uw-students-publish-online-rape-list-out-of-hopelessness-community-reels/

rebeccacremona:
Repeatable on 3.5, 3.6, 3.7, and 3.8, in a clean env containing only the latest Requests. In all cases, curl works.

rebeccacremona:
This is now behaving differently. This morning, I could retrieve it locally via requests when not on the VPN. Now, VPN on or off doesn't seem to matter. Also, this morning, the attempt to connect hung, where as now, I'm getting 
```
>>> requests.get('https://www.seattletimes.com/seattle-news/education/uw-students-publish-online-rape-list-out-of-hopelessness-community-reels/')
Traceback (most recent call last):
  File "/root/.local/share/virtualenvs/perma_web-kkDLCj8f/lib/python3.7/site-packages/urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "/root/.local/share/virtualenvs/perma_web-kkDLCj8f/lib/python3.7/site-packages/urllib3/connectionpool.py", line 384, in _make_request
    six.raise_from(e, None)
  File "<string>", line 2, in raise_from
  File "/root/.local/share/virtualenvs/perma_web-kkDLCj8f/lib/python3.7/site-packages/urllib3/connectionpool.py", line 380, in _make_request
    httplib_response = conn.getresponse()
  File "/usr/lib/python3.7/http/client.py", line 1336, in getresponse
    response.begin()
  File "/usr/lib/python3.7/http/client.py", line 306, in begin
    version, status, reason = self._read_status()
  File "/usr/lib/python3.7/http/client.py", line 275, in _read_status
    raise RemoteDisconnected("Remote end closed connection without"
http.client.RemoteDisconnected: Remote end closed connection without response
```
Still no problem with curl or in browsers.

rebeccacremona:
Looks like this is actually a user agent thing:
```
>>> requests.get('https://www.seattletimes.com/seattle-news/education/uw-students-publish-online-rape-list-out-of-hopelessness-community-reels/', headers={'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:75.0) Gecko/20100101 Firefox/75.0'})
<Response [200]>
[...]

PR #2780, opened 2020-04-23
Tweak HTTP headers

rebeccacremona:
Some websites are evading capture... and after a couple of days investigating, it looks like it's a combination of our IP address, the headers we are sending (or not sending), and the order the headers are specified in. 

We aren't the first people to notice this sort of thing (e.g. https://sansec.io/labs/2017/05/02/http-header-order-is-important/).

After a couple days of experimenting, it looks like the headers and order in this PR are working... for now. Who knows how long that will last... but I'll plan to check on this weekly or so, and if it looks like this is insufficient or if the situation changes, I'll readdress.

Closes https://github.com/harvard-lil/perma/issues/2779

### DIFF
--- a/perma_web/perma/models.py
+++ b/perma_web/perma/models.py
@@ -1169,13 +1169,20 @@ class Link(DeletableModel):
     @cached_property
     def headers(self):
         try:
-            return requests.get(
-                self.ascii_safe_url,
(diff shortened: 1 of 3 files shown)


rebeccacremona:
Yup, no longer working.

PR #2811, opened 2020-06-24
Bump image version to get new Chrome installed at Circle

bensteinberg:
https://chromereleases.googleblog.com/2020/06/stable-channel-update-for-desktop_22.html


bensteinberg:
I realize it may not be necessary to do this every time; we can discuss.

rebeccacremona:
I love it. With every upgrade they do, perhaps less chance we crash on startup or seize taking screenshots. (I should look into updating selenium too.)

PR #2822, opened 2020-07-17
Bump image version to get new Chrome

bensteinberg:
As of this writing, this replaces Chrome 83.0.4103.116 with 84.0.4147.89 -- see https://chromereleases.googleblog.com/2020/07/stable-channel-update-for-desktop.html and https://chromium.googlesource.com/chromium/src/+log/83.0.4103.116..84.0.4147.89?pretty=fuller&n=10000

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,7 +30,7 @@ services:
     build:
       context: .
       dockerfile: ./perma_web/Dockerfile
-    image: perma3:0.57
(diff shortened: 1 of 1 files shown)


bensteinberg:
For future reference, this change does not affect the version of Chrome installed locally; you'll need to run `docker system prune` (or `docker system prune --all`?) before a build will get the new version.

PR #2824, opened 2020-07-28
Bump image version to get new Chrome at Circle CI

bensteinberg:
This upgrades Chrome from 84.0.4147.89 to 84.0.4147.105 at Circle CI. To get this upgrade locally, I had to run 
```
docker-compose down
docker system prune --all
docker-compose up -d
```
which seems heavy-handed; I did not find the `--no-cache` option mentioned in https://github.com/harvard-lil/perma/pull/2822#issuecomment-664456351

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,7 +30,7 @@ services:
     build:
       context: .
       dockerfile: ./perma_web/Dockerfile
-    image: perma3:0.58
+    image: perma3:0.59
     tty: true
     command: bash
     # TO AUTOMATICALLY START PERMA:


bensteinberg:
See https://chromereleases.googleblog.com/2020/07/stable-channel-update-for-desktop_27.html and https://chromium.googlesource.com/chromium/src/+log/84.0.4147.89..84.0.4147.105?pretty=fuller&n=10000

rebeccacremona:
I forced the upgrade locally with:
```
docker-compose build --no-cache
docker-compose up -d
```

bensteinberg:
Neat, thanks!

PR #2825, opened 2020-07-28
Script for updating Chrome, and its first effects

bensteinberg:
I considered wrapping the script in git moves to branch and commit, but I'm not sure that's a good idea.

This will break if we change the Perma image numbering scheme or the capture user agent much.


rebeccacremona:
Very cool.

PR #2848, opened 2020-10-07
Registration link expiry

rebeccacremona:
This PR changes the way Perma handles account activation. In recent LIL projects, we reuse Django's "[reset password](https://github.com/django/django/blob/master/django/contrib/auth/tokens.py)" logic both for initial account activation and for standard password resets. This has a number of advantages, including a) Django password reset links can only be used once, and b) Django password reset links expire after `settings.PASSWORD_RESET_TIMEOUT_DAYS`.

I updated the activation email to include a message about how long the link is valid for. I want to do that for the password reset email too... but the whole sign up flow needs a little untangling and modernization, so will do it all at once in a bit.

Sample:
```
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 8bit
Subject: A Perma.cc account has been created for you
From: webmaster@localhost
To: new@new.com
Date: Wed, 07 Oct 2020 15:53:41 -0000
Message-ID: <160208602157.4699.16889732655275746156@2af27e90f434>

Thanks for signing up for Perma.cc!

In order to complete your account activation, please confirm your email address and set a password. It’s as simple as clicking the link below. The link will be valid for the next 72 hours.

http://localhost:8000/password/reset/Mzg/5kj-376caad7a6e87252f36c

Thanks,
The Perma.cc Team
```

This PR includes a migration that drops the custom confirmation code from the DB. As ever... Perma migrations might be slow, but I hope dropping a column is not.


rebeccacremona:
Oh... before merging, I should check and see what happens if people click on an old-style link... it shouldn't 404; they should get directed to the "invalid link" flow....

rebeccacremona:
Okay: users who follow an old-style activation link will 1) get redirected to a page that says the link is invalid, and offers a button for getting a new link; 2) that button takes them to the reset password page; 3) when they enter their email address, they'll get a reset-password email like this:
```
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit
Subject: Password reset for Perma.cc
From: webmaster@localhost
To: user22@example.com
Date: Wed, 07 Oct 2020 17:54:46 -0000
Message-ID: <160209328676.11322.2128529068637558830@2af27e90f434>

You're receiving this e-mail because you requested a password reset for your user account at Perma.cc

Please go to the following page and choose a new password:
http://perma.test:8000/password/reset/MzY/5kj-abbc32ca659b89017b85

The link will be valid for the next 72 hours.

Thanks,
The Perma.cc Team
```
That's not identical to the text of the email already-activated users get, when they reset their password.... but I don't think I can fix all this today, without totally re-addressing how Perma sets `is_active`, `is_confirmed`.

PR #2855, opened 2020-10-13
Registrar default privacy

rebeccacremona:
This PR adds a field to the Registrar model, indicating whether they want newly-created orgs in their network to default to public or private Perma Links. Public is the norm; on request, administrators can set a Registrar to default to private, either via the Django admin or via the manage/registrars page.

image

Registrars can't configure this themselves, just like they can't change the Registrar's name, email address, or website: the manage/registrars page is admin-only.

I also added info to the docs about the configuring a single org's default privacy, and a link to contact us about registrar-level config: `/docs/libraries#privacy`

image


bensteinberg:
Good call on "registrators"

PR #2857, opened 2020-10-14
Try authenticated pulls from Docker Hub

bensteinberg:


bensteinberg:
I believe for this to work, we'd have to enable "Pass Secrets to Builds From Forked Pull Requests" at Circle CI, which I think is not acceptable.

bensteinberg:
See https://circleci.com/docs/2.0/oss/#pass-secrets-to-builds-from-forked-pull-requests

PR #2860, opened 2020-10-14
Limit login attempts

rebeccacremona:
This PR adds the [`django-axes`](https://github.com/jazzband/django-axes) library, which seems to be the most common way for Django apps to limit login attempts.

Config: 
- in a pinch, we can turn it off by setting `AXES_ENABLED = False`
- I have it set to `AXES_FAILURE_LIMIT = 6` and `AXES_COOLOFF_MINUTES = 30`, in accordance with the feedback we received
- For now I have it set to limit by username only, and ignore IP addresses (though it tracks IP addresses and user agents and records them in the database, both for successful and unsuccessful login attempts). That might be the wrong decision. To be discussed. Motivation: I want to see how this IP code, which uses a helper library, etc., interacts with our custom IP proxy middleware, before we use that for anything. There's all kind of config for apps behind proxies; we might need to tweak some things before this code is looking at the right IP address.
- I have it set so that, if you get it right, we forget about previous errors

I also set things up so that, when you get locked out, we offer to send you a password reset email, and if you reset your password, you get unlocked.

image

You get the same lockout page whether or not the email address you are attempting to log in as is or is not in reality associated with a Perma account, just like with the password reset page. (So, it doesn't reveal whether a given email valid or not.)

I do not have anything in place for forgetting about old failed attempts. E.g., if you fail three times today, go away sad, and come back next week and fail another three times, that will be a total of 6 and you'll be locked out.

It can track login attempts using either the database + standard Django models or a redis cache. I initially set it up to use redis.... but, it turns out that the `reset` commands currently only work with the database backend. I would be surprised if we have enough login activity from real users for this to cause a problem...... If we have enough mischievous bot traffic, despite cloudflare, to cause a problem, I have no idea. If the AccessAttempts table is becoming humongous or we notice performance issues on login, then I'll see if I can fix the library (which I'm kind of tempted to do anyway).

The additional config to use redis, if that becomes a possibility in the future, would look something like:
```
AXES_HANDLER = 'axes.handlers.cache.AxesCacheHandler'
AXES_CACHE = 'axes'
CACHES['axes'] = {
    "BACKEND": "django_redis.cache.RedisCache",
    "LOCATION": "redis://perma-redis:6379/2",  # don't clobber default cache (/0) or Celery (/1) if in use
    "OPTIONS": {
        "CLIENT_CLASS": "django_redis.client.DefaultClient",
        "IGNORE_EXCEPTIONS": False
    }
}
```

### DIFF
--- a/perma_web/perma/settings/deployments/settings_common.py
+++ b/perma_web/perma/settings/deployments/settings_common.py
@@ -196,7 +196,7 @@ AXES_LOCK_OUT_AT_FAILURE = True
 AXES_COOLOFF_MINUTES = 30
 AXES_COOLOFF_TIME = 'perma.utils.cooloff_time'
 AXES_ONLY_USER_FAILURES = True  # If True, only lock based on username, and never lock based on IP if attempts exceed the limit. Otherwise utilize the existing IP and user locking logic. Default: False
-AXES_RESET_ON_SUCCESS: True  # If True, a successful login will reset the number of failed logins. Default: False
+AXES_RESET_ON_SUCCESS = True  # If True, a successful login will reset the number of failed logins. Default: False
 
 
 AUTH_PASSWORD_VALIDATORS = [
--- a/perma_web/perma/settings/deployments/settings_common.py
+++ b/perma_web/perma/settings/deployments/settings_common.py
@@ -115,6 +115,12 @@ MIDDLEWARE = (
     'simple_history.middleware.HistoryRequestMiddleware',  # record request.user for model history
     # Uncomment the next line for simple clickjacking protection:
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
+    # AxesMiddleware should be the last middleware in the MIDDLEWARE list.
+    # It only formats user lockout messages and renders Axes lockout responses
+    # on failed user authentication attempts from login views.
+    # If you do not want Axes to override the authentication response
+    # you can skip installing the middleware and use your own views.
+    'axes.middleware.AxesMiddleware',
 )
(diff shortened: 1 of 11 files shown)


perma_web/perma/settings/deployments/settings_common.py
```diff
@@ -171,6 +178,27 @@
...
+AXES_LOCKOUT_TEMPLATE = 'registration/login_attempts_exceeded.html'
+
+# Axes, configure behavior
+AXES_ENABLED = True
+AXES_FAILURE_LIMIT = 6
+AXES_LOCK_OUT_AT_FAILURE = True
+AXES_COOLOFF_MINUTES = 30
+AXES_COOLOFF_TIME = 'perma.utils.cooloff_time'
+AXES_ONLY_USER_FAILURES = True  # If True, only lock based on username, and never lock based on IP if attempts exceed the limit. Otherwise utilize the existing IP and user locking logic. Default: False
+AXES_RESET_ON_SUCCESS: True  # If True, a successful login will reset the number of failed logins. Default: False
```
bensteinberg: `:` should be `=`

rebeccacremona: Oh wow! I'm weirded out that nothing seems broken with it this way.

jcushman: Whoa, it's a type annotation!

```
>>> AXES_RESET_ON_SUCCESS: True
>>> __annotations__
{'AXES_RESET_ON_SUCCESS': True}
```

See also https://www.python.org/dev/peps/pep-0526/#other-uses-of-annotations where they're like "but don't do that." :)

rebeccacremona: 🙌

PR #2861, opened 2020-10-14
Add a trap for bots on the signup forms.

rebeccacremona:
As with the [contact form](https://github.com/harvard-lil/perma/pull/2703), but with a realistic input `name`, "telephone", as per advice on the internet.


jcushman:
It might help to rename an existing required field as well, so any existing "submit these three fields" scripts will fail.

PR #2864, opened 2020-10-22
Try using our own registry

bensteinberg:


bensteinberg:
Need to change Dockerfile, too....

bensteinberg:
It is redundant and possibly confusing to replicate all the necessary images from Docker Hub to our own registry, but it may save us from rate-limiting by IP in CI, and having our own registry may be useful in other contexts.

bensteinberg:
Closing and reopening to kick Travis....

bensteinberg:
The pending Travis CI build at .org is spurious and can be ignored.

If we choose to do this, we'll have to figure out a workflow or automation for replicating images to the registry when we change any Dockerfile or docker-compose*.yml.

PR #2870, opened 2020-11-09
Add fab task, count_pending_ia_links

bensteinberg:
I got the syntax for the query from https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/tasks.py#L1597-L1599 rather than immediately above this new function, https://github.com/harvard-lil/perma/blob/develop/perma_web/fabfile/dev.py#L341-L345

### DIFF
--- a/perma_web/fabfile/dev.py
+++ b/perma_web/fabfile/dev.py
@@ -348,6 +348,15 @@ def upload_all_to_internet_archive():
         upload_to_internet_archive(link.guid)
 
 @task
+def count_pending_ia_links():
+    from perma.models import Link
+
+    count = Link.objects.visible_to_ia().filter(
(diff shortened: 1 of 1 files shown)


rebeccacremona:
> I got the syntax for the query from https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/tasks.py#L1597-L1599 rather than immediately above this new function, https://github.com/harvard-lil/perma/blob/develop/perma_web/fabfile/dev.py#L341-L345

I think that's correct.... The celery task is optimized and is in hourly use. The fab task, which I didn't totally realize was there, is outdated. I'll look into cleaning up our fabfile, either deleting old things or moving them into a clearly-labeled section.

PR #2882, opened 2021-01-08
Package updates

bensteinberg:
These are the result of `docker-compose exec web pipenv lock`. `chardet`, `pyopenssl`, and `importlib-metadata` have major version upgrades; I am going to look at the changelogs, but will kick off testing by filing this PR first 😛


bensteinberg:
This is now reduced to the update of a single package, which required manual editing of `Pipfile.lock`....

rebeccacremona:
Oof. Thx. I'll look into what's breaking, when the rest of the packages are upgraded. And probably take us back over to pip-tools.

PR #2889, opened 2021-02-02
Package updates

bensteinberg:


bensteinberg:
The additional change in f43bcaf means the save of a `Sponsorship` is less efficient than it might otherwise be. I'm sorry to say that I haven't pinned down which change in `django-model-utils` is at fault here, but in any event, I believe that `FieldTracker.has_changed` no longer returns `True` after the test's submission of the forms in `test_views_user_management.py::UserManagementViewsTestCase::test_admin_user_can_deactivate_active_sponsorship`, `::test_admin_user_can_reactivate_inactive_sponsorship`, `::test_registrar_user_can_deactivate_active_sponsorship`, and `::test_registrar_user_can_reactivate_inactive_sponsorship`.

bensteinberg:
I believe all of our other uses of `has_changed`, in `models.py` and `tasks.py`, are called, correctly, prior to `save()`.

bensteinberg:
Thanks to @rebeccacremona for the additional and better approach in 18edbae.

PR #2905, opened 2021-04-06
Minor updates to Django and other packages

bensteinberg:


rebeccacremona:
LGTM: the 1.99 -> 2.0.0 for the internetarchive package doesn't seem to include breaking changes. I don't think they are really doing semantic versioning.

PR #2906, opened 2021-04-06
Remove remaining evidence of phantomJS

bensteinberg:
I'm not at all confident about the changes in `npm-shrinkwrap.json`, and I was expecting to see changes in `static/`.


perma_web/perma/tasks.py
```diff
@@ -695,10 +679,6 @@ def get_srcset_image_urls(dom_tree):
             src = src.strip().split()[0]
             if src:
                 urls.append(src)
-    if settings.CAPTURE_BROWSER != 'PhantomJS':
-        # Get src, too: Chrome (and presumably Firefox) doesn't do this automatically, although PhantomJS does.
-        for el in dom_tree('img[src]'):
-            urls.append(el.attrib.get('src', ''))
```
rebeccacremona, line 701: Whoops, we need these lines.

bensteinberg: oh whoops....

PR #2907, opened 2021-04-09
Bump payments image version to 0.31

bensteinberg:


bensteinberg:
This is mildly interesting: around the time that Cloudflare makes a change to its list of IP addresses (which it just announced), the results of e.g. `test_cloudflare_ip4_cache` are indeterminate; just now, I've got it to pass and fail with the new list.

rebeccacremona:
Hmm, maybe we should be sorting in some way. (Though I note that when I visit https://www.cloudflare.com/ips-v4, I see the old list, not the new one, so sorting probably isn't the problem.)

bensteinberg:
Note that the change in 49ff881 from running `fab dev.update_cloudflare_cache` exactly matches the change in Cloudflare's announcement:
> If your security model relies on allowing a list of trusted Cloudflare IPs from cloudflare.com/ips (or via API) on your origin, please make the following changes to your allow list by May 7, 2021. This change is safe to make today. 
>
> Remove: 
> 104.16.0.0/12 
> 
> Add: 
> 104.16.0.0/13 
> 104.24.0.0/14 
> 
> This change delists the 104.28.0.0/14 prefix, which is no longer in use by Cloudflare infrastructure. These addresses will be repurposed for use with our Gateway and WARP (secure web gateway and VPN) products, and may carry traffic from untrusted sources in the future.

bensteinberg:
> Hmm, maybe we should be sorting in some way

I don't think this is a sorting issue -- I think it might be a cached upstream response?

bensteinberg:
I'm not sure where to put this, but it might as well go here until I figure out a better place. In addition to the possible caching issue with Cloudflare IP address ranges, there are multiple sources of truth for the list of ranges. Right now, https://www.cloudflare.com/ips/ and https://www.cloudflare.com/ips-v4 say one thing, but https://api.cloudflare.com/client/v4/ips says another. The latter is the older range. I'm guessing these will be reconciled on or around May 7, as specified in CF's email of April 12. Note that the mechanism in Perma uses https://www.cloudflare.com/ips-v4 and https://www.cloudflare.com/ips-v6, while our AWS Lambda for updating Security Groups uses https://api.cloudflare.com/client/v4/ips.

PR #2916, opened 2021-04-28
Update django-debug-toolbar

bensteinberg:


rebeccacremona:
Django debug toolbar 1.11.1 won't actually work without some additional changes, but I think that's okay: I'll merge this and then will follow up with a PR that gets the toolbar working later.

PR #2919, opened 2021-04-29
Use a cache-busting trick to force Chrome update in image.

rebeccacremona:
This should significantly reduce the time it takes to run `update_chrome.sh` (once your machine has built perma3:0.97), because you no longer have to build with `--no-cache` to look for updates, and because I'm moving the Chrome installation layer lower in the Dockerfile, since we do Chrome updates more often than python or node package updates. 

This also makes sure that other developers get the update: with the previous setup, the newly-tagged image produced when they pulled the code and ran `docker-compose up` would use their cached chrome layer, if one was present.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,7 +30,9 @@ services:
     build:
       context: .
       dockerfile: ./perma_web/Dockerfile
-    image: perma3:0.96
+      args:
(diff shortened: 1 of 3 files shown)


bensteinberg (approved):
Great! Thanks!

Issue #2914, opened 2021-04-26
Docker upgrade makes test fail locally


rebeccacremona:
```
====================================================================== FAILURES ======================================================================
______________________________________________ LinkValidationTestCase.test_should_reject_unloadable_url ______________________________________________

self = <perma_web.api.tests.test_link_validation.LinkValidationTestCase testMethod=test_should_reject_unloadable_url>

    def test_should_reject_unloadable_url(self):
        self.rejected_post(self.list_url,
                           user=self.org_user,
                           # http://stackoverflow.com/a/10456069/313561
>                          data={'url': 'http://0.42.42.42/'})
```
because in my Docker environment,
```
HLSC02VK05ZHTDG:perma rcremona$ d curl -i http://0.42.42.42/
HTTP/1.1 500 Internal Server Error
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Mon, 26 Apr 2021 16:24:13 GMT
Content-Length: 17

host unreachable
```
compared with just in the terminal
```
HLSC02VK05ZHTDG:perma rcremona$ curl http://0.42.42.42/
curl: (7) Couldn't connect to server.
```
Hmmm.

rebeccacremona:
Fixed by Ben in https://github.com/harvard-lil/perma/pull/2920

PR #2920, opened 2021-05-04

bensteinberg:
These are minor updates, except for `django-simple-history` and `importlib-metadata`. I had to make one change to get `test_should_reject_unloadable_url` in `api.tests.test_link_validation.LinkValidationTestCase` to pass, and I don't understand why yet. The failure was `AssertionError: 201 != 400`. I'm continuing to look into it.


rebeccacremona:
The test thing could easily be https://github.com/harvard-lil/perma/issues/2914

bensteinberg:
Oh right! I saw that go by and forgot about it. Thanks.

rebeccacremona (approved):

PR #2923, opened 2021-05-13
Package updates

bensteinberg:
This is primarily a point version update of Django:

https://docs.djangoproject.com/en/3.2/releases/2.2.23/

I manually removed major version upgrades of Werkzeug and django-webpack-loader -- the latter broke many tests.

https://werkzeug.palletsprojects.com/en/2.0.x/changes/#version-2-0-0
https://github.com/django-webpack/django-webpack-loader/blob/master/CHANGELOG.md


rebeccacremona:
(Future self: see Slack for a discussion about manually updating Pipfile.lock vs specifying supported minor versions in Pipfile)

PR #2929, opened 2021-05-21
Hopefully fix image on Twitter cards

rebeccacremona:
This is presently my only guess for why the Twitter cards for the home page, about, etc. are not displaying the intended image.

Twitter caches card info once a week, so even if this works, we shouldn't expect already-posted Tweets' cards to update any time soon. We should, however, be able to see a fresh card by making a bitly link and Tweeting about that.

If this works, we can talk about whether we want to allow Twitterbot to access playback pages as well. Either a) they didn't initially respect robots.txt for this activity, or b) we didn't notice we broke this in https://github.com/harvard-lil/perma/commit/7ea2b5d8f5f333a18caf50d3d4cea5763ad15f0f#diff-079d9f44880048d8baeeee35385e87378256a2072c951e6a6a6b4f443953f01d, about a year after https://github.com/harvard-lil/perma/commit/c21ab341621cb002a49304f384fd729d3443f9a2


bensteinberg:
Closing and reopening to see if CI is working

PR #2932, opened 2021-06-03
DB cleanup

rebeccacremona:
In preparation for moving from MySQL to Postgres, let's tidy up a bit.

## Legacy tables
I loaded a dump of the production database from January 2021 and inspected it locally. In reviewing a complete list of its tables, I saw a bunch that seemed to be relics of the past:
```
EXPECTED TABLES

| auth_group                    |
| auth_group_permissions        |
| auth_permission               |

| axes_accessattempt            |
| axes_accesslog                |

| django_admin_log              |
| django_content_type           |
| django_migrations             |
| django_session                |
| django_site                   |

| lockss_mirror                 |

| perma_apikey                  |
| perma_capture                 |
| perma_capturejob              |
| perma_folder                  |
| perma_genericstringtaggeditem |
| perma_historicallink          |
| perma_historicallinkuser      |
| perma_historicalorganization  |
| perma_historicalregistrar     |
| perma_link                    |
| perma_link_folders            |
| perma_linkbatch               |
| perma_linkuser                |
| perma_linkuser_organizations  |
| perma_minutestats             |
| perma_organization            |
| perma_registrar               |
| perma_sponsorship             |
| perma_uncaughterror           |
| perma_weekstats               |

| taggit_tag                    |
| taggit_taggeditem             |


LEGACY TABLES, POSSIBLE TO DELETE?

| celery_taskmeta               |
| celery_tasksetmeta            |

| djcelery_crontabschedule      |
| djcelery_intervalschedule     |
| djcelery_periodictask         |
| djcelery_periodictasks        |
| djcelery_taskstate            |
| djcelery_workerstate          |

| perma_asset                   |
| south_migrationhistory        |
| tastypie_apiaccess            |
| tastypie_apikey               |
| thumbnail_kvstore
``` 
It turns out that yes, all those tables can and should be deleted, so I added a [migration to do so,  via hard-coded SQL](https://github.com/harvard-lil/perma/pull/2932/files#diff-b543a65eb5b5ea79696569941f8645f738b35808fe39723829cae3ec7258192b).

### `celery_*`
The `celery_*` tables aren't needed if you aren't using the DB as a celery backend.... which we aren't. To make extra sure:

The `celery_taskmeta` table hasn't been in use since 2014, and only contains rows for about a day: I think this was never truly in use:
```
select date_done from celery_taskmeta order by date_done limit 1
    -> ;
+---------------------+
| date_done           |
+---------------------+
| 2014-07-30 08:00:00 |
+---------------------+
1 row in set (0.011 sec)

select date_done from celery_taskmeta order by -date_done limit 1;
+---------------------+
| date_done           |
+---------------------+
| 2014-07-31 17:29:20 |
+---------------------+
1 row in set (0.005 sec)
```

And, the `celery_tasksetmeta` table is empty:
```
select count(*) from celery_tasksetmeta;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.001 sec)
``` 

### `djcelery_*`
The six `djcelery_*` tables can be dropped because [we no longer use django-celery](https://github.com/harvard-lil/perma/commit/d67f74d1817991d5e07479a776eecc4137a48b48#diff-8551b027d76f1ba183c6a39a0339a7ee89e211cd33dfdbdb90ab1b5897d1fab9L130): modern versions of celery don't need it any more.

### `perma_asset`
An earlier version of Perma made use of an `Asset` model instead of today's `Capture` model. After that migration, [Jack removed the code for `Asset`](https://github.com/harvard-lil/perma/pull/1369), but the table stuck around. It has data, but we don't need it anymore.
```
select count(*) from perma_asset;
+----------+
| count(*) |
+----------+
|   123398 |
+----------+
1 row in set (0.688 sec)
```

### `south_migrationhistory`
Isn't it delightful? We used Django for Perma before migrations were part of the framework, and instead used a migration package called South. [Jack upgraded us to Django 1.7](https://github.com/harvard-lil/perma/pull/837) and switched us over to the built-in migration system we now know and love, making this table obsolete. I'm sure we hung onto it just in case, but we don't need to anymore. 
```
select count(*) from south_migrationhistory;
+----------+
| count(*) |
+----------+
|       36 |
+----------+
1 row in set (0.005 sec)
```

### `tastypie_*`
Now we [use Django Rest Framework instead of TastyPie for the API](
https://github.com/harvard-lil/perma/pull/2005/files#diff-530420a43fa3362df78bc8c31ab3c97fe30233769e1d2b0d91b4d48dfbafcba4).

Jack migrated users' API from tastypie's table into a new perma table in a migration:
```
    try:
        cursor.execute("INSERT INTO perma_apikey SELECT * FROM tastypie_apikey")
    except django.db.utils.ProgrammingError:
        pass  # old tastypie api_key table doesn't exist for this deployment
```
We still have the old table, and it still has data in it, but we don't need it any more. 
```
select count(*) from tastypie_apikey;
+----------+
| count(*) |
+----------+
|      208 |
+----------+
1 row in set (0.011 sec)
```
The other tastypie table, whose purpose I don't know, is empty:
```
select count(*) from tastypie_apiaccess;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.007 sec)
```

### `thumbnail_kvstore`
We used to have thumbnail images for each Perma Link. We don't anymore, though we've talked about re-implementing it, somehow. In any case, we no longer have any thumbnails in this table, we just have the empty table lying around:
```
select * from thumbnail_kvstore;
Empty set (0.002 sec)
```

## HistoricalLinkUser and user ids
An early experiment with [dumpdata](https://docs.djangoproject.com/en/3.2/ref/django-admin/#dumpdata) failed because the HistoricalLinkUser table made references to users in its `id` column for which there were no longer corresponding rows in the LinkUser table: the users had been deleted.

Via the admin user / org user / registrar user manage interface, you can in fact truly delete users who have not yet confirmed their email address. You can also delete users via the django admin (for instance, all those many spam users I deleted). While my first instinct would be to expect the HistoricalLinkUser rows to get deleted too via CASCADE... this makes a lot of sense: another row is added when you do the deletion, tracking who did it, in a recoverable way. (The ability to specify deletions [should also delete the historical record](https://django-simple-history.readthedocs.io/en/latest/historical_model.html#deleting-historical-record) was added in 2.5.0 (2018-10-18), several years after this was set up.)  I took a look at the rows in question, and found that while there are a lot of them, there's only one more such row on Dec 15 as there was back in June, which is a good sign. And of the rows that include a history_user_id, the vast majority have mine, which suggests to me that a big chuck of these orphaned rows are from the spam user cleanup project. Nothing seems amiss.

I no longer have a record of precisely what error message I received during dumpdata/loaddata (indeed, or even which command errored out). I wrote down that there was a foreign key constraint problem, but that turns out not to be the case: there is no foreign key constraint on the `id` column of `perma_historicallinkuser` locally or on prod. So..... what I saw probably was just a problem with dumpdata/loaddata, and will not surface when using `pgloader`.... which empirically seems to be the case, because Ben did not report a problem with his initial experimental run.

So.... I think we're good.

For record keeping: before coming to that conclusion, locally, I ran a command in the dbshell to delete the orphaned rows manually:
```
delete from perma_historicallinkuser where id not in (select id from perma_linkuser);
2:21
Query OK, 34683 rows affected (19.670 sec)
```
Then I decided the thing to do was delete these rows via an elidable RunSQL migration, and set `cascade_delete_history=True` going forward. I made a commit to that end.....

.... but subsequently removed it. 

There doesn't seem to be any need to do anything.

## Squashing migrations

I decided to squash the migrations.

### Why?

1) We already had enough accumulated in 2018 that we found a noticeable improvement in test setup speed by [running tests without migrations](https://github.com/harvard-lil/perma/commit/6ea1d489181685753ed5b628b5b6370d3a4b6eba), and 3 years of adding on can't have improved things.

2) Several of the migrations run hard-coded SQL in order to migrate data from a legacy architecture to a new one. That SQL is by necessity MySQL-flavored, and so will not run on postgres. I -could-, when we move to postgres, comment them out and replace them with the equivalent postgres-flavored SQL, but that seems wicked silly: all those moves were necessary at the time, but we've long since advanced past the point where they matter. Most could never be reversed in the first place. So, I decided instead to mark them all elidable and then squash, in order to produce a nice, tidy migration file that can be used with new postgres databases (for example, by test runners and when setting up new dev instances.)

We got all the way down to 24 operations!

### What's an elidable migration?

From [django-squash](https://pypi.org/project/django-squash/)
> “Elidable” means the operation can be eliminated when you squash migrations. For example, if you split an existing table into two or more tables, you must populate the new tables from the old. Once that is done, you never need to do it again, even if you are creating a brand new DB. This is because the source table is empty when creating a new DB so it’s pointless to populate those two new tables from the empty old one.
>
> “Non-elidable” means the operation cannot be eliminated when you squash migrations. Perhaps you are creating a very special index on a table that cannot be configured when describing the model. You have to write raw SQL that your flavor of DB understands. If you are creating a new DB, that operation must be run sometime after that table has been created.

All of our RunPython and RunSQL migrations are elidable: when creating a new DB, none of them need to be run.

I fixed that, going back. I'll now get in the habit of setting it routinely.

## HistoricalLink

At first, I thought we should take this opportunity to delete the HistoricalLink table, because it's ginormous and not that useful (see followup questions from Ben in the conversation part of this PR), but we're not sure we want to do that after all. The migration seems to go fine even with it. So, kicking that can a little further down the road.


bensteinberg:
FWIW, we are now running `./manage.py clean_old_history --days 60 --auto` once a month on H2O.

bensteinberg:
> I don't think we want to run the thing that deletes old history: for Links, I think we either want all the history, or none of it.

What is the scenario where we'd want to look at HistoricalLink?

rebeccacremona:
> > I don't think we want to run the thing that deletes old history: for Links, I think we either want all the history, or none of it.
> 
> What is the scenario where we'd want to look at HistoricalLink?

The history of public/private, changes maybe? The history of which folders it belonged to (which we might not even be tracking, I think)? Who changed its user-supplied text fields, to and from what (which we might be tracking, in excess)? What else.....

rebeccacremona:
(Codecov, I think you are somewhat confused)

bensteinberg (approved):
Looks great! Thank you!

PR #2942, opened 2021-06-22
Ask crawler to buzz off

rebeccacremona:


bensteinberg:
https://www.seroundtable.com/google-does-not-use-nutch-30210.html (https://perma.cc/QC3Y-4ZXL)

Issue #2845, opened 2020-10-05
404 via Webrecorder, success via client-side


rebeccacremona:
Why are all the URLs (except SeattleTimes) in https://app.frontapp.com/open/msg_hcoe3lz failing to play back via Webrecorder, but succeeding via client side? Indexing problem?

rebeccacremona:
Local stack trace:
```
recorder_1        | Traceback (most recent call last):
recorder_1        |   File "/usr/local/lib/python3.7/site-packages/pywb-2.4.1-py3.7.egg/pywb/recorder/multifilewarcwriter.py", line 208, in _write_to_file
recorder_1        |     new_size - start)
recorder_1        |   File "./webrecorder/rec/webrecrecorder.py", line 286, in add_urls_to_index
recorder_1        |     add_urls_to_index(stream, params, filename, length))
recorder_1        |   File "/usr/local/lib/python3.7/site-packages/pywb-2.4.1-py3.7.egg/pywb/recorder/redisindexer.py", line 64, in add_urls_to_index
recorder_1        |     writer_cls=params.get('writer_cls'))
recorder_1        |   File "/usr/local/lib/python3.7/site-packages/pywb-2.4.1-py3.7.egg/pywb/indexer/cdxindexer.py", line 316, in write_cdx_index
recorder_1        |     for entry in entry_iter:
recorder_1        |   File "/usr/local/lib/python3.7/site-packages/pywb-2.4.1-py3.7.egg/pywb/indexer/archiveindexer.py", line 339, in __call__
recorder_1        |     for entry in entry_iter:
recorder_1        |   File "/usr/local/lib/python3.7/site-packages/pywb-2.4.1-py3.7.egg/pywb/indexer/archiveindexer.py", line 212, in join_request_records
recorder_1        |     for entry in entry_iter:
recorder_1        |   File "/usr/local/lib/python3.7/site-packages/pywb-2.4.1-py3.7.egg/pywb/indexer/archiveindexer.py", line 188, in create_record_iter
recorder_1        |     record.raw_stream)
recorder_1        |   File "/usr/local/lib/python3.7/site-packages/pywb-2.4.1-py3.7.egg/pywb/warcserver/inputrequest.py", line 262, in __init__
recorder_1        |     data = cgi.FieldStorage(**args)
recorder_1        |   File "/usr/local/lib/python3.7/cgi.py", line 489, in __init__
recorder_1        |     self.read_multi(environ, keep_blank_values, strict_parsing)
[...]

PR #2962, opened 2021-07-23
Special 404 page for a Link's submitted URL during playback.

rebeccacremona:
We have received reports about perma links that, when visited by one user, show the "Link Not Found" archive error page, that when visited subsquently or in a different browser are fine. A diagnosis has been elusive. But, in the meantime, we can show a custom error page, when the target/submitted URL of a perma link gets a 404 error from Webrecorder.

image

This may display in other situations as well, such as https://github.com/harvard-lil/perma/issues/2845, but I think that's okay: so long as WR returns 404 and not 500 for errors, there's not a lot we can do on this end to serve up nuanced responses for each situation.

### DIFF
--- a/perma_web/perma/templates/archive/archive-error.html
+++ b/perma_web/perma/templates/archive/archive-error.html
@@ -7,16 +7,31 @@
 {% block mainContent %}
 
   <div class="record-message">
-    {% if status == '404 Not Found' %}
+
+    <div id="loading">...loading</div>
+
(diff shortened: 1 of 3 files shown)


rebeccacremona:
If we can manage to observe this problem ourselves, let's take careful note of what steps (if any), causes a successful playback in the same browser tab. There's a possibility that adding a cache-busting `?bust={timestamp}` to the target of the iframe might work; if we think we want to try that, I can add a button to this template.

PR #2966, opened 2021-08-17
Update Chrome

bensteinberg:
This updates Chrome from 92.0.4515.131 to 92.0.4515.159:

https://chromereleases.googleblog.com/2021/08/stable-channel-update-for-desktop.html

The addition of `--allow-releaseinfo-change` to the Dockerfile is due to the [release on August 14](https://www.debian.org/News/2021/20210814) of Debian 11 ("bullseye"). If I didn't have cache locally for all the lines including `apt-get update` above this one, I'd have needed to add it there, as well. I'm not sure whether CI will have cache for those lines, but my guess is no. I will put this PR in, expecting failure, then decide whether to add `--allow-releaseinfo-change` everywhere, or begin to switch from buster to bullseye -- which will need to happen at some point, though probably not within the first week of the release. ;)

The error messages, without allowing release info change, look like `E: Repository 'http://security.debian.org/debian-security buster/updates InRelease' changed its 'Suite' value from 'stable' to 'oldstable'`, btw.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,8 +31,8 @@ services:
       context: .
       dockerfile: ./perma_web/Dockerfile
       args:
-        chrome-layer-cache-buster: 006d813c-8350-4486-9339-9e816bb20183
-    image: perma3:0.109
+        chrome-layer-cache-buster: f51c8c61-4b8d-4aef-93ec-282e99f99e55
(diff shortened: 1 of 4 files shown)


bensteinberg:
Rather than changing the order of CF IPs, I suppose we could compare sets.

PR #2969, opened 2021-09-09
Special delay for iurisprudentia.

rebeccacremona:


rebeccacremona:
This is failing because of https://github.com/pypa/setuptools/issues/2775

rebeccacremona:
See also https://github.com/pypa/pipenv/issues/2364

PR #2999, opened 2021-12-07

bensteinberg:
https://www.djangoproject.com/weblog/2021/dec/07/security-releases/


rebeccacremona:
There are some big jumps here, including the adoption of new version numbering schemes (`cryptography` from 3.4.8 to 36.0.0). I'm going to review all the changelogs before merging, even if the tests pass.

Issue #2690, opened 2020-01-14
Look for py2/3 remarks


rebeccacremona:
We are now solidly in python 3.
Look for and clean up any things like:
```
    # In Python 3: def __str__(self):
    def __unicode__(self):
        return u"LinkBatch %s" % (self.pk,)
```

matteocargnelutti:
I would consider the following out of scope for this issue: https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/templatetags/current_query_string.py#L17 

Closing.

PR #3006, opened 2022-01-04
Cleanup python 2

holtchesley:
- Remove unnecessary `u"..."`s
 - We are now living in the future
 - F-strings where appropriate
 - Avoid abc deprecation warning
 - `iteritems` -> `items`


holtchesley:
Re: #2690

rebeccacremona (changes requested):
Awesome! Thanks for doing this!

Can you revert the changes to the files in `services/docker/webrecorder`? Those files are actually verbatim copies of files in the Webrecorder/Conifer codebase, with Perma-specific patches delineated with comments. I think it's probably best if we keep that code in its original state, so that diffs against Webrecorder are cleaner.

rebeccacremona (approved):

PR #3022, opened 2022-01-26
reworked invalid capture messages and control statements

ChefAndy:
Let me know what changes, if any, you'd like me to make.


perma_web/api/views.py
```diff
@@ -388,35 +388,47 @@ def post(self, request, format=None):
...
+
         # Disallow creation of links in top-level sponsored folder
         if folder.is_sponsored_root_folder:
-            error = "You can't make links directly in your Sponsored Links folder. Select a folder belonging to a sponsor."
-            raise_invalid_capture_job(capture_job, error)
+            message = message_template.format_map({
+                'error': 'Folders no longer sponsored by a registrar are read-only.',
+                'resolution': 'Select a folder that is currently sponsored.'
+            })
+            raise_invalid_capture_job(capture_job, message)
```
rebeccacremona, line 409: This error message seems to describe a folder that was once sponsored, but is no longer, and thus is read-only. But this bit of code is for the parent of all sponsored folders, the container. The container may not contain links, its only job is to group your sponsored folders together, and if you click on it, it doesn't get "selected", it just toggles open or closed. So, nobody using a browser will ever see this particular message: this is just to handle when an API user gets up to some shenanigans. (Well, unless some javascript is very broken.)

I think you are wanting this new message to appear in the `if folder.read_only` section below? Instead of the currently-proposed change, `{registrar.name} set this folder to read-only.`?

perma_web/api/views.py
```diff
@@ -388,35 +388,47 @@ def post(self, request, format=None):
...
-                error = "Perma.cc cannot presently make additional Perma Links on your behalf. Visit your subscription settings page for more information."
-                raise_invalid_capture_job(capture_job, error)
+                error = 'Your account needs attention—'
+                resolution = 'see your usage plan for details.'
+
+                if request.user.in_trial:
+                    error = 'You already created every link allotted to your trial plan.',
+                    resolution = 'Our Usage Plans page explains how you can get more.'
+                elif request.user.nonpaying:
+                    resolution = 'Contact Perma support at info@perma.cc with this error and your account information.'
```
rebeccacremona: I'm not sure about this one: if a non-paying user has reached their limit, should we prompt them to contact support, or just tell them that they've reached their limit and let us know if they need more? I don't think this is an error state.

rebeccacremona (changes requested):
Neato!

Looks good to me, except I think one message is in the wrong bit of code, and I think we might want to tweak the wording of one of them: see inline comments. @clare-stanton does that sound right to you?

Also, transferring one idea from Slack:
```
error = 'Your account needs attention—'
resolution = 'see your usage plan for details.'
```
sounds perfect for when a subscribed customer's account is on hold due to credit card problems, but something like
```
error = "You have reached your current usage limit." 
resolution = "For more information visit your Usage Plan page."
```
sounds perhaps more appropriate for the general case, when a person has just reached their limit? How about we throw that in here, a new branch inside `if not request.user.link_creation_allowed()` that checks `if request.user.cached_subscription_status == 'Hold'` and gives them `Your account needs attention` if yes, but gives everybody else that message about their usage limit?

Thanks for making this PR!

perma_web/api/views.py
ChefAndy: Ah— makes sense. So it looks like both non-paying accounts and my understanding of non-paying accounts are limited. Fix momentarily.

perma_web/api/views.py
ChefAndy: Got it. I'm always tempted to put Easter eggs in error messages like that, but I'll wait until I have a better understanding of the project.

rebeccacremona (approved):

PR #3023, opened 2022-01-31
Screenshots with playwright

holtchesley:


playwright/Dockerfile
```diff
@@ -0,0 +1,70 @@
+FROM ubuntu:focal
```
holtchesley: This is playwrights recommended docker setup. This image allows us to use Chromium, WebKit, and Firefox for tests. Below is a Debian based Dockerfiles that keeps us closer to what we're using for other images in the project, but only works with chromium currently.

playwright/Dockerfile.debian
```diff
@@ -0,0 +1,71 @@
...
+COPY playwright/Pipfile /playwright/Pipfile
+COPY playwright/Pipfile.lock /playwright/Pipfile.lock
+RUN pip3 install -U pip \
+    && pip install pipenv \
+    && pipenv --python 3.7 install --ignore-pipfile
+
+RUN pipenv run python -m playwright install
+
+# To support firefox and webkit
+# RUN pipenv run python -m playwright install-deps
```
holtchesley: This currently fails and so it's commented out. I may be able to do a bit more work to get FF and webkit to work without this line though.

holtchesley:
Some questions with this:
 - Do we go with Ubuntu or Debian for the base image?
 - Is having functional testing infra in a separate container still seeming like a viable path?
 - Are we still using Sauce Labs? Where is it being called from if we are?
 - Do we want to use a service like Save Labs/Browserstack/etc for testing IE,iPhone, multiple OSs... ?
 - The `dev.screenshot` task that we've got right now doesn't work for me(FireFox isn't installed in the container), is this worth fixing/documenting?

holtchesley:
> Do we go with Ubuntu or Debian for the base image?

The pro for Debian is that it's similar to our other images and that might simplify some maintenance. The upside of Ubuntu is that we get more reliable long term support with less fussing for FF+Webkit in the image. One of the advantages of playwright is that it's got compatible browsers bundled in, so I'm leaning towards Ubuntu.

holtchesley:
> Is having functional testing infra in a separate container still seeming like a viable path?

We lose access to some of the nice internal-to-django stuff like url reversal and such, but perhaps that's a reasonable tradeoff.

rebeccacremona:
> Are we still using Sauce Labs? Where is it being called from if we are?

I have no memories of ever having seen this run via Sauce Labs; I think its use predates me. It was never run via CI: people initiated runs from their machine. It would be kind of interesting to find out if this setup works, but I could easily imagine Sauce Labs has a put out a new API or whatever in the interim. I think we can very reasonably ignore/remove any Sauce-related stuff here, and make a separate issue for re-setting it up, referencing the original code as necessary, if we want to get it going again.

rebeccacremona:
> The dev.screenshot task that we've got right now

How about that! I didn't know this was around. Added in https://github.com/harvard-lil/perma/commit/cb61ebbc19c46652f3c2ab644f5d6efded7a9f85 in order to produce the screenshots that were used for the docs at that point in time. Fun!

I think at that time, devs ran Perma on the metal of their machine. I don't remember Firefox being installed in the VirtualBox VM we moved to..... And, we did lots of experimenting and got extremely close to adding Firefox to our Docker container, once we moved from the VM to Docker, but I don't think we ever ended up merging that in. 

So! I think this task hasn't worked in years. I wonder if, if working, the selectors it targets are still in the markup, if those screenshots are still in the docs... and if they are, if they presently up to date. This sounds like a very low priority side quest, maybe something fun for me to look at on a Friday.

If we were re-implementing today, might be able to use headless Chrome's [`--screenshot` flag](https://developers.google.com/web/updates/2017/04/headless-chrome#screenshots): Chrome is installed already for making captures.

holtchesley:
> > The dev.screenshot task that we've got right now
> 
> How about that! I didn't know this was around. Added in [cb61ebb](https://github.com/harvard-lil/perma/commit/cb61ebbc19c46652f3c2ab644f5d6efded7a9f85) in order to produce the screenshots that were used for the docs at that point in time. Fun!
> 
> So! I think this task hasn't worked in years. I wonder if, if working, the selectors it targets are still in the markup, if those screenshots are still in the docs... and if they are, if they presently up to date. This sounds like a very low priority side quest, maybe something fun for me to look at on a Friday.

The ~equivalent task in the playwright fab file works now (as far as I can tell), so the selectors are all there.

rebeccacremona:
> Do we want to use a service like Save Labs/Browserstack/etc for testing IE,iPhone, multiple OSs... ?

If it's affordable, would kind of love it for testing, say, the playback of one public Perma Link and one private one. It seems like overkill for the majority of our frontend code, but playback seems intricate enough that concrete assurance it works broadly might be appreciated. 

What are you thinking, worth pulling the thread, or likely to be more work (either resurrecting the Sauce Labs code, or switching to Browserstack, etc.) than it's worth?

playwright/Dockerfile
rebeccacremona: I wonder if rather than building our own, we want to use one of the pre-built images that Microsoft publishes (https://playwright.dev/docs/docker, https://hub.docker.com/_/microsoft-playwright, https://mcr.microsoft.com/v2/playwright/tags/list, https://playwright.dev/docs/ci#docker).

> These image includes all the dependencies needed to run browsers in a Docker container, and also include the browsers themselves.

I am encouraged to see that they are pushing for both `amd64` and `arm64`... though I am not willing to believe that all three browsers work until I see it. 

Maybe `v1.17.0-focal` would do it for us, at least as a base image? I think we're not attached to debian in this context, since this won't be deployed anywhere. @bensteinberg would you agree, ubuntu is fine for a testing container?

Or are you thinking we should build our own?

bensteinberg: I agree, Ubuntu is fine here.

rebeccacremona:
> Is having functional testing infra in a separate container still seeming like a viable path?
>
> We lose access to some of the nice internal-to-django stuff like url reversal and such, but perhaps that's a reasonable tradeoff.

How's it feeling to you? Are you having doubts? I am kind of delighted any time we can get stuff out of the main container, so that the main container gets closer and closer to what's actually running in prod. I am remembering, for instance, this one very fun bug where I pushed code that unknowingly relied on a package that was installed incidentally as a dependency for a test library. Worked locally, tests passed including in CI.... boom, busted in real life lol. 

Especially if we are able to take advantage of the pre-built playwright images, complete with installed browsers (which, as said, I'll believe when I see lol), this still sounds like a good plan to me.

Re: not having access to Django niceties like URL reversal, we could potentially set up something clever so that the main Django-running container passes the info the functional tests need. For instance: we could give the main perma container the ability to run docker commands (https://github.com/harvard-lil/perma-capture/blob/new-arch-with-vue/docker-compose.yml#L43), and then have pytest fire off each functional tests, sending along whatever info was needed as args to the invocation, or even using a volume to share files, or whatever. Might be more baroque than we need, but might be kind of slick.

playwright/pyproject.toml
```diff
@@ -0,0 +1,5 @@
+# pyproject.toml
+[tool.pytest.ini_options]
+filterwarnings = [
+    'ignore::DeprecationWarning',
```
holtchesley, line 4: pytest-playwright uses the https://pyee.readthedocs.io/en/latest/index.html library for managing some of it's async and async-adjacent aspects. We should revisit this periodically to see if we can get rid of this after an update.

services/docker/webrecorder/nginx/nginx.conf
```diff
@@ -53,7 +53,8 @@ http {
     # Content Server
     server {
         listen 81;
-
+        listen 8092;
```
holtchesley, line 56: A fun bit about the docker port mapping is that it's between containers and the host, and not containers themselves. This lets the playwright container use the same `perma-archives.test:8092` urls that local dev does. 

I don't _think_ this steps on anything else.

holtchesley:
A quick summary of changes:

- Got rid of the Debian Dockerfile
- Added a functional test to verify the homepage loads, and one that captures example.com and verifies that it can playback regularly.
- Updated the Nginx container to make the functional test work. 
- Functional tests can now be run with `docker-compose exec playwright pipenv run pytest` with som additional params outlined in https://playwright.dev/python/docs/test-runners

holtchesley:
Since the last update:
 - Created #3026 for a follow up on cross browser/os testing with an external service. 
 - CircleCI runs the tests.
 - Perhaps more improved fixtures for any future testing we want to do. 

I think this is probably ready to send out unless there's an issue with the nginx config change I made, or anything else in the PR.

playwright/Pipfile
```diff
@@ -0,0 +1,20 @@
...
+pyquery = "*"
+pytest = "*" 
+playwright = ">=1.18"
+pytest-playwright = "*"
+
+[requires]
+python_version = "3.8"
+
+[pipenv]
+allow_prereleases = true
```
rebeccacremona: Do you happen to know if we need pre-releases here? Perma's main env had to add that to solve a particular problem, but it sometimes bites us, switching us to alpha packages that are not ready for primetime. Might be nice to -not- include, unless we do in fact need it.

playwright/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM mcr.microsoft.com/playwright:v1.18.1-focal-amd64
+
+ARG DEBIAN_FRONTEND=noninteractive
+ARG TZ=America/Los_Angeles
```
rebeccacremona: Do you know if we need these two args? I'd imagine that since we are using `-y` with `apt-get install` then we don't need `DEBIAN_FRONTEND=noninteractive`; I wonder about setting the timezone to LA.

playwright/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM mcr.microsoft.com/playwright:v1.18.1-focal-amd64
```
rebeccacremona: I wonder, does this still work on your machine if, instead of ending with `-amd64`, you do `-${BUILDARCH}`? The work to alter the other Dockerfiles and etc. to make this cross-platform hasn't landed yet, but, if that pull the correct thing on your machine, I'd be super interested. Either way, no pressure to tweak: will run this on an M1 and play after merged.

playwright/fabfile.py
```diff
@@ -0,0 +1,58 @@
+import asyncio
+from playwright.async_api import async_playwright
+
+
+from fabric.decorators import task
+
+@task
+def screenshots(base_url='http://perma.test:8000', output_path='/playwright/screenshots', browser='chrome'):
+    async def screenshot(page, upper_left_selector, lower_right_selector, file_name, upper_left_offset=(0,0), lower_right_offset=(0,0)):
```
rebeccacremona, line 9: 🤘 https://hlslil.slack.com/archives/C5U7T2B35/p1643401570966799

playwright/tests/test_functional.py
```diff
@@ -0,0 +1,32 @@
...
+    password.type(user.password)
+    page.locator("button.btn.login").click()
+    return page
+
+def test_example_dot_com(logged_in_user):
+    url_field = logged_in_user.locator('#rawUrl')
+    url_field.focus()
+    url_field.type("https://example.com/")
+    logged_in_user.locator('#addlink').click()
+    logged_in_user.wait_for_url(re.compile('/[A-Za-z0-9]{4}-[A-Za-z0-9]{4}$'), timeout=two_minutes)
```
rebeccacremona, line 29: Using the GUID regex there is fun.

services/docker/webrecorder/nginx/nginx.conf
rebeccacremona: Neat.

So, since it seems to be working locally and in Travis, I think we're good: this is not the nginx conf used in production, so a tweak won't affect any other environments.

docker-compose.yml
```diff
@@ -90,6 +90,26 @@ services:
     networks:
       - default
 
+    #
+  # Perma Functional Tests
+  #
```
rebeccacremona, line 95: Stray keystroke: mind out-denting that `#`?

playwright/Dockerfile
rebeccacremona: Oh, and we should ask @bensteinberg to mirror the relevant images up to our registry and switch this line to point there to avoid Docker Hub's limits on pulling the image.

rebeccacremona (commented):
Jazzed about this. I haven't seen playwright before; everything seems so clean and tidy!

I had a couple of minuscule questions about the Dockerfile and Pipfile, but LGTM!

playwright/Dockerfile
holtchesley: Removed!

playwright/Pipfile
holtchesley: Confirmed that we do not. I was cleaving close to the perm_web setup, and we can remove this.

playwright/Dockerfile
bensteinberg: I don't think it's necessary here, since we're pulling from Microsoft, not Docker Hub.

playwright/Dockerfile
holtchesley: > I believe you can just use `ARG BUILDARCH` here, which will give you access to a correct `${BUILDARCH} later in the file.

This fails for me locally unless I also supply an appropriate `--build-arg` .

holtchesley:
> I believe you can just use `ARG BUILDARCH` here, which will give you access to a correct `${BUILDARCH} later in the file.

This fails for me locally unless I also supply an appropriate `--build-arg` (which, for the record, is something I'm happy to do if it makes it easier for other build stuff to work).

rebeccacremona:
> > I believe you can just use `ARG BUILDARCH` here, which will give you access to a correct `${BUILDARCH} later in the file.
> 
> This fails for me locally unless I also supply an appropriate `--build-arg` (which, for the record, is something I'm happy to do if it makes it easier for other build stuff to work).

Huh, funny, I see that https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope says these args are ":only available when using the [BuildKit](https://docs.docker.com/engine/reference/builder/#buildkit) backend." I wonder if you have a different backend.

🤷 tbd

holtchesley:
I tested the bare `ARG BUILDARCH` in the latest pull, and it looks like the current CircleCI setup also doesn't recognize it.

rebeccacremona:
> I tested the bare `ARG BUILDARCH` in the latest pull, and it looks like the current CircleCI setup also doesn't recognize it.

How about that. That must be what's up with https://github.com/harvard-lil/h2o/pull/1494 too.

PR #3025, opened 2022-02-01

bensteinberg:
https://docs.djangoproject.com/en/dev/releases/2.2.27/

The only major version update here is to `pyopenssl`: 

https://www.pyopenssl.org/en/stable/changelog.html#id1

### DIFF
--- a/perma_web/Pipfile.lock
+++ b/perma_web/Pipfile.lock
@@ -63,19 +63,19 @@
         },
         "boto3": {
             "hashes": [
-                "sha256:42dd9fcb9e033ab19c9dfaeaba745ef9d2db6efe4e9f1e1f547b3e3e0b1f4a82",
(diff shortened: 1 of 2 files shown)


perma_web/Pipfile.lock
```diff
@@ -739,16 +739,14 @@
...
             ],
             "index": "pypi",
-            "version": "==4.1.0"
+            "version": "==4.1.2"
         },
         "requests": {
-            "extras": [
-                "security"
-            ],
+            "extras": [],
```
rebeccacremona, line 749: Oh look at that, the "security" extras is finally gone, as long threatened.

rebeccacremona: I should remind myself about this whole thing, later, and probably remove the "extras" thing from Pipfile. I'll make an issue for it.

Issue #2056, opened 2017-06-07
PDF in Perma Link displaying only 1st page on mobile


bjohnson747:
PDF at http://www.nyulawreview.org/sites/default/files/pdf/NYULawReview-77-6-Daines.pdf was made into a Perma Link: https://perma.cc/M3MM-P95L

On desktop, Perma Link PDF displays in full, no problem. On mobile, Perma Link only lets 1st page be viewed (issue re-created in both safari and chrome on mobile).

rebeccacremona:
I think we should leave this for Apple to fix: it's a long-standing issue with mobile Safari with no satisfactory workarounds. For some discussion, see the following thread on StackOverflow and similar: https://stackoverflow.com/questions/15480804/problems-displaying-pdf-in-iframe-on-mobile-safari

bjohnson747:
Jack had the idea for a button that, on mobile, let's the user "view PDF", which would take them directly to the doc/out of the iFrame.

rebeccacremona:
Reports are that on some people's iOS devices, even switching to Chrome or Firefox doesn't help. So, we'll try something like this:

![download_button](https://user-images.githubusercontent.com/11020492/27405345-3d26bc1a-569f-11e7-937a-d06d84e7988a.png)

bjohnson747:
That looks great. (Should it be changed to - when mobile is detected - saying "Perma.cc can't display this file *on mobile* but [...]"

rebeccacremona:
For the record: this is still a thing. Just checked https://discussions.apple.com/thread/250075244 with my shiny new iPhone.

Sean-In-The-Library:
Still a problem in Oct, 2023

rebeccacremona:
@Sean-In-The-Library alas, yes, unfortunately we think this is unsolvable at this time https://github.com/harvard-lil/perma/issues/3358

PR #3036, opened 2022-02-09
Client-side playback: offer download via an interstitial when file won't display in browser.

rebeccacremona:
Perma shows visitors an interstitial when they visit a Perma Link whose capture cannot be played back live in the browser due to its file format. For example, this is how a capture of the DOCX at https://github.com/harvard-lil/h2o/raw/develop/web/test/files/export/export-casebook-no-annotations.docx looks:

image

With normal client-side playback, the file is served immediately on page load:

image

Similarly, because of https://github.com/harvard-lil/perma/issues/2056, and because PDF-reading experience on mobile is pretty lousy in general, even without the interference of our UI, we also show mobile users an interstitial in lieu of playing back PDFs.

This PR makes those interstitials work with client-side playback locally: a little more complicated, because we don't have a predictable, server-side URL for arbitrary resources inside the warc that we can easily link to anymore. When a user clicks the download button, we load the replay in a hidden iframe. In the case of DOCX and other file formats, that is enough to trigger my Firefox to download the file. In the case of PDF, in simulated mobile situations, a further step is needed: reaching into the playback iframe and finding its `src`, then causing an `<a>` pointing there to be clicked, thus evading the browser's PDF plugin.

I'm pretty jazzed it seems to work, though the PDF workaround is likely fragile due to the complexity of reaching all the way in to the playback iframe.

I'll add playwright-mediated functional tests... maybe in a separate PR (I want to set up an apparatus for capturing local assets during functional tests, copied from the perma-capture repo, first, and am not sure whether that belongs here...)

This will definitely want some cross-browser and cross-platform exercise. I'm particularly interested to see if it works on mobile Safari.

Draft while deciding whether to add tests here or take that on separately.


rebeccacremona:
Also just noting, there's something up with the webpack builds.... I got this relatively clean diff by inserting my function into the file all at once, and hitting save while the server is running. Anything else caused basically every file in the bundle to be updated with all sorts of stuff..... I'm not going to try and figure out exactly what's up (it's probably the common chunks plugin), given the age of this pipeline: let's just see if it can hold on until we are able to port to Vue etc. and set up a pipeline like our other projects have.

rebeccacremona:
Gah, the tests failed because of the webpack bundle. Well, we'll see.

rebeccacremona:
Wrapped into https://github.com/harvard-lil/perma/pull/3068

PR #3042, opened 2022-02-25
Minimal changes for upgrade to bullseye

bensteinberg:
This is work in progress; two tests are failing locally.


perma_web/Dockerfile
```diff
@@ -69,14 +68,14 @@ COPY perma_web/Pipfile /perma/perma_web
...
     && pip install pipenv \
-    && pipenv --python 3.7 install --ignore-pipfile --dev \
+    && pipenv --python 3.9 install --ignore-pipfile --dev \
     && rm Pipfile.lock
 
 # Install Chrome
 # technique from https://github.com/SeleniumHQ/docker-selenium/blob/master/NodeChrome/Dockerfile.txt
 ARG chrome-layer-cache-buster
 RUN apt-get --allow-releaseinfo-change update \
-    && apt-get install -y unzip libappindicator3-1 libasound2 libatk-bridge2.0-0 libatk1.0-0 libatspi2.0-0 libgbm1 libgtk-3-0 libnspr4 libnss3 libxcomposite1 libxcursor1 libxi6 libxrandr2 libxss1 libxtst6 xdg-utils \
```
bensteinberg, line 79: `libappindicator3-1` is not in bullseye and does not appear in the list of deps from `docker-compose exec web dpkg -s google-chrome-stable`. I haven't confirmed that the packages installed here match those deps one-to-one, but the installer would bark if we were missing anything.

perma_web/Pipfile
```diff
@@ -21,7 +21,7 @@ uwsgitop = "*"                              # uWSGI monitoring
 
 
 # databases
-psycopg2 = ">2.7, <2.9"
+psycopg2 = "==2.8.6"
```
bensteinberg, line 24: This change is not strictly necessary; see https://github.com/psycopg/psycopg2/issues/1293

perma_web/api/tests/test_link_resource.py
```diff
@@ -684,7 +684,7 @@ def test_media_capture_in_iframes(self):
             # test_media_b.html
             ("test.mp4", "video/mp4"), ("test2.mp4", "video/mp4"),
             # test_media_c.html
-            ("test.swf", "application/x-shockwave-flash"), ("test2.swf", "application/x-shockwave-flash"), ("test3.swf", "application/x-shockwave-flash"),
+            ("test.swf", "application/vnd.adobe.flash.movie"), ("test2.swf", "application/vnd.adobe.flash.movie"), ("test3.swf", "application/vnd.adobe.flash.movie"),
```
bensteinberg, line 687: The MIME type for .swf files was changed in bullseye; see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=985039 for some discussion.

PR #3051, opened 2022-03-23
Limit maximum size of IA uploads

bensteinberg:
The point of this is to prevent repeated attempts to upload large WARCs that time out and jam the upload queue.

@rebeccacremona I think this does what we were talking about; I'm not sure if it would be better to leave the setting unset by default, and set it only in production.

This does not address whatever manual or other process we need to upload larger WARCs to IA, nor, of course, does it address the problem of creating oversized WARCs ;)


rebeccacremona:
Looks great! Yes, how we want to handle the larger warcs, post hoc, TBD..... Perhaps designed into the current project I'm scoping today......

PR #3056, opened 2022-04-04
Fix honeypot

rebeccacremona:
1) The logic needed to be run earlier in two cases (courts and firms)
2) The contact form's field wasn't consistently tasty enough

~If the bundles test fails, it's because of whitespace... I manually edited instead of switching to an Intel computer: will fixup if needed. Done~


bensteinberg (commented):
:+1:

PR #3060, opened 2022-04-06
Celery upgrade

jcushman:
First cut at updating celery from 4.x to 5.2.6.

* First commit is a bare `pipenv update`, which Becky recommended to work with pipenv.
* Second commit updates celery, drops the `future` dependency that had been there for celery 4.x, and drops the `[redis]` extra. Redis is a top level requirement anyway, and the extra was causing a false version conflict with fakeredis.
* Third and fourth commits aren't directly necessary, but borrow some celery techniques from CAP as I was checking how Perma uses celery:
  * Third commit removes Perma's custom `RUN_TASKS_ASYNC` and `run_task` helpers, and replaces with built-in `CELERY_TASK_ALWAYS_EAGER` setting (which is the same but inverse from our custom setting) and builtin `task.delay()`, because it's simpler to use builtins.
    * Also adds `CELERY_TASK_EAGER_PROPAGATES` in dev, which makes stack traces better when you hit an error while eager=True.
    * Also switches the default in settings_dev.py to eager=True, which I find more convenient as a default. Feel free to disagree though.
  * Fourth commit adds watchdog to auto-restart celery when it's running in the background, so you don't have to remember to restart it depending on your eager setting.

### DIFF
--- a/developer.md
+++ b/developer.md
@@ -342,25 +342,13 @@ kicked off by `d fab dev.test_sauce`.
 
 Celery does two things in Perma.cc: it runs the capture tasks and it runs scheduled jobs (to gather things nightly like statistics, just like cron might).
 
-In development, it's sometimes easier to run everything synchronously, without the additional layer of complexity a Celery worker adds. To run synchronously, set `RUN_TASKS_ASYNC = False` in settings.py. `RUN_TASKS_ASYNC` must be true if you are specifically testing or setting up a new a Celery-Django interaction, and must be true when working with LinkBatches (otherwise subtle bugs may not surface).
+In development, it's sometimes easier to run everything synchronously, without the additional layer of complexity a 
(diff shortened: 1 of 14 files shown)


rebeccacremona (approved):
LGTM!

I pulled this down and added a commit to increment the Docker image no. Then:

- tested a capture with (the new) default settings
- toggled to `CELERY_TASK_ALWAYS_EAGER = False` and watched a link batch work
- scheduled some tasks to run via celerybeat and watched them run on schedule
- wired up my instance to talk to Internet Archive and manually uploaded then deleted an item from the test collection.
- threw in a `1/0` at the top of `run_next_capture`, and observed how the error message was presented with `CELERY_TASK_EAGER_PROPAGATES = True` (looks normal to me, I think I've had this in my personal `settings.py` for years and forgotten about it) and `CELERY_TASK_EAGER_PROPAGATES = False` (error email to admins printed in console; suggests things will work as desired in prod)

I'm going to push the commit to increment the docker image no. Then will coordinate about what might need to change on stage, where celery is daemonized and handled by systemd, before merging.

PR #3061, opened 2022-04-07
Pipenv to pip tools

jcushman:
Just for fun, here's what it would look like to switch from pipenv to pip-tools. Not sure if we actually want to mess with this right now, but I realized I could try it out in like an hour, which is less time than it took to upgrade celery.

The process to generate requirements.in from Pipfile for each of `web` and `playwright` is unfortunately manual, so it'll have to be done again if requirements are changed by other PRs:

1. `pip freeze > web_pipenv.txt` to record what pipenv installs.
2. `pip install pip-tools`
3. Manually copy the requirements from Pipfile to requirements.in and update the format. Keep any max-version constraints for now.
4. `pip-compile --allow-unsafe --generate-hashes` to create requirements.txt, or use `fab pip-compile` which calls that.
5. `pip-sync` to install requirements.txt
6. `pip freeze > web_piptools.txt` to record what pip-tools installs.
7. Diff web_pipenv.txt and web_piptools.txt to make sure they're not too different.
8. Remove max-version constraints from requirements.in (I didn't actually do this)

I based this one on the celery PR since it had some package updates; could rebase if we wanted to keep it.

Everything seems to work after this. Note one potential side effect is that dev requirements are installed everywhere, where pipenv has them in a separate `[dev-packages]` section.


jcushman:
Force-pushed to make a couple of package versions match in the playwright container to get the tests to pass.

Note that I did this in a way that didn't preserve the versions of anything that wasn't pinned in `Pipfile`, so there's some small variations in package versions which I figured would be fine since Perma's pretty up to date anyway. If we wanted an exact match, we could modify my procedure above to `pip freeze > requirements.txt` at the start, and then `pip-compile` would respect all the existing pinned versions.

PR #3068, opened 2022-04-13
Replay from subdomain

rebeccacremona:
## The status quo

Right now on `develop`, behind a feature flag, we give admins and specially-authorized users the ability to preview client-side playback. By using the buttons in the playback "tray" or by manipulating the Perma Link URL's querystring, authorized users can request a playback that, instead of loading an `iframe` pointed at a domain hosting Perma's Webrecorder instances, instead installs a Javascript service worker at `perma.cc/replay` and inserts a custom web element on the page (`<replay-web-page>`) that work together to play back a Perma Link with only one a single network request to our servers: a request to download the warc.

## This PR

This PR makes two very small changes...
- updates to use the most recent release of the replayweb.page
- removes the feature flag around client-side playback 

...and replaces that initial experiment with a more secure setup:
- When a client-side playback is requested, we load an `iframe` pointed at a page on a subdomain (e.g. replay.perma.test)
- That page installs the service worker at `replay.perma.test/static/vendors/replay-web-page` and loads the custom `<replay-web-page>` web element responsible for the playback. 
- Its request to download the warc goes to a route served at `perma.test`, not to a route served by the subdomain. we may want to move this route to the subdomain, but I put it here for now in case we can get the service worker to tell fetch to send cookies instead of making the request [anonymously](https://github.com/harvard-lil/perma/projects/10#card-76651126). If we can, and if it doesn't mess up our [security model](https://github.com/harvard-lil/perma/projects/10#card-75836075), that would make the playback of private perma links very easy. But, if we do something else (say, issue a one-time download token/nonce or something), we may want to move this route to the subdomain.

## SSL required

This required a significant change to our development setup: we now use SSL for Perma and the public/browser-facing part of Webrecorder locally and in CI. (Browsers will only install service worker JS for named hosts, like `perma.test` and `replay.perma.test`, if said JS is served via HTTPS. They will only install service worker JS served over HTTP for `localhost`, verbatim... and subdomains of localhost, like `replay.localhost` aren't universally supported.)

After this is merged, developers need to:
- update their `/etc/hosts` as per `install.md`
- install [mkcert](https://github.com/FiloSottile/mkcert)
- and run `bash make_cert.sh`

The application will then be available at https://perma.test:8000. (I naively would have expected http://perma.test:8000 to redirect to https://perma.test:8000 because of `SECURE_SSL_REDIRECT`, but I guess that's not how runserver_plus works.)

I updated most of the tests accordingly by passing `secure=True` to each request made to the test client. A few were proving tough to get working that way; I overrode `SECURE_SSL_REDIRECT` instead of worrying about it too much. The Selenium functional tests also all run with `SECURE_SSL_REDIRECT=False`, because the LiveServerTestCase server doesn't support SSL.

## Settings

`OFFER_CLIENT_SIDE_PLAYBACK` was removed.

`PLAYBACK_SUBDOMAIN` was added with a default of `replay`.

`PLAYBACK_UI_JS_URL` and `SERVICE_WORKER_URL` have been replaced with `REPLAYWEBPAGE_VERSION` and `REPLAYWEBPAGE_SOURCE_URL`: if you are overriding anything in `settings.py`, you'll need to adjust accordingly. 

## See it in action

The tests still all use our server-side, Webrecorder playback, and Webrecorder playback is still the default for the running application. To see the client-side playback, log in as an admin (or as one of the `USERS_WITH_CLIENT_SIDE_PLAYBACK`), visit a Perma Link, open the tray, and click the "replay" or "compare" buttons. 

To see a screenshot played back via the client-side tech, manually append `&type=image` to the querystring.

Try capturing a page that can't be viewed in-browser, like a [Word document](https://github.com/harvard-lil/h2o/blob/develop/web/test/files/export/export-casebook-no-annotations.docx?raw=true) to see our interstitial. Try capturing a PDF and play it back in Chrome to see that the `noSandbox` option works. Try playing back a PDF on a mobile device, if possible in Safari, and see if that second, even fancier, interstitial works. (And make sure you have `?client-side=replay`... easy enough to get confused :-))

You will not be able to trigger a playback by visiting https://replay.perma.xxx directly: the page should only attempt a replay if it is loaded in an iframe and otherwise renders some placeholder text. (Right now, 'Welcome', but I'm sure we can come up with something more informative and appropriate).

## Next up

I did not re-implement the "explore contents" or "list pages" buttons yet, and private playbacks do not work.


bensteinberg:
`--retry-all-errors` was added in curl 7.71.0; Ubuntu 20.04 has curl 7.68.0.

PR #3072, opened 2022-04-19
Use Github Actions

jcushman:
This moves Perma from CircleCI to the same Github Actions setup as CAP and H2O. Some benefits along the way:

- Docker images are stored in our registry on pull to develop, with updated tags that include a hash of all files contributing to their contents. This means you can do `docker pull` once the images are built to avoid having to rebuild locally, and you don't have to (shouldn't) update image versions locally.
- Webpack bundles are automatically committed if they're changed by the tests, so it doesn't matter whether you commit them locally or not.

Like H2O this will need some updated secrets set before pulling.


rebeccacremona:
wow!!!!!!!

rebeccacremona:
> you don't have to (shouldn't) update image versions locally

I'd love to chat about what your local flow looks like, when you are working on updating images, with this setup. My routine is always:
1) update image tags in `docker-compose.yml`
2) make my alterations to `requirements.in` or `package.json` or `Dockerfile` or whatever
3) run `docker-compose up -d --build`
4) iterate on 2 and 3 as necessary

With this setup, I can continue that flow, except the new tag in 1) would be temporary, and I would revert it before committing.

But I'm guessing you have a different pattern?

jcushman:
(note tests are expected to fail here)

jcushman:
> With this setup, I can continue that flow, except the new tag in 1) would be temporary, and I would revert it before committing.

This sounds right to me, unless there's some docker-compose flag that does the equivalent of changing to a temp flag before building.

I think the one difference with my pattern is (maybe because I think of the running docker containers as a remote machine I'm logging into) I tend to always be in `docker compose web bash`, and I'd test commands there before adding them to anything. So I'd tend to do `(guest)# apt-get install new thing; test new thing` or `(guest)# fab pip-compile; pip install -r requirements.txt; test new packages`, before editing Dockerfile or rebuilding. And at that point I might be confident enough that I'd just ship it for CI to take a try on instead of rebuilding locally, though it's a gamble whether that's faster.

bensteinberg:
I just made a PR to your branch to suppress auto-deploy for the moment.

PR #3075, opened 2022-04-27
Update Chromium

bensteinberg:
In the absence of a release post or other announcement, here's a log of source changes:

https://chromium.googlesource.com/chromium/src/+log/100.0.4896.127..101.0.4951.41?pretty=fuller&n=10000

This PR also tweaks the update script to prevent it from writing a spurious change to `settings_common.py` if it can't get the new Chromium package.


bensteinberg:
https://www.debian.org/security/2022/dsa-5125 -- and the arm64 Debian package is now present.

PR #3081, opened 2022-05-12
Client-side playbacks use S3 pre-signed URLs for warcs

rebeccacremona:
Today on prod, if playing back a warc using the client-side JS, Perma serves the warc from a [route of its own](https://github.com/harvard-lil/perma/blob/3d087952b5a9ed66d7f0f71c65075072167121d6/perma_web/perma/views/common.py#L328). That works for public Perma Links, but not for private ones: by design, the playback JS does not have access to the credentials necessary to authenticate with Perma.

This PR adopts a different strategy: when you visit https://perma.test:8000/AAAA-AAAA, Perma either denies you access...
image
... or loads the playback subdomain in an iframe whose `name` is set to an S3 pre-signed URL that can be used to download the warc.
image

For that to work, there are two important things on the S3 end: 
- the bucket needs to be configured to serve with the right CORS headers (see https://replayweb.page/docs/embedding#cors-restrictions, see https://github.com/harvard-lil/perma/pull/3081/commits/970ca402e3e1331b26338c1279b4b1644d8cf8b7#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R93); we'll need to make sure this is set up for stage and prod.
- each warc needs to be served with the correct content-type and encoding; the app will take care of this: no config changes necessary for stage or prod.

The content-type and -encoding is surprisingly finicky: it turns out, these are immutable metadata fields set during S3 object creation... and we, via our django-storages integration, [didn't quite set them correctly](https://github.com/jschneier/django-storages/issues/917): we need `content-type: application/x-gzip`, not `content-type: application/octet-stream, content-encoding: gzip`. This PR has us upload warcs with the [right metadata going forward](https://github.com/harvard-lil/perma/pull/3081/commits/eef07bea873ece2c7f58c2bfdb4f89cb9d001c78), and uses a fantastic feature of pre-signed URLs to [override the headers for already uploaded warcs](https://github.com/harvard-lil/perma/pull/3081/commits/99e5649c6d6221c6d9e4725d97c92df2d6f456b9) (with many thanks to jcushman for that trick!!)

To get this working locally, we're making a big switch: dev instances now store objects in a locally-run (containerized) instance of minio instead of on the file system. Developers will need to **add perma.minio.test to their /etc/hosts** file and **re-run make_cert.sh** to get this up and running on their machines.

That change uncovered a couple tiny long-standing bugs: a couple with our test suite (https://github.com/harvard-lil/perma/pull/3081/commits/0fa204f6b870a4f6c10ef213640c9be241f82439, https://github.com/harvard-lil/perma/pull/3081/commits/1da8e4091bee647de09ca57aba1d88d38be7ec39) and one real one: when uploading registrar logos, admin users could [accidentally overwrite existing logo files](https://github.com/harvard-lil/perma/pull/3081/commits/327ebd29a8bf6fa4e0d2d0adce5357590d947983). But, I checked and there's no evidence that actually ever happened.

Finally, I make this all a little easier to share and test: right now, if you visit https://perma.test:8000/YQQ5-DZWC?client-side=replay while logged out, or while logged in as a non-admin user, you will not get the client-side playback, but rather, the normal server-side playback mediated by Webrecorder. I removed that logic: now, anybody can add `?client-side=replay` to the URL and get a client-side playback. Look for the replayweb.page navigation to easily prove you are really seeing the client-side playback:
image

I also threw in some code for [cleaning up between replays](https://github.com/harvard-lil/perma/pull/3081/commits/6d8a4806ef2d2bebe46c115e77a3627b559d4efe), for dev convenience while this is still under active development. It's unlikely we'll want to keep it long-term, but we probably want [something similar](https://github.com/harvard-lil/perma/projects/10#card-80836972). And, I threw in a work-around for a [Chrome-specific problem](https://github.com/harvard-lil/perma/projects/10#card-82115949).

So, it's a bit of a grab-bag PR, for which I apologize.... but, it gets us much closer to being able to ship this to users.


bensteinberg:
I think we need to add `perma.minio.test` to the instructions about `/etc/hosts` on line 34 of `install.md`.

bensteinberg (approved):
👍

PR #3087, opened 2022-05-23
Add Brewfile and installation instructions

lizadaly:
Totally optional and no worries if this isn't desired, I don't mind closing!

`brew` supports providing "bundle" files, similar to `requirements.txt`, which can be installed via `brew bundle`. In this project it's not really any more convenient than installing the single `mkcert` dependency itself (though `mkcert` suggests installing `nss` too) but if you like `brew bundle` it could be standardized across projects that may have more brew dependencies.

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,0 +1,2 @@
+brew "mkcert"
+brew "nss"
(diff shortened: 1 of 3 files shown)


rebeccacremona:
Oh, I didn't know about that, and I love it! Thx!

Issue #3088, opened 2022-05-25
Port active selenium tests to playwright and remove old suite


lizadaly:
The previous selenium-based tests were in `../perma_web/functional_tests/` but were not being called and weren't in use.

We have a new suite written with Playwright that will be the source for future integration/functional tests.

Port any seemingly-useful tests from the old suite and delete the old code and its dependencies.

PR #3089, opened 2022-05-25
[#3088] Move Selenium-based test to Playwright; remove old functional suite

lizadaly:
A quick port of one of the older functional tests to playwright, plus some amount of rewriting of the playwright tests to be less "hello world."

As @rebeccacremona said, most of the old suite was really just scaffolding or helper files, and a lot of the "real" tests were commented out. I actually decided to just wholesale delete them (I'm a fan of deleting code that's unused and just using git history as a reference) and I think it'd be better to make a new set of functional tests from scratch. 

Draft while I ensure that the CI runs.


lizadaly:
This was too aggressive; let me excise just the Sauce portions of the previous functional tests.

lizadaly:
After sleeping on it, I think the drop in coverage is expected—the playwright tests run in a separate `pytest` process from the coverage report, so any reductions in coverage from removing the `StaticLiveServerTestCase` aren't made up by coverage added in the playwright tests.

There probably won't be a useful way to measure coverage coming from these tests, but that's probably OK. 

(Apologies for the messy PR, I un-drafted this prematurely.)

PR #3097, opened 2022-05-27
Update Chromium

bensteinberg:
This is a major version upgrade of Chromium, from `101.0.4951.64-1~deb11u1` to `102.0.5005.61-1~deb11u1`:

https://www.debian.org/security/2022/dsa-5148


lizadaly (approved):
🙌

Issue #3093, opened 2022-05-26
Improve tooling/workflow for Chromium upgrades


bensteinberg:
Because we're now building images for arm64 as well as amd64, Chromium upgrades have two problems:

The first, relatively minor problem is that the arm64 build lags behind the amd64 build by hours or even days, so that it is impossible to upgrade until both architectures are built.

The second problem is that in at least some cases, Debian withdraws the older, presumably vulnerable version of Chromium, so that our multi-arch image can't be built in CI, and developers with M1 machines can't build -- during the period when only amd64 has been built.

bensteinberg:
The current theory is to set up a package repo to serve the five relevant files in both the previous version and the new version for both architectures. This will require changes to the Dockerfile and to deployment config. The exact mechanism for replicating packages to the repo remains to be determined.

bensteinberg:
(Keeping the packages in this repo would mean twenty files totalling around 300MB.)

bensteinberg:
https://github.com/harvard-lil/perma/pull/3099 makes Perma use our own repo for Chromium; I will leave this issue open until we've been through one complete upgrade cycle, and possibly until some or all of the upgrade workflow is automated.

bensteinberg:
The upgrade workflow is not automated, but this appears to be working fine as is.

PR #3099, opened 2022-05-27
Use LIL repo for Chromium

bensteinberg:
This addresses #3093 but is not a complete fix; the setup of the repo is somewhat provisional and does not yet automate the addition of new Chromium versions when they appear.

The infrastructure components of this are a matching change to the salt config for spinning up Perma workers, and the salt config for setting up the repo.

The proof will be in the pudding: when the present version of Chromium is no longer available from Debian, can we build a new dev environment (and spin up a new stage or prod worker)?


bensteinberg:
I think this is ready to use; I am still working on internal workflow documentation for the upgrade process.

rebeccacremona (approved):
LGTM! 🎉

PR #3102, opened 2022-06-02
Minor web performance and a11y tweaks

matteocargnelutti:
Very minor tweaks:
- **Side-wide:** Added `font-display: swap` on custom fonts to prevent text from being invisible while fonts load 
- **Homepage:** Lazy loading tweaks


matteocargnelutti:
Ended up being very small in scope 😄. 


I think there's an opportunity to make Perma's front-end much lighter and less prone to side-effects by shaving off the amount of CSS / JS it ships with every page: I'll have a stab at this as part of my efforts to migrate from Webpack to Vite.

Screen Shot 2022-06-03 at 12 38 48 PM

Screen Shot 2022-06-03 at 12 39 03 PM

PR #3103, opened 2022-06-13
Preparation for a chromium upgrade

bensteinberg:
This upgrades `requests` in order to get past a pip resolution problem, discovered when trying to update chromium (and exercise our new chromium-upgrade workflow). It also explicitly installs each chromium package with the specified version, necessary when the repo contains more than one version.


bensteinberg:
Now that this passes, and I've seen it correctly hitting our package repo, I'm going to bump the Chromium version and take it out of draft mode.

bensteinberg:
I wonder if this did not automatically get a reviewer because I created it in draft mode, or because I have the team set up wrong, or because I don't understand the mechanism. Maybe I need to assign the team as a reviewer?

matteocargnelutti (approved):
Looks good to me and everything still works as expected locally after the upgrade(s) 🚀

PR #3106, opened 2022-06-22
CORS headers for api.domain.ext

matteocargnelutti:
- Allows for the API to be used without proxy in a browser context.
- Adds `Access-Control-Allow-Origin: *` to the headers of the response of every API route. 
- Only applies to requests coming out of `api.domain.ext`.


bensteinberg:
LGTM

PR #3109, opened 2022-06-23
Perma API: Reworked CORS management

matteocargnelutti:
### Problems identified with previous implementation
- We [can't / shouldn't have both `Access-Control-Allow-Origin: *` and use the `Authorization` header](https://javascript.info/fetch-crossorigin#summary) in that context.
- Setting `Access-Control-Allow-Origin` to the origin of the incoming request still results in failed preflight requests, as the `Authorization` header is not passed during preflight _(Django therefore returns HTTP 401)_. 

### Proposed fix
- Set `Access-Control-Allow-Origin` to the origin of the incoming request when possible
- _"Force"_ HTTP 204 on `OPTIONS` requests that resulted in a 401, so CORS preflight passes.

### Not included / commented out
- I don't believe we want passive cross-origin credentials passing _(i.e cookies, stored HTTP auth)_. I left `Access-Control-Allow-Credentials` commented-off on purpose. 

Workaround tested on Chrome latest, Firefox latest.

### DIFF
--- a/perma_web/api/middleware.py
+++ b/perma_web/api/middleware.py
@@ -2,14 +2,27 @@ from django.middleware.common import CommonMiddleware as DjangoCommonMiddleware
 
 class CORSMiddleware(DjangoCommonMiddleware):
     """
-       Set `Access-Control-Allow-Origin: *` for API responses, including redirects.
-       Only applies if requests come from api.domain.ext.
+       Add CORS headers to requests made to `api.domain.ext`, so the API can be used in the browser.
     """
     def process_response(self, request, response):
         response = super().process_response(request, response)
+
         host = request.get_host()
+        origin = request.headers.get("Origin")
+
+        if not origin:
+            origin = "*"
 
         if host.startswith("api."):
-            response["Access-Control-Allow-Origin"] = "*"
+            # Always return HTTP 200 for preflight requests in that context (?)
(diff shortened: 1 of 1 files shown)


matteocargnelutti:
- Added a few notes after our quick chat w/ @jcushman 
- @bensteinberg I would really like to have your thoughts on this 😄

bensteinberg (approved):
Forcing the 200 for preflight seemed weird to me, but it sounds like that is what we are supposed to do here.

matteocargnelutti:
Thank you for having a look @bensteinberg 😄 . 
Minor edit: I ended up setting it to HTTP 204 _(an `OPTIONS` request doesn't expect content)_.

matteocargnelutti:
**For future reference:** @rebeccacremona mentioned that we use a similar approach to deal with preflight issues on Perma's player `<iframe>` [[1]](https://github.com/harvard-lil/perma/blob/956cf9447e69fee9b98352c0af5f42775bc8495f/perma_web/perma/views/common.py#L289) [[2]](https://github.com/harvard-lil/perma/blob/956cf9447e69fee9b98352c0af5f42775bc8495f/perma_web/perma/utils.py#L737).

If we end up finding of a better way of dealing with this issue, it would probably be worth revisiting both this PR and the snippets linked above 🎗️ .

PR #3111, opened 2022-07-05
First cut at instrumentation with Sentry

bensteinberg:


bensteinberg:
If you want to try this out locally, check with me for Sentry DSN and membership.

perma_web/perma/settings/utils/post_processing.py
```diff
@@ -71,3 +71,25 @@ def post_process_settings(settings):
         settings['WORKER_COUNT'] = len([key for key in active.keys() if key.split('@')[0][0] == 'w']) if active else 0
     except TimeoutError:
         pass
+
```
matteocargnelutti: Would moving this block to the beginning of `post_process_settings()` make any difference?

matteocargnelutti (approved):
Looks good to me @bensteinberg 😄 .

perma_web/perma/settings/utils/post_processing.py
bensteinberg: Good point, it would catch subsequent errors in the file.

bensteinberg:
I *think* that to instrument our JS I'd have to add code to both entry points, `static/js/global.js` and `static/js/single-link.module.js`. I'm less clear on how to do settings in JS -- will ask for help if I get stuck. Instrumenting JS can be a separate PR, of course.

matteocargnelutti (approved):
🚀

bensteinberg:
I think I've added Sentry to our JS so that it will only make requests on errors, that is, without the `BrowserTracing` integration.

matteocargnelutti:
@bensteinberg This looks good to me on the implementation front 😄 .

---

**One question:**
Do we want to add `replay.domain.ext` to  Sentry's `ignoreUrls`? 
I think that, by default, exceptions arising from third-party iframes are caught by Sentry (?). 
In our case, even playbacks that work properly might come with dozens of uncaught exceptions that might clutter our logs.

@rebeccacremona : Any thoughts on this?

---

**One thing I think could be worth investigating:**
The error reporting seems to only be considering the bundle - and not the associated map file - when generating reports, which makes us less likely to identify the origin of issues.

Screen Shot 2022-07-06 at 2 43 11 PM

I _think_ it is possible to have Sentry use map files to get more precise reporting, but I might be misremembering.

matteocargnelutti (approved):

bensteinberg:
Yeah, we can use source maps: https://docs.sentry.io/platforms/javascript/sourcemaps/#2-provide-source-maps-to-sentry -- not sure which option makes more sense ("Uploaded directly to Sentry (strongly recommended)" or "Served publicly over HTTP alongside your deployed code"). I also see 

> If you use [Webpack](https://webpack.js.org/) in your project, we recommend generating and uploading your source maps by way of [sentry-webpack-plugin](https://github.com/getsentry/sentry-webpack-plugin).

matteocargnelutti:
@bensteinberg No strong preference personally. 
We have almost everything set up for the self hosted option, [according to their docs](https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/hosting-publicly/).

`https://perma.cc/static/bundles/global.js` finishes with:
```
//# sourceMappingURL=global.js.map
```

_"Just"_ removing `.map` from `.gitignore` so map files are versioned and published might get us there 😄 .

bensteinberg:
As for ignoring errors in replays, I think it's a good idea. What is the cleanest/most reliable way to get `replay.domain.ext` in this context? Is `"replay." + document.location.hostname` safe enough?

rebeccacremona:
> As for ignoring errors in replays, I think it's a good idea. What is the cleanest/most reliable way to get `replay.domain.ext` in this context? Is `"replay." + document.location.hostname` safe enough?

If you want to get it from Django settings: https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/views/common.py#L225

matteocargnelutti:
@bensteinberg I like @rebeccacremona's idea of getting that info from the Django. Alternatively, `ignoreUrls` accepts a regex, so we could use something like `^https\:\/\/replay\.perma[\W\w]+$` ?

rebeccacremona:
I love the idea that we might be able to get reports of -some- kinds of replay errors, maybe with a future version of the replay software, one that makes it easier to distinguish between errors from the replay software itself rather than errors from the playback.

bensteinberg:
In [fb51d0e](https://github.com/harvard-lil/perma/pull/3111/commits/fb51d0e7dbe1c7d492c168a3febed77f7246a153) I think I've suppressed reporting of JS errors in client-side playbacks. Source map change coming up next....

perma_web/npm-shrinkwrap.json
```diff
@@ -1702,7 +1748,7 @@
     "autoprefixer": {
       "version": "6.7.7",
       "resolved": "https://registry.npmjs.org/autoprefixer/-/autoprefixer-6.7.7.tgz",
-      "integrity": "sha512-WKExI/eSGgGAkWAO+wMVdFObZV7hQen54UpD1kCCTN3tvlL3W1jL4+lPP/M7MwoP7Q4RHzKtO3JQ4HxYEcd+xQ==",
+      "integrity": "sha1-Hb0cg1ZY41zj+ZhAmdsAWFx4IBQ=",
```
rebeccacremona: Okay this diff still has some sha512 hashes replaced by sha1:
```
    "autoprefixer": {
      "version": "6.7.7",
      "resolved": "https://registry.npmjs.org/autoprefixer/-/autoprefixer-6.7.7.tgz",
      -"integrity": "sha512-WKExI/eSGgGAkWAO+wMVdFObZV7hQen54UpD1kCCTN3tvlL3W1jL4+lPP/M7MwoP7Q4RHzKtO3JQ4HxYEcd+xQ==",
      +"integrity": "sha1-Hb0cg1ZY41zj+ZhAmdsAWFx4IBQ=",
```
Pulling down and experimenting......

rebeccacremona: You know what, I think we can just revert those undesired changes: nothing untoward seems to happen if we do, and subsequent runs of `install` etc. don't try to change them back.

I note that with it reverted, there are still some 76 sha1 hashes present. The shrinkwrap file on develop has 267. So, we're getting closer.......

PR #3112, opened 2022-07-06
Perma API: CORS Fix update

matteocargnelutti:
**Tentative:** Using `request.get_host()` to determine which requests need CORS headers might prove problematic on prod. This PR uses a sub-segment of `request.get_raw_uri()` instead as a tentative workaround.

Also changed HTTP status code from 204 to 200 _(responses are not empty)_.


bensteinberg (approved):
You could use `settings['API_VERSION']` here if you wanted, though I don't think it's absolutely necessary.

matteocargnelutti:
> You could use `settings['API_VERSION']` here if you wanted, though I don't think it's absolutely necessary.

Ah sorry I saw your comment too late: this is a very good suggestion!

PR #3113, opened 2022-07-06
Perma API: CORS Fix update

matteocargnelutti:
Now that we confirmed that this approach fixes our problem, implements https://github.com/harvard-lil/perma/pull/3112#pullrequestreview-1030208582 as suggested by @bensteinberg


bensteinberg (approved):
👍

PR #3123, opened 2022-07-14
Set cache-from and cache-to in x-bake

bensteinberg:
This is a test of https://github.com/harvard-lil/docker-compose-update-action/commit/d34050bacc80f864e5f202e5595a2e7bd2f5552b

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -10,6 +10,10 @@ services:
         platforms:
           - linux/amd64
           - linux/arm64
+        cache-from:
+          - type=registry,ref='registry.lil.tools/harvardlil/perma-playwright:buildcache'
+        cache-to:
+          - type=registry,ref='registry.lil.tools/harvardlil/perma-playwright:buildcache,type=max'
       x-hash-paths:
         - requirements.txt
   web:
@@ -22,6 +26,10 @@ services:
         platforms:
           - linux/amd64
           - linux/arm64
+        cache-from:
+          - type=registry,ref='registry.lil.tools/harvardlil/perma-web:buildcache'
+        cache-to:
+          - type=registry,ref='registry.lil.tools/harvardlil/perma-web:buildcache,type=max'
       x-hash-paths:
         - perma_web/requirements.txt
         - perma_web/npm-shrinkwrap.json


matteocargnelutti (approved):
LGTM

bensteinberg:
@jcushman do you have any comment on the "Rebuild docker images" section of the build log? It's not clear to me whether
```
#7 [playwright] importing cache manifest from 'registry.lil.tools/harvardlil/perma-playwright:buildcache'
#7 ERROR: invalid reference format
```
is actually wrong, or just means the cache isn't there yet.

Also, I think 
```
+ python /home/runner/work/_actions/harvard-lil/docker-compose-update-action/main/update_tags.py -a load
errors:
denied: requested access to the resource is denied
unauthorized: authentication required
errors:
denied: requested access to the resource is denied
unauthorized: authentication required
```
is just because this is a PR, not a commit?

jcushman:
> @jcushman do you have any comment on the "Rebuild docker images" section of the build log? It's not clear to me whether
> 
> ```
> #7 [playwright] importing cache manifest from 'registry.lil.tools/harvardlil/perma-playwright:buildcache'
> #7 ERROR: invalid reference format
> ```
> 
> is actually wrong, or just means the cache isn't there yet.

Hopefully that the cache isn't there yet? Would see if that one persists.

> Also, I think
> 
> ```
> + python /home/runner/work/_actions/harvard-lil/docker-compose-update-action/main/update_tags.py -a load
> errors:
> denied: requested access to the resource is denied
> unauthorized: authentication required
> errors:
> denied: requested access to the resource is denied
> unauthorized: authentication required
> ```
> 
> is just because this is a PR, not a commit?

Huh, no, I don't know why that one is happening. For context that script is run here:

https://github.com/harvard-lil/docker-compose-update-action/blob/main/action.yml#L39

And the contents are here:

https://github.com/harvard-lil/docker-compose-update-action/blob/main/update_tags.py

"denied: requested access to the resource is denied" looks like the stderr from a docker command that isn't logged in. No docker command is run at the top of that script ... so let's assume that stderr is being printed in the log before stdout, even though the docker command comes later? But then, ok, what docker command isn't logged in? The only one I see in the script is `remote_tag_exists`, which should run `docker manifest inspect registry.lil.tools/harvardlil/perma-playwright:4-8ab77242e64440f2323b0ed4d63b8b58` to check if that tag exists yet. I think that command succeeds locally (i.e. reports unknown artifact since the new tag doesn't exist) even when logged out, so I don't know why it would print an error.

If `remote_tag_exists` is indeed the command that's complaining, the effect of it not working as expected would be that the tests would rebuild the images when they don't need to, as indicated by `Returning services-to-rebuild: ['playwright', 'web']`. So we could just charge ahead and see later if more work is being done than necessary.

jcushman (approved):

bensteinberg:
The push failed: `error: failed to solve: unknown cache exporter: "max'"`

jcushman:
The examples I saw didn't have single quotes; are those needed? Looks like it's reading them literally.

bensteinberg:
Ah, I bet the error of having `type` as the key for `max` is why I had to use the quotes, let me adjust #3124 accordingly

PR #3125, opened 2022-07-20
API CORS: Added "Content-Type" to the list of allowed headers

matteocargnelutti:
**Reason for adding this headers to the allow list:** 
Some preflight requests may have `Content-Type: "application/json"` in their headers, which may get rejected depending on client implementation of CORS.


bensteinberg (approved):
LGTM

PR #3126, opened 2022-07-20
For more efficient calculation of CaptureJob.order

rebeccacremona:
...especially when `human=False`.

Follow the adventure on Slack:
- diagnosing https://hlslil.slack.com/archives/C07URASMC/p1658246493230719
- fixing https://hlslil.slack.com/archives/C07URASMC/p1658344710507349

There is an extra migration in here, one I forgot to add back in https://github.com/harvard-lil/perma/commit/327ebd29a8bf6fa4e0d2d0adce5357590d947983... whoops.


bensteinberg (approved):
👍

PR #3127, opened 2022-07-21
More efficient calculation of CaptureJob.order, take 2

rebeccacremona:
Further discussion suggests that [adding an index](https://github.com/harvard-lil/perma/pull/3126) is not the best way to solve our problem: the query for adding capture jobs to the end of the queue is slow if `human=False`. Instead of adding an index, and thus making insert/update/delete somewhat slower and RAM pressure somewhat higher... we can remove the filter, and add to the end of the queue without considering the value of `human`. (Thanks, @jcushman !)

Since that code has not yet been deployed to prod, we are reverting instead of rolling forward. This PR deletes the migration that creates the index and tweaks the query as suggested.

This PR also adds (or more accurately, retains from #3126/does not revert) a migration that I forgot [back in May](https://github.com/harvard-lil/perma/commit/327ebd29a8bf6fa4e0d2d0adce5357590d947983), which we realized while working on this issue. 

Before deploying to stage, we will need to run `python manage.py migrate perma 0003_auto_20220120_1522` on stage; I do not know for sure if the following stage deployment will proceed as usual, but think it will.


matteocargnelutti (approved):
Yeeeeeeee 🚀 !!!

PR #3133, opened 2022-08-01
Hide the replayweb.page chrome unless `embed=False` is requested.

rebeccacremona:
The team decided that we'd rather hide the replayweb.page interface by default, displaying it only on demand or to power users. This PR hides it unless you supply the GET param `embed=False`. It does not add anything to the UI for power users: this will be just for us (or anyone who reads the code), for the time being.

New default:

image

With &embed=False:
image


bensteinberg (approved):
Thanks!

PR #3136, opened 2022-08-04
Wait for <h1> instead of just <html>

### LINKED ISSUE #3135: `test_create_link` needs to wait
Sometimes, `test_create_link` manages to load the deeply-nested element we use to check if the playback succeeded too quickly, and we fail with:
```
AssertionError: assert 'Example Domain' in ''
```
Let's tweak so that this test is less flaky. A sleep? Or, maybe, just making the selector more detailed, so that Playwright's built-in waiting mechanism works even better: perhaps we wait for the `h1`?

This should be quick to fix.

rebeccacremona:
Closes https://github.com/harvard-lil/perma/issues/3135

This passed 20 times in a row locally.


matteocargnelutti (approved):
🚀  !

PR #3138, opened 2022-08-04
Don't target the submitted URL; target the primary capture's URL.

rebeccacremona:
In most cases, a `Link`'s `.submitted_url` is the same as the `Link`'s [`.primary_capture.url`](https://github.com/harvard-lil/perma/blob/develop/perma_web/api/views.py#L486)... but not when a user uses the upload-your-own feature to [replace a broken capture with a PDF or image](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/models.py#L1702).

This PR updates client-side playbacks to point at the correct target URL... as we've long been doing with [server-side playback](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/views/common.py#L243).


matteocargnelutti (approved):
LGTM

Issue #3143, opened 2022-08-08
Add `noCache` to `<replay-web-page>` 


matteocargnelutti:
Prevent `<replay-web-page>` from caching payloads when loading `.warc.gz` files by using the `noCache` attribute [that was added to `<replay-web-page>` 1.6.4](https://github.com/webrecorder/replayweb.page/commit/6c6d66f03a86997d5fb8c7368664d326bb658420).

Quickly investigate and report on overall browser-side cacheing situation _(we know what `<replay-web-page>` stores, not what the browser decides to cache)_ to help determine that impact this feature could have on our bandwidth usage. 

Also have a look at our usage of [Clear-Site-Data: '"cache", "storage"'](https://github.com/harvard-lil/perma/blob/356567698d29d027c9885cee58fad7696b36ecf5/perma_web/replay/views.py#L35), and see if it's still necessary with `<replay-web-page noCache>`

matteocargnelutti:
Related: #3038

matteocargnelutti:
Implemented and merged -- Discussion about caching in #3142

PR #3145, opened 2022-08-13
replayweb.page 1.6.4 + `noCache`

matteocargnelutti:
✅  Implements #3143 

## Caching issue workaround 
Something _interesting_ happened while working on this PR: everything seemed to work fine, but a few minutes after a first successful playback, `<replay-web-page>` would fail to load the archive. 
How many minutes you ask? 15 exactly 👀 , the value of `WARC_PRESIGNED_URL_EXPIRES`. 

It seemed like `<replay-web-page>` was ignoring changes to its `source` attribute, and kept requesting an expired WARC url. 

It took me a while to figure out that it was actually the HTML documents for both `[GET] perma.test/{guid}` and `[GET] replay.perma.test/?guid={guid}` that were cached by the browser for longer than the WARC url was valid for. 

I repurposed the `bypass_cache_middleware()` middleware to make sure the HTML document served by `perma.test/{guid}` is not cached by the browser, and added the same headers to the response of `replay.perma.test/?guid={guid}`.

This seems to work locally, let's see how it goes on staging ... and prod?  
Let's also see if we want to fine-tune this: do we maybe want to let the browser cache these documents, but only for `WARC_PRESIGNED_URL_EXPIRES` seconds?

## Also included in this PR:
- Some of refactoring of `iframe.html`'s JavaScript. 
As I was trying to figure this caching situation out, I ended up re-using some of the clean-up code @rebeccacremona devised in multiple places. It therefore made sense to modularize things, which I ended up doing for the entire file. 
Happy to make changes, or revert to its original state. 
- Transition to `<script type="module">` so we can use top-level await, and more reliably predict when things are run. 
- A note in `install.md` to help deal with `mkcert` + Mac OS issue, when things go wrong.  
- Changed `DEFAULT_PLAYBACK_MODE` in `settings_common.py` so client-side playback.  

## Other caching considerations 
When using `noCache`, `<replay-web-page>` only stores CDX content in indexedDB, alongside some meta data about the  file it parsed. It uses this information to make ad-hoc range requests to the WARC file in order pull resources. 

I _think_, based on my observations of the network inspector on Chrome and Firefox, that the browser does cache these resources, as it would if they were loaded in a _"normal"_ context. 
It is not clear to met, yet, if the original `Cache-Control` header of the captured resource applies, or if it's been overridden somewhere in the process. TBD.

In any case, using `noCache` will mechanically generate many more HTTP requests to the WARC file _(many "small" instead of 1 large)_ than our current setup. Impact on our infrastructure: TBD as well.


perma_web/perma/middleware.py
```diff
@@ -20,11 +21,17 @@ def bypass_cache_middleware(get_response):
 
     def middleware(request):
         response = get_response(request)
-        if request.path.startswith(LOGIN_ROUTE) or request.path.startswith(LOGOUT_ROUTE):
+
+        if (request.path.startswith(LOGIN_ROUTE) or 
+              request.path.startswith(LOGOUT_ROUTE) or 
+              re.match(r'^\/[A-Z0-9]{4}\-[A-Z0-9]{4}', request.path)):
```
rebeccacremona: Ah, interesting idea to do this here. Maybe a good reminder to have it along with the code for the Cloudflare caching cookie, so that if we ever turn Cloudflare caching back on for these pages, we recall that things are complicated!

I wonder if a clearer alternative, that would not require route matching, would be to add to the responses in the view [here](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/views/common.py#L247) like you did with the replay.perma.test view? Do you prefer to handle this one at the middleware level?

If keeping the middleware approach, let's import and reuse this [regex pattern for GUIDS](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/urls.py#L17), so it catches old-style Perma Links like https://perma.cc/02EHS9c9WsR. (There are about 6K links with GUIDs like that, all of which start with "0")

rebeccacremona (changes requested):
Fantastic! Thanks for this and all the research/exploration that went into it!

I am especially appreciative of the refactor of the JS in `iframe.html`! It's great for me to read and see a clearer way of structuring things.

The adventure w/ `mkcert` and MacOS trust is such a bummer... So weird that that happens to some people.... And only 30 steps to fix 😅.

One quick thing in the middleware: I think either the regex needs tweaking, or this wants to be done in the view. But other than that, LGTM!

Since you said you want to discuss w/Ben before this goes to prod, maybe let's not merge until then, in case we need to merge security patches, etc. before then?

perma_web/perma/middleware.py
matteocargnelutti: @rebeccacremona I wasn't too sure, so figured I would group features that _"change headers on a subset of responses in order to deal with caching-related issues"_ together.

It makes a lot more sense to do it at `views` level in that case 😄 ! Moved it there. Thanks!

matteocargnelutti:
@rebeccacremona Thanks a lot for having a look! 

Yes, if everything looks good to you, we may want to approve this PR and make sure to discuss its devops-y implications with  @bensteinberg next week before merging 😄. 

Cheers.

rebeccacremona (approved):

perma_web/perma/middleware.py
rebeccacremona: Either approach seems fine to me 🤷‍♀️; thanks for making the tweak, looks good!

bensteinberg:
I'm taking the liberty of merging this so we can get it on stage.

bensteinberg:
(Whoops, jumped the gun, images from previously-merged PR were not built and pushed yet. I will re-run this build when that's done.)

Issue #3032, opened 2022-02-04
Title field length


rebeccacremona:
Looks like we also need to do this for [`Link.submittled_title`](https://github.com/harvard-lil/perma/pull/3015).

```
Exception Type: DataError
Exception Value: value too long for type character varying(2100)
```

PR #3150, opened 2022-08-25
updates instructions and fixes bug 3032

kilbergr:
I updated a couple of instructions to get going for perma.cc and fixed this bug: https://github.com/harvard-lil/perma/issues/3032
I did test with pdb, but I'm not sure about the kinds of expected unit tests. LMK!
Screen Shot 2022-08-25 at 9 55 15 AM

### DIFF
--- a/developer.md
+++ b/developer.md
@@ -31,7 +31,7 @@ SSL certs and keys should already be present. If they are not, or if they have
 expired, you can run `bash make_cert.sh` to generate new files.)
 
 To log in and explore Perma, try logging in as one of our
-[test users](https://github.com/harvard-lil/perma/blob/develop/perma_web/fixtures/users.json). All test users have a password of "pass".
+[test users (the `linkuser` objects)](https://github.com/harvard-lil/perma/blob/develop/perma_web/fixtures/users.json#L167). All test users have a password of "pass".
 
 The server will automatically reload any time you made a change to the
 `perma_web` directory: just refresh the page to see your changes.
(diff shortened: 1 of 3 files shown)


matteocargnelutti (approved):
Thanks a lot @kilbergr 😄 ! LGTM.

> I did test with pdb, but I'm not sure about the kinds of expected unit tests. LMK!

As you can see, this part of the codebase is only partially and indirectly covered by our test suite. I think it is okay as is, but that is something we may want to revisit 😄 .

Cheers.

rebeccacremona:
🎉 Awesome!

+1 on Matteo's approval, but quick observation: I think we can truncate title at the much longer [`2100`](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/models.py#L1362) chars instead of the [`300`](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/models.py#L1363) at which we truncated description. 

I cannot begin to imagine why, in 2013, title was added with a 2100 char limit and description, which I would have expected to be longer that title for most pages on the web when present, was added in 2017 with a 300 char limit: that seems wrong... But, given that is in fact how our models are set up, we should probably slice accordingly.

kilbergr:
@rebeccacremona @matteocargnelutti can I get one of you two to approve the workflows for me (for my first contribution)? https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks#approving-workflow-runs-on-a-pull-request-from-a-public-fork

Issue #3152, opened 2022-08-31
Case-insensitive email addresses


matteocargnelutti:
As reported by @clare-stanton: Some of our users are running into issues because email addresses are considered case sensitive during sign-up / sign-in.

- [x] Figure out if the entire email address, or just the part before `@` should be case insensitive
- [x] See how many existing users would be impacted by proposed fix _(and how)_
- [x] Choose our preferred solution: see this [list of decisions to be made](https://docs.google.com/document/d/10mZ2e2evg-A5H-Jdv4FUg7DITqw-nxoWyiEv6x7lbNU/edit#heading=h.gqw169l0mpvz)
- [x] Apply the fix and remediate edge cases: [project board](https://github.com/orgs/harvard-lil/projects/11/views/1)

---

**Resources:**
- [Design Doc](https://docs.google.com/document/d/10mZ2e2evg-A5H-Jdv4FUg7DITqw-nxoWyiEv6x7lbNU/edit#)
- @lizadaly's work and thoughts on fixing a similar problem on H2O: https://github.com/harvard-lil/h2o/pull/1592

kilbergr:
I found this blog post referencing RFCs around case sensitivity in email: https://verifalia.com/help/email-validations/are-email-addresses-case-sensitive. It says that domains are not case-sensitive, which would be my instinct. Regarding the username, it says that though they can be case-sensitive, it's rare that they are. I have found other articles citing the same RFC5321 (https://datatracker.ietf.org/doc/html/rfc5321) confirming that they can be case sensitive. I wonder how commonly it's an issue though. The problem is that people are putting in MYUSER@dogs.cats and myuser@dogs.cats and they're being treated as two separate users, right? Or is it you're getting myuser@dog.CATS and myuser@Dog.Cats? Or is it both?

matteocargnelutti:
@kilbergr Main issue is the former, as [Django follows said RFC](https://django-improved-user.readthedocs.io/en/stable/email_warning.html), in `EmailField` and `normalize_email()`.

Another argument in favor of considering usernames as case sensitive _(or at least saving them that way)_, besides the fact that it is the most "RFC compliant" thing to do, is that it sometimes makes sense to save how users prefer to see their email displayed _(ie: `PersonName@domain.ext` being arguably easier to read than `personname@domain.ext`)_.

[...]

PR #3153, opened 2022-08-31
Case insensitive email addresses

matteocargnelutti:
First attempt at figuring out a solution for #3152. 

---

I decided to start with the smallest fix I could think of to deal with this issue, and see where the discussion takes us 😄 .

### What does it do?
- It forces `str.lower()` on the `email` field of `UserForm`, so all new users will have a lower case email. This applies to the entire email address.

### Why this might be a good idea?
- It deals with our problem _"moving forward"_, without creating problems for existing users (?).
- We can then handle existing casing conflicts on a case-by-case basis, as brought to our attention.

### Why this might not be a good idea?
- The sign-in form remains case sensitive to accommodate for existing records (and conflicts). 
One can easily imagine users registering as `PersonName@domain.ext` and try to sign-in as `personname@domain.ext`, which would not work.
  - Is this something that can be fixed with a better error message?
  - Or would we want to implement an _"auto-retry"_ on the sign-in form, the second attempt using a lowercase version of the email address?
- [Saving email usernames with their original casing but checking them in a case-insensitive way is sometimes recommended](https://django-improved-user.readthedocs.io/en/stable/email_warning.html). 

---

PS: @rebeccacremona 

I haven't added a test yet, and wanted to ask how you would go about testing this.

**First idea:** adding a test case in `test_views_user_management.py`, something that tries to call the `/sign-up` form twice with the same email, but different casing.

```python3
def test_signup_email_case_insensitive(self):
    self.submit_form('sign_up',
                    data = {'e-address': "personname@test.com", 'first_name': 'Person', 'last_name': 'Name'},
                    success_url = reverse('register_email_instructions'))

    self.submit_form('sign_up',
                    data = {'e-address': "personNAME@test.com", 'first_name': 'Person', 'last_name': 'Name'},
                    error_keys=['email'])
```

---

Ping @kilbergr

### DIFF
--- a/perma_web/perma/forms.py
+++ b/perma_web/perma/forms.py
@@ -137,6 +137,10 @@ class UserForm(forms.ModelForm):
         super().__init__(*args, **kwargs)
 
     def save(self, commit=True):
+        # From Sept 2022 onwards: all email addresses are lowercase
+        if self.instance.email:
+            self.instance.email = self.instance.email.lower()
+
         # save user, and set a password so that the password_reset flow can be
         # used for email confirmation
         self.instance.set_password(LinkUser.objects.make_random_password(length=20))


rebeccacremona:
I am trying to think of anything else we'd need to change if we do this... 

So far, I can only think of one other spot: on the `https://perma.cc/manage/{admin,registrar,sponsored,organization}-users pages`, there is a button for adding a user, e.g. 
image
(click that one to expose the actual button in question)
image

It [does a thing](https://github.com/harvard-lil/perma/blob/f681d87ab3f86fe1c5e53e0258df22056aaba911/perma_web/perma/views/user_management.py#L716-L717), where it [checks to see if a user already exists](https://github.com/harvard-lil/perma/blob/f681d87ab3f86fe1c5e53e0258df22056aaba911/perma_web/perma/views/user_management.py#L734-L753), and [directs you to a different form](https://github.com/harvard-lil/perma/blob/f681d87ab3f86fe1c5e53e0258df22056aaba911/perma_web/perma/views/user_management.py#L724-L726) depending. 

I think downcasing on new `UserForm` save will get weird, if we don't also add special handling here. How should this work, thinking.....

rebeccacremona:
Does this sound right?

- An org user wants to add Joe Yacoboski to their org, and enters `Joe_Yacoboski@example.com` to the form
- We check to see if there's already an account associated with `"Joe_Yacoboski@example.com"` or `"Joe_Yacoboski@example.com".lower()`.

- If there is one associated with `Joe_Yacoboski@example.com`, we send them to the "he's already got an account; add him to the org" form with that User object.

- If there is not, but there is exactly one associated with `"Joe_Yacoboski@example.com".lower()`, we send them to the "he's already got an account; add him to the org" form with _that_ User object. If there is more than one associated with `"Joe_Yacoboski@example.com".lower()`, we tell them that there is a complication and ask them to contact us.

- Otherwise, we send them to the "create a new user and add them to the org" form, with the pre-filled email address field downcased, so that they see what we are doing.

That will work, I think?

matteocargnelutti:
@rebeccacremona That makes sense to me and I think it's a valid solution!

I'm trying to think of an alternative, to see if we can "get away" with a solution that doesn't involve adding more complexity to this system. 

How about we switched [this filter](https://github.com/harvard-lil/perma/blob/f681d87ab3f86fe1c5e53e0258df22056aaba911/perma_web/perma/views/user_management.py#L740) to `email__iexact` so this mechanism catches existing accounts regardless of casing?
If this returns more than 1 account, we should probably show and error message. 

Since we know how many accounts would be affect by this, dealing with this problem on a per-case / request basis seems somewhat reasonable?

rebeccacremona:
Re: testing...

Right, the test fixtures in Perma are still the old-fashioned JSON style, rather than the nicer pytest fixture + factoryboy setup we have elsewhere....

We probably want to create another [user like this one](https://github.com/harvard-lil/perma/blob/develop/perma_web/fixtures/users.json#L238), maybe `Test_User@example.com` (which also means adding a [folder](https://github.com/harvard-lil/perma/blob/develop/perma_web/fixtures/folders.json) for them), and maybe `TEST_USER@example.com` as well, so we can test explore the above imaginary Joe Yacoboski scenario.

I like your idea of a test hitting the most basic signup form twice. I feel like we want some sort of assertion that it works via the other forms well, even though that sounds tiresome.

rebeccacremona:
> How about we switched [this filter](https://github.com/harvard-lil/perma/blob/f681d87ab3f86fe1c5e53e0258df22056aaba911/perma_web/perma/views/user_management.py#L740) to `email__iexact` so this mechanism catches existing accounts regardless of casing?
> If this returns more than 1 account, we should probably show and error message.

Sure, showing an error whenever there is > 1 sounds just as good or better.

I'd be interested talking through the flow for people who attempt to sign up with capital letters.

I think without more code, this is what we have: I enter `Joe_Yacoboski@example.com`, possibly let LastPass or my browser remember that, submit, succeed... but then, I have to log in as `joe_yacobowski@example.com`, because login is case-sensitive... so I fail. I try to reset my password. I enter `Joe_Yacoboski@example.com` in the password reset form. I do not get sent a reset email, because that, like login, is case-sensitive.

How can we smooth that out? Nothing great is coming to mind: rejecting capital letters in the signup form with a validation error seems bad too.

rebeccacremona:
I am going to study the seemingly duplicative accounts we have, and see how many can be easily merged/deleted (for instance, if only one account has been activated, or only one has links, etc.)

rebeccacremona:
I am making bold to convert this to a draft, pending further discussion about auth-related flows for users.

rebeccacremona:
Superseded by https://github.com/orgs/harvard-lil/projects/9/views/3

PR #3166, opened 2022-09-21
Rek pp perma instructions

kilbergr:
This PR adds instructions and settings to allow users to run perma on the same network as perma-payments locally. There is repetition in the settings file, but that seems ok given how other settings files work and that the settings need to overwrite a var set in the core setting file from which every other file inherits.
It also adds a ToC to the dev README.
Works in conjunction with https://github.com/harvard-lil/perma-payments/pull/155

### DIFF
--- a/developer.md
+++ b/developer.md
@@ -5,6 +5,37 @@ This document contains tips and tricks for working with Perma.
 
 See the [installation documentation](./install.md) to get up and running.
 
+  - [Perma - developer notes](#perma---developer-notes)
+   - [Common tasks and commands](#common-tasks-and-commands)
+     - [Run Perma](#run-perma)
+     - [Run the tests](#run-the-tests)
+     - [Run a particular test (python only)](#run-a-particular-test-python-only)
+     - [Update the python dependencies](#update-the-python-dependencies)
+     - [Update the node dependencies](#update-the-node-dependencies)
+     - [Update Chromium](#update-chromium)
+     - [Migrate the database](#migrate-the-database)
+     - [Reset the database](#reset-the-database)
+     - [Run arbitrary commands](#run-arbitrary-commands)
+   - [Git and GitHub](#git-and-github)
+   - [Logs](#logs)
+   - [Code style and techniques](#code-style-and-techniques)
+     - [User roles and permissions tests](#user-roles-and-permissions-tests)
+     - [Javascript templates](#javascript-templates)
+     - [Sending email](#sending-email)
+     - [Asset pipeline](#asset-pipeline)
+     - [Managing static files and user-generated files](#managing-static-files-and-user-generated-files)
+     - [Hosting fonts locally](#hosting-fonts-locally)
+   - [Schema and data migrations](#schema-and-data-migrations)
+     - [Schema migrations and data migrations](#schema-migrations-and-data-migrations)
+     - [Track migrations in Git and get started](#track-migrations-in-git-and-get-started)
+   - [Testing and Test Coverage](#testing-and-test-coverage)
+     - [Linting with flake8](#linting-with-flake8)
+   - [Working with Celery](#working-with-celery)
+   - [Working with Redis](#working-with-redis)
+   - [Running with DEBUG=False locally](#running-with-debugfalse-locally)
+   - [Perma Payments](#perma-payments)
+     - [Test Perma Interaction with Perma Payments](#test-perma-interaction-with-perma-payments)
+   - [Internet Archive](#internet-archive)
 
 Common tasks and commands
 -------------------------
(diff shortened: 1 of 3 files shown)


lizadaly (approved):
I've never run payments but this all looks reasonable to me. I assume it works!

rebeccacremona:
Before merging, let's give one more think about whether we can reduce any of the duplication in the compose files and settings.

developer.md
```diff
@@ -313,6 +344,21 @@ Aspects of Perma's paid subscription service are handled by the companion applic
 
 By default, Perma's `docker-compose` file will spin up a local Perma Payments for you to experiment with. For more fruitful experimentation, configure this Perma Payments to interact with Cybersource's test tier, by running Payments with a custom settings.py that contains our credentials. See `docker-compose.yml` and `/services/docker/perma-payments/settings.py.example` for more information. CyberSource will not be able to communicate its responses back to your local instance, of course, but you can simulate active subscriptions using the Django admin.
 
+### Test Perma Interaction with Perma Payments
+
+You may also decide to run the two services locally to experiment by running both on the same network.
```
rebeccacremona, line 349: What would you think of:
```suggestion
You may also decide to run the two Docker Compose projects side-by-side on the same network.
```

developer.md
```diff
@@ -313,6 +344,21 @@ Aspects of Perma's paid subscription service are handled by the companion applic
...
+
+The Docker containers for `perma-payments` have to be built before you can successfully run  `perma` on the same network. Visit the [`perma-payments` repo](https://github.com/harvard-lil/perma-payments/blob/develop/README.md#running-locally) for those instructions. Once you've done that, build the Docker containers for perma:
```
rebeccacremona, line 351:
```suggestion
The Docker containers for `perma-payments` must be up and running before you can successfully run  `perma` on the same network. Visit the [`perma-payments` repo](https://github.com/harvard-lil/perma-payments/blob/develop/README.md#running-locally) for those instructions. Once you've done that, start up the Docker containers for perma with the below command, which instructs Docker to ignore `docker-compose.override.yml`: 
```

developer.md
```diff
@@ -313,6 +344,21 @@ Aspects of Perma's paid subscription service are handled by the companion applic
...
 
+### Test Perma Interaction with Perma Payments
+
+You may also decide to run the two services locally to experiment by running both on the same network.
+
+The Docker containers for `perma-payments` have to be built before you can successfully run  `perma` on the same network. Visit the [`perma-payments` repo](https://github.com/harvard-lil/perma-payments/blob/develop/README.md#running-locally) for those instructions. Once you've done that, build the Docker containers for perma: 
+
+- `docker compose -f docker-compose.yml up -d --build`
+
+Then, to run perma, run:
```
rebeccacremona, line 355:
```suggestion
Then start perma's development server normally:
```

docker-compose.override.yml
```diff
@@ -17,3 +17,55 @@ services:
...
+    # volumes:
+    #   - ./services/docker/perma-payments/settings.py:/perma-payments/config/settings/settings.py
+    networks:
+      - default
+      - perma_payments
+
+networks:
+  default:
+  # perma_payments:
+      # name: perma-payments_default
```
rebeccacremona: Interesting to see this is commented out!

Have you tested the normal setup, where Perma Payments is not running externally, and Perma's containers have been started with plain old `docker compose up -d`? If yes, and if that works, then I bet that means the entire `network` stanza can be dropped here.

developer.md
```diff
@@ -313,6 +344,25 @@ Aspects of Perma's paid subscription service are handled by the companion applic
...
+
+`d fab run`
+
+When you're finished, take down the containers by running:
+
+`docker compose -f docker-compose.yml down`
+
+Don't worry if you get the following error when shutting down these containers while still running the `perma-payments` containers. That's because the network is maintained until both containers are down:
```
rebeccacremona:
```suggestion
Don't worry if you get the following error when shutting down these containers while still running the `perma-payments` containers. That's because the network is maintained until both Docker Compose projects are down:
```

rebeccacremona (commented):
Thanks for slogging through this!!

docker-compose.override.yml
kilbergr: Yes, I have. I thought the default would need to be set. I'll see what happens without it.

kilbergr: I think I'm going to leave the webrecorder piece and have it and the networks just commented out for people to add according to their preference.

rebeccacremona: Okay! To me, having it present in `docker-compose.yml` is sufficient, but if you think having something in the override file too is better, cool cool!

rebeccacremona (approved):
🎉 👏  

Let's squash-merge this one :-)

docker-compose.override.yml
kilbergr: I think you know better than I how people actually use this, so I am removing.

PR #3170, opened 2022-09-28
Clean up a bunch of old stuff

rebeccacremona:
The unified diff is a little difficult to review; commit-by-commit I think it's easy to see that this is trivial.

The only interesting thing that I'm ripping out is the middleware that can be used to set a cookie that cloudflare page rules can look out for. We used to tell Cloudflare to cache everything, unless it saw that cookie, and we ran into problems, so we turned off the caching. The presence of this middleware has confused us on several occasions. (We may also want to delete the disabled page rules entirely.)

See settings_common.py for a bunch of feature flags and old settings I'm removing. We should check to see if we are setting them in production, and if yes, stop doing so in tandem with deploying this PR (cc @bensteinberg).


perma_web/perma/templates/user_management/settings-usage-plan.html
```diff
@@ -227,60 +227,58 @@ <h3 class="body-ch">Beginning {{ subscription.pending_change.effective | date:"F
...
+          <dd>
+            {{ purchase_history.total_links }} Links
+          </dd>
+          <dt>
+            REMAINING
+          </dt>
+          <dd>
+            {{ request.user.bonus_links }} Links
+          </dd>
+        </dl>
```
rebeccacremona, line 282: This diff is just removing a wrapping `{% if ENABLE_BONUS_LINKS %}` and `{% endif %}` and outdenting what was inside: I didn't actually change the anything else.

rebeccacremona:
Converted to a draft until https://github.com/harvard-lil/perma/pull/3169 is on stage, so we can look at them there separately.

bensteinberg:
I propose to remove the following lines from `settings.py` in staging and production deployments:
```
ENABLE_AV_CAPTURE = True
ENABLE_BONUS_LINKS = True
ENABLE_SPONSORED_USERS = True
```

lizadaly (approved):
Let’s goooo

Issue #3159, opened 2022-09-08
Google search results appearance


rebeccacremona:
If you google https://perma.cc, the snippet is uninformative
![image](https://user-images.githubusercontent.com/11020492/189149066-4c5dd959-5e35-42b5-abf3-e0fa99028c31.png)

I think we probably want to add / to robots.txt? (https://perma.cc/robots.txt) Unless adding / adds everything under slash?

rebeccacremona:
As of today:
![image](https://user-images.githubusercontent.com/11020492/193627856-afed3def-b543-443e-96d9-6fe7f3d72713.png)

The main results still say, "No information is available for this page." I'll check again next Monday.

rebeccacremona:
The main results still say, "No information is available for this page."

mdellabitta:
It looks like `/` is specified in robots.txt as `/$`, which is odd, although not technically incorrect?

Best thing to do would be to confirm in Google Search Console.

PR #3171, opened 2022-09-29
update robots.txt to allow homepage to be crawled for google search results

sabzo:
This addresses https://github.com/harvard-lil/perma/issues/3159. 

The way to truly find out is to publish and wait for Google to crawl and then observe resuts


rebeccacremona (approved):
LGTM!

PR #3174, opened 2022-09-29
Correct requests to api.perma.cc that don't specify a route

sabzo:


sabzo:
This ticket is ready to review :)

sabzo:
Are there unit tests to write for this feature?

bensteinberg (commented):
These redirects are already working correctly, I think, or at least they do for me. What symptom or issue are you seeing that this PR addresses?

sabzo:
@bensteinberg This solves https://github.com/harvard-lil/perma/issues/3159 where Google doesn't show the links on search.

sabzo:
This PR tries to address the scenario where If a developer accidentally `curl -X POST https://api.perma.cc/` (instead of POSTing to one of the actual API endpoints), our app tries to send them an HTML error page, but the error page itself throws a 500 error, because it only works for perma.cc routes, not api.perma.cc routes.

bensteinberg (approved):
LGTM. Thanks!

Issue #3178, opened 2022-10-03
Folder tree doesn't handle paginated results


rebeccacremona:
The javascript for [loading the folder tree](https://github.com/harvard-lil/perma/blob/develop/perma_web/static/js/folder-tree.module.js#L237) assumes that all results will be returned in a single page. By default, Perma's API calls have an implicit [LIMIT 300](https://github.com/harvard-lil/perma/blob/0e9db948de89ee93b62af820f80732aef28aa5c9/perma_web/perma/settings/deployments/settings_common.py#L621). So, if a folder has more than 300 children, only the first 300 children are displayed.

This javascript should be updated to handle paginated results.

Until we have time to do that, we can update the API to use a limit known to be large enough (e.g., advised by `SELECT parent_id, COUNT(*) as children FROM perma_folder GROUP BY parent_id ORDER BY children DESC`)

PR #3179, opened 2022-10-03
Work-around: load up to 500 child folders at a time

rebeccacremona:
This PR adds a short-term fix for #3178.

Right now, no Perma folder has many more than 300 children: the largest counts are 335, 331, 278, 231, 225, 209, 208, 206, 184, 181, 179, 176, 170, 159, 154, 145, 143, 142, 141...

If we request up to 500 children at a time, that should buy us some time to update this JS properly.


kilbergr (approved):
This looks good--do you want to write a little comment above it to indicate it's a temp fix?

Issue #3182, opened 2022-10-05
Upgrade to `<replay-web-page>` 1.7.0


matteocargnelutti:
Upgrade to freshly-released `<replay-web-page>`.

 - [x] Dependency update
 - [x] Remove auto `noCache` logic for browsers that do not support the `StorageManager.estimate` API (now handled at `<replay-web-page>` level)
 - [x] Remove auto `noWebWorker` logic for Safari < 16 (now handled at `<replay-web-page>` level)
 - [x] Add `requireSubdomainIframe` (new). 
 - [x] Add `sandbox` if necessary (new, replaces `noSandbox`)
 - [x] Update styling as necessary (i.e: background is now transparent)

Issue #3140, opened 2022-08-05
Replace the "welcome" text at replay.perma.cc...


rebeccacremona:
...with text that explains we are clearing out your storage and asks you to please stand by. Maybe even a progress indicator of some kind.

matteocargnelutti:
@rebeccacremona Are we ok with the replacement, or should we leave this issue open?

rebeccacremona:
LGTM!

PR #3183, opened 2022-10-05
`<replay-web-page>` 1.7.0 and other minor playback page edits

matteocargnelutti:
Implements #3182 and proposes "something" for #3140 

---

### Proposed replacement for _"Welcome"_ when landing on `replay.perma.cc`:

Screen Shot 2022-10-05 at 3 17 12 PM

---

### Other considerations
Tightened sandboxing rules for parent `<iframe>` on playback page. Will need to test extensively on staging to make sure it doesn't interfere with playback.


perma_web/replay/templates/iframe.html
```diff
@@ -21,31 +33,59 @@
...
+      color: #333333;
+      padding: 1rem;
+    }
+
+    #no-playback div {
+      text-align: center;
+    }
+
+    #no-playback p {
+      font-size: clamp(20px, 3vmax, 36px);
```
lizadaly, line 72: I just started using `clamp()` and I love it.

lizadaly (approved):

rebeccacremona (approved):
Huzzah!

One thought: on reading, "Tip: You may use this page to clear your playback cache," I might ask, "Cool, how?" Perhaps something like, "Tip: visiting this page clears your playback cache" would be clearer?

perma_web/replay/templates/iframe.html
rebeccacremona: Ooh, I have not yet learned about `clamp()`.

matteocargnelutti:
Thank you both, @lizadaly and @rebeccacremona !

---

@rebeccacremona: Yes indeed! Edited as suggested 😄 .

Issue #3132, opened 2022-07-29
Check for `Content-Type: application/json` in `[POST] /v1/archives/batches` and error-out early


matteocargnelutti:
In reference to https://github.com/harvard-lil/perma/issues/2529 

It could be interesting to make API views that only support `Content-Type: application/json` to error out early when using an incompatible content type, beyond what Django Rest Framework does automatically for us.

As far as I know, `[POST] /v1/archives/batches` is the only route that has this issue. 

Alternatively, we could amend this route so it also supports `multipart/form-data`, like other routes do.

sabzo:
Adding notes from a conversation with @rebeccacremona:

### How to cause the error
There is a route in Perma that expects you to POST data in JSON format. https://perma.cc/docs/developer#batches
If a person forgets to include `-H 'Content-Type: application/json'`   for ex:

`curl -H "Authorization: ApiKey REMEMBER-API-KEY-HERE!!" -X POST -d '{"urls": ["http://example.com", "http://example.com/page", "http://example.com/another-page"], "target_folder": 24}' https://api.perma.test:8000/v1/archives/batches` (and curl sends the data in `multipart/form-data` format instead), we throw an internal server error, which looks like this:

### What the error looks like
When DEBUG=False
```
[ERROR] log.py 222: Internal Server Error: /v1/archives/batches
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.9/dist-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.9/dist-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.9/dist-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.9/dist-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
[...]

PR #3185, opened 2022-10-12
if application content is not json, return 400

sabzo:
Addresses  #3132

### DIFF
--- a/perma_web/api/views.py
+++ b/perma_web/api/views.py
@@ -693,7 +693,7 @@ class LinkBatchesListView(BaseView):
         if not request.user.is_authenticated:
             raise PermissionDenied()
         if request.content_type != 'application/json':
-            content = {'content-type error': 'content-type must be aplication/json'}
+            content = {'detail': 'content-type must be aplication/json'}
             return Response(content, status=status.HTTP_400_BAD_REQUEST)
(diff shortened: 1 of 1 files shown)


perma_web/api/views.py
```diff
@@ -692,6 +692,9 @@ def post(self, request, format=None):
         # mark batch with user
         if not request.user.is_authenticated:
             raise PermissionDenied()
+        if request.content_type != 'application/json':
```
sabzo, line 695: This could be a function that can be used elsewhere, but not sure what routes also need to have this error. All of them?

matteocargnelutti (commented):
Hey @sabzo 👋 !

LGTM, only thing I'd recommend revisiting a bit:
It seems like other routes return errors via a `detail` property, since it's Django Rest Framework's default (https://www.django-rest-framework.org/api-guide/exceptions/). 

I'd suggest either creating an exception class deriving from [DRF's `APIException()`](https://www.django-rest-framework.org/api-guide/exceptions/#apiexception), or more simply to return a JSON object with a `detail` key, since this is a custom route.

sabzo:
@matteocargnelutti I've made the change where the key for the error message is now "detail". Now merging!

Issue #3098, opened 2022-05-27
Review `is not None` check in admin filters


rebeccacremona:
See https://github.com/harvard-lil/h2o/pull/1552

Could be in some places we actually want it..... but I may have just copied and pasted boilerplate willy-nilly.

PR #3187, opened 2022-10-13
Track Internet Archive uploads in dedicated models

rebeccacremona:
### Context

### This PR

This PR adds models to Perma's codebase to help us keep track of the contents of Perma's IA collection better than we can with the [single field on `Link`](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/models.py#L1384) currently in use. 

It also adds Celery tasks to populate the database with information about the current content of the collection.

### A few details

`Link`s are associated with `InternetArchiveItem`s in a many-to-many relationship; you can make queries like `Link.objects.get(guid='AAAA-AAAA').internet_archive_items.all()` and `InternetArchiveItem.objects.get(identifier='daily_perma_cc_2020-10-02').links.all()`.

I added some basic Django admin views for the new models, and tried to make sure they will work well even with a few million rows. The case-insensitive "contains" [identifier filter](https://github.com/harvard-lil/perma/pull/3187/files#diff-47ac31691c3121c7694dd61abc250c309cbd706bc2a07b61223a56ca3e9ecc0cR143) on the `InternetArchiveItem` list view might prove too unwieldy; I'll keep an eye out.

I did not add any indexes to the db at this time. I'm not really sure what we might need.

If we run too many of these tasks in parallel, we might hammer our database. I tried to be gentle where possible, but this *is* creating a lot of rows. I thought especially about whether we might want to be careful when creating all the `InternetArchiveFile` objects for a daily item, but after weighing the options with @matteocargnelutti, decided to [go with the simplest option for now](https://github.com/harvard-lil/perma/pull/3187/files#diff-7d26136c89ca9052be7fa8d0f7a96b8fed12581e8338ef91078ec434eb8a06feR1849)... with the idea that we can see how it goes and adjust if necessary, by adding more resources, adding a `sleep` to the loop, switching to batched requests, etc.

### Testing

I didn't make any effort to add unit or integration tests for this code: it seemed like more work than it was worth, for these one-time tasks. I intend to exercise the new models in the future, when writing code that will actually update our Internet Archive collection.

I tested the individual link tasks locally by making a capture and [uploading it to the IA test_collection](https://archive.org/details/perma_cc_rcremonaBPM2-D424), and then triggering the tasks in the Django shell.

I tested the daily tasks locally by temporarily forcing `InternetArchiveFile` objects to be associated with a GUID of my choosing, present in my local database (namely, BPM2-D424, mentioned above) instead of the [actual GUID as found in from the filename](https://github.com/harvard-lil/perma/pull/3187/files#diff-7d26136c89ca9052be7fa8d0f7a96b8fed12581e8338ef91078ec434eb8a06feR1867), and then running the task against https://archive.org/download/daily_perma_cc_2021-10-01, a "daily" Item created for us by IA.

It seems fine, and I think that's about as thorough as we can get without a development database that more closely resembles production.

I recommend that we test further by running a small number of tasks synchronously in the production shell and inspecting the results.

### Deploying

The new Celery tasks will not be run automatically, so this should be safe to deploy whenever. I do not expect the migration to be slow, since it's just creating a couple new tables.

We can trigger the queuing tasks whenever we are ready in the production shell.

Tasks can be run any number of times against the same `Link`s/IA Items with no problem. If it turns out we want more control over queuing batches of tasks for particular `Link`s/IA Items, I can tweak the code as desired, but I don't think we'll need to.

I think that probably wants to happen only after we've removed the legacy tasks from the codebase: let's not fire up any IA workers until all that's gone, to completely remove any possibility that we are accidentally still updating the collection while these new tasks are running.


perma_web/perma/admin.py
```diff
@@ -134,6 +137,25 @@ def queryset(self, request, queryset):
...
+        if value is not None:
+            return queryset.filter(item_id=value)
+
+class IALinkIDFilter(InputFilter):
+    parameter_name = 'guid'
+    title = 'GUID'
+
+    def queryset(self, request, queryset):
+        value = self.value()
+        if value is not None:
```
rebeccacremona: Oh. I may be copy/pasting an existing [bug](https://github.com/harvard-lil/h2o/pull/1552)

perma_web/perma/tasks.py
```diff
@@ -1719,3 +1720,137 @@ def populate_warc_size(link_guid):
...
+    #     batch = list(islice(objs, batch_size))
+    #     if not batch:
+    #         break
+    #     InternetArchiveFile.objects.bulk_create(batch, batch_size)
+    #
+    # We are trying simple iteration first for improved observability and because get_or_create is repeatable
+    for ia_warc in files:
+        link_id = ia_warc.name.split('.')[0]
+        file, created = InternetArchiveFile.objects.get_or_create(
+            link_id='38W9-SUV9',
```
rebeccacremona: Committed with the test value.

rebeccacremona:
Oh, I also took care of https://github.com/harvard-lil/perma/issues/3098 while I was in there.

bensteinberg (approved):
This looks great. Let's try it out.

PR #3188, opened 2022-10-17
Remove legacy internet archive code

rebeccacremona:
### Context

### This PR

This PR removes the code we have been using to create new Items and to redact existing Items to "delete" them, when once-public Links were toggled to private.

### Deploying 

The legacy setting, `UPLOAD_TO_INTERNET_ARCHIVE` can be removed from [config](https://github.com/harvard-lil/salt/blob/develop/perma-settings/files/settings.py#L49), as can the stanza for [temporarily turning off the queueing of IA-related tasks](https://github.com/harvard-lil/salt/blob/2f48e411710dd07ecd296a7f9fa10cbd8f024cc9/perma-settings/files/settings.py#L166-L169)

The [script for uploading oversized warcs](https://github.com/harvard-lil/salt/blob/develop/perma-worker/files/ia_upload.py), [until recently run via cron](https://github.com/harvard-lil/salt/blob/2f48e411710dd07ecd296a7f9fa10cbd8f024cc9/perma-worker-prep.sls#L213-L222), relies on the to-be-deleted code; we may want to delete it as well. (We have not yet determined whether we'll need something similar going forward, or whether the new process will handle oversized warcs in a different way). I'm hanging on to the setting `INTERNET_ARCHIVE_MAX_UPLOAD_SIZE`, presuming that the new process will still use it one way or another.


lizadaly (approved):
🗑️

bensteinberg (approved):

PR #3191, opened 2022-10-20
Less expensive URL filter

rebeccacremona:
Make the API's URL filter do an exact match on `surt` instead of a fuzzy match on `submitted_url` so that queries over large link sets require less CPU. See [Slack](https://hlslil.slack.com/archives/C07URASMC/p1664542903409909) for context.


rebeccacremona:
Whoops, pushed the wrong branch, closing, and opening another.

PR #3192, opened 2022-10-20
Less expensive URL filter

rebeccacremona:
Make the API's URL filter do an exact match on surt instead of a fuzzy match on submitted_url so that queries over large link sets require less CPU. See [Slack](https://hlslil.slack.com/archives/C07URASMC/p1664542903409909) for context.

### DIFF
--- a/perma_web/api/views.py
+++ b/perma_web/api/views.py
@@ -12,6 +12,7 @@ from rest_framework.filters import SearchFilter, OrderingFilter
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.views import APIView
+import surt
 
 from perma.utils import stream_warc, stream_warc_if_permissible
 from perma.tasks import run_next_capture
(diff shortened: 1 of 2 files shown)


perma_web/api/views.py
```diff
@@ -279,11 +280,19 @@ class LinkFilter(django_filters.rest_framework.FilterSet):
...
+
     class Meta:
         model = Link
         fields = ['url', 'date', 'min_date', 'max_date']
 
+    def surt_filter(self, queryset, name, value):
+        try:
+            canonicalized = surt.surt(value)
+        except ValueError:
+            return queryset
```
lizadaly: What are the conditions that might throw a `ValueError` here? Would it be dangerous/non-secure in some way to return an unfiltered queryset? Should it at least log a warning?

perma_web/api/views.py
```diff
@@ -279,11 +280,19 @@ class LinkFilter(django_filters.rest_framework.FilterSet):
...
     min_date = django_filters.IsoDateTimeFilter(field_name="creation_timestamp", lookup_expr='gte')   # ?min_date=
     max_date = django_filters.IsoDateTimeFilter(field_name="creation_timestamp", lookup_expr='lte')   # ?max_date=
-    url = django_filters.CharFilter(field_name="submitted_url", lookup_expr='icontains')              # ?url=
+    url = django_filters.CharFilter(method='surt_filter')                                             # ?url=
+
     class Meta:
         model = Link
         fields = ['url', 'date', 'min_date', 'max_date']
 
+    def surt_filter(self, queryset, name, value):
```
lizadaly: Rename `name` to `_` to hint that it's intentionally unused? (This is totally up to your personal style.)

lizadaly (approved):
The tests look like they do legitimately need updating, but otherwise looks good! Nice work on the query performance improvement!

perma_web/api/views.py
rebeccacremona: Ah, thanks, I will add a comment. `ValueError` is thrown when you pass something that can't be parsed into a surt, say, "sajhkdfkjasdhfjadkhfhas". I think it's perfectly fine to return a non-filtered queryset here: it would be just like they visited /api/v1/archives instead of /api/v1/archives?url=asljdhfladskjfhaldshf

perma_web/api/views.py
rebeccacremona: Sure, can do; I copied those var names straight from the docs and didn't think about it.

rebeccacremona:
Updated! Thanks for the review. I think the tests will pass now.

Issue #3193, opened 2022-10-21
`cached_path` not populated for new organizations' folders


rebeccacremona:
I noticed locally that the Folder serializer was making an unexpected number of database queries. 

It turns out to be [this line](https://github.com/harvard-lil/perma/blob/develop/perma_web/api/serializers.py#L71), which serializes the folder's `cached_path`. 

I appear to have left this code unfinished: that serializer method was only [intended to be temporary](https://github.com/harvard-lil/perma/blob/develop/perma_web/api/serializers.py#L71). The plan was to [set `cached_path`](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/models.py#L1242) on the creation of every new folder and [populate the field for already existing folders](https://github.com/harvard-lil/perma/blob/1400671951e804a4bd20e5c6476c0ba83b2c2e15/perma_web/fabfile/dev.py#L675) via a task, and then remove this db-hitting method. 

But, we never removed the temporary method, even after populating the field for existing folders. And that turns out to have covered for a bug in the code for creating new organizations and their top-level shared folder. [This stanza](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/models.py#L716-L719) instantiates the folder and saves it: `cached_path` gets populated in the DB. But, we don't `refresh_from_db` before associating the folder object with the new organization object and calling `save`... which results in an update to the folder with `cached_path=None`. 
 
There are 350-ish such folders.
```
In [1]: from perma.models import Folder

In [2]: Folder.objects.filter(cached_path__isnull=True).count()
Out[2]: 358
```

Let's fix this, then run the task to populate `cached_path` for these org folders, and then remove the intended-to-be-temporary serializer code.

rebeccacremona:
`fab dev.populate_folder_cached_path` ran in a few seconds, now:
```
In [2]: Folder.objects.filter(cached_path__isnull=True).count()
Out[2]: 0
```

PR #3194, opened 2022-10-21
Fix `cached_path` bug for new org folders

rebeccacremona:
See https://github.com/harvard-lil/perma/issues/3193 for details.

Without this, when new orgs are created, their shared folder doesn't get its "path" (location in the folder tree) cached on the model, so every time we need to know it, we have to re-calculate it, which results in a lot of unnecessary database queries.

After this fix is deployed, we'll want to take care of the folders that fell through the cracks.


matteocargnelutti (approved):
Oh yeah 🤘 🎸 !

PR #3196, opened 2022-10-24
Folder's `cached_path` field is now mandatory

rebeccacremona:
In 2021, we started [caching each folder's "path" ](https://github.com/harvard-lil/perma/commit/686176f559a12d41fedff28cae6b7161d9f38c65) (location in the folder tree) on the model for performance reasons. We noticed last week that a) there was a bug in that code and b) I never finished the project: see https://github.com/harvard-lil/perma/issues/3193.

This PR finishes up the work: `cached_path` is now mandatory.

Production and stage have already been prepared for this migration; developers should run `fab dev.populate_folder_cached_path` before migrating. I thought about adding a data migration to make that unnecessary, but decided not to since [instance methods aren't available inside migrations](https://docs.djangoproject.com/en/4.1/topics/migrations/): we'd have to copy over all the code for Django MPTT's `get_ancestors` method, which is [extensive](https://github.com/django-mptt/django-mptt/blob/62b8892a3dd5cd70c299af9144e79674b4f71c11/mptt/managers.py#L76).


bensteinberg (approved):
:+1:

Issue #1470, opened 2016-04-20
Let a user indicate that the image capture is preferred to the web capture


jcushman:

jcushman:
Next steps: sketch a couple of options. Radio button in tray, or offer to link owner to change the default when they view a non-default view, or ... ?

kilbergr:
When you visit https://perma.cc/F857-95FJ, Perma serves up the high-fidelity playback. You can direct people to a screenshot by adding the query string https://perma.cc/F857-95FJ?type=image or advising them to click.
People have asked for the ability to check a box or something that would cause Perma to redirect to the screenshot view (or similar) if people visited https://perma.cc/F857-95FJ.
Would involve adding a field to track, adding redirect behavior, and arranging things such that a person could still click to get to the "standard" view, for such links.
(special cases: when there is no screenshot; when there is only a screenshot (are there any like this?); when a user replaces our capture with a PDF upload after the fact; when a capture is in process; when a capture failed; when other redirect logic is in play, like a replacement link.....)

kilbergr:
Three options (note the blue and typeface _would_ match):
<img width="1728" alt="Screen Shot 2022-09-29 at 1 23 02 PM" src="https://user-images.githubusercontent.com/7401870/193103464-22e37120-23e3-45dd-9e3c-81ca7179686c.png">
<img width="1522" alt="Opt2" src="https://user-images.githubusercontent.com/7401870/193104111-99ebcae9-f4e5-40a0-8bd8-67d05e864d09.png">
<img width="1522" alt="Opt3" src="https://user-images.githubusercontent.com/7401870/193107304-0049d8b6-1239-48d6-b976-1993dbdddd6b.png">
I'm not sure what you'd like done with the View mode selection on the header--should that disappear in this view?
AKA:
<img width="1522" alt="Opt3_noheader" src="https://user-images.githubusercontent.com/7401870/193109073-2c1978e5-e7fb-4821-999f-60c76224d13d.png">

matteocargnelutti:
The radio toggle makes a lot of sense to me. It helps indirectly explain what this feature is about.

kilbergr:
Tagging @rebeccacremona and @clare-stanton for review, but have also posted in Slack.

kilbergr:
Takeaway from the group is: go with the radio buttons and disappear the View mode indicator in the header. ALSO, this drop down tray should be redesigned as a modal. That work may be started in the course of this ticket but hopefully will be done after this ticket instead.

[...]

PR #3197, opened 2022-10-25
Let a user indicate that the image capture is preferred to the web capture

kilbergr:
Ok so! This PR addresses [this open issue](https://github.com/harvard-lil/perma/issues/1470) to "Let a user indicate that the image capture is preferred to the web capture".

The PR adds a boolean field to the `perma_link` object called `screenshot_view`. When it's true, the URL (i.e. https://perma.cc/5WFL-AJ27) without any type specified defaults to displaying a screenshot view if the screenshot is present. If the screenshot is not present, the capture will be shown. When `screenshot_view` field is false, the opposite is true.

When a user specifies image type, (i.e. https://perma.cc/5WFL-AJ27?type=image), the screenshot, if present, will be shown regardless of `screenshot_view` value. This PR introduces a `standard` type (i.e. https://perma.cc/5WFL-AJ27?type=standard) for user to specify viewing the capture instead of the screenshot.

A minor aesthetic change: when the `Show record details` dropdown is displayed, the `Screenshot` `Standard` View Mode options are no longer visible.


bensteinberg (approved):
LGTM, but I have added @rebeccacremona as a reviewer and will defer to her.

rebeccacremona (changes requested):
We chatted about a few suggested changes offline; will re-review when ready!

kilbergr:
Hoorah passing and all comments addressed ! @rebeccacremona ready for a fresh look when you have the chance.

rebeccacremona (approved):
This looks great 🎉 !!!

rebeccacremona:
I saw a tiny rendering problem with the expandable link list on Firefox and Chrome that didn't show up in Safari (😅), so took the liberty of pushing up a couple of lines of css for that:

image

While messing with that, happened to notice that the two radio buttons there had the same ID's, so clicking on the "screenshot" label filled in the "standard" button, so just went ahead and updated that as well.

Congrats, @kilbergr at landing a feature Perma users have been wanting forever! Folks are going to be so jazzed. Can't wait until this is publicized.

Issue #3198, opened 2022-10-27
Use `iterator()` when iterating through large querysets in Celery tasks


rebeccacremona:
Back when, I found that `iterator()` when used with `values_list` was causing querysets to be evaluated twice: we saw the (at the time expensive and un-optimized) queries running on the database twice. So, we [removed `iterator()`](https://github.com/harvard-lil/perma/commit/3620c87fa7ba19e5a8038561cfb29a1725a92982).

I can no longer reproduce that problem. Lots of things have changed in the meantime: Django upgrades, a migration from MySQL to Postgres, etc. 

Let's put `iterator()` back and thereby use RAM more gently.

rebeccacremona:
((I went so far as to spin up a `mysql` container, and wire it up to `web`, install fixtures and make a few dozen captures, and try to reproduce there: even there, I was not able to reproduce. Just some mild gaslighting underway...  I'm going to move on, and not worry about it too much...))

rebeccacremona:
Whoops, this is done already. https://github.com/harvard-lil/perma/pull/3202/commits/876f53ffb4c6eabdbc427dc4c831d3855419eb7e

PR #3202, opened 2022-10-28
Populate the DB with info about Perma's IA Collection, part 2

rebeccacremona:
### Context

### The Story

We had a go at creating `InternetArchiveItem` and `InternetArchiveFile` objects to track what exists in Perma's Internet Archive collection, and had some success. We got through the entire set of tasks for [daily Items](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/tasks.py#L1613) with no apparent problems, but ran into some difficulty with the tasks for [individual Items](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/tasks.py#L1544):

- the enqueuing task used up too much RAM and crashed after processing ~1.6 million records, two tries in a row

- we realized that we should be creating `InternetArchiveFile` objects in all cases, not just when we find a Perma Link's warc is indeed currently attached to an Item on IA: we want the foreign key relationships to exist in our DB for all extant IA Items regardless, and, going forward, we want to track what's up with a file (uploaded successfully? uploaded half-way before the process crashed out? uploaded and then deleted? etc.) in more detail

- we learned about the existence of a previously unknown variety of IA Items: "dark" items, that exist, but have no metadata associated with them (at least, any metadata that we can access using our credentials). 

### This PR

This PR gets us ready for another attempt at populating our DB.

1) The enqueuing task should no longer use up too much RAM, because we are now using `iterator` to page results without any internal caching. (We were not using `iterator` before because I thought it was actually problematic when paired with `values_list`, but that turns out to [no longer be the case](https://github.com/harvard-lil/perma/issues/3198).

2) The enqueuing task no longer queues tasks for all candidate links, and instead limits itself to those we haven't processed yet.

3) The task can handle dark IA items and creates an `InternetArchiveFile` object with the appropriate status even when a warc is not present in IA.

### Order of Operations and Deploying

We've learned to be cautious about migrations on tables with millions of rows. I've accidentally caused hour+ outages on Perma before by making alterations to `Link` that I expected to be quick, but were not.

This PR does several things that I think have the potential to be slow.

- The [creation of the indexes on `InternetArchiveItem.span`](https://github.com/harvard-lil/perma/pull/3202/files#diff-4ba0dda553321bb4ffcb8686cef0b043b6258ea0d8d97161a9595e60c4a98392)? I honestly don't know. I thought we might want to create these indexes outside of the deployment process. Since I was already writing the SQL for the index on `isempty`, I decided to [`CREATE INDEX IF NOT EXISTS "empty_span_idx"`](https://github.com/harvard-lil/perma/pull/3202/files#diff-4ba0dda553321bb4ffcb8686cef0b043b6258ea0d8d97161a9595e60c4a98392R19) so that the migration is happy even if the index is present. I'm wondering whether we might want to do something similar for the [GiST index](https://github.com/harvard-lil/perma/pull/3202/files#diff-4ba0dda553321bb4ffcb8686cef0b043b6258ea0d8d97161a9595e60c4a98392R16) too. It's a little silly, but I was thinking we could run `CREATE INDEX "perma_inter_span_bafa26_gist" ON "perma_internetarchiveitem" USING gist ("span");` manually, and see if it was slow. If it's not slow, we can just drop it and let the migration do it again during deployment. If it IS slow, then I can rewrite this migration to use a [`RunSQL with state_operations`](https://docs.djangoproject.com/en/4.1/ref/migration-operations/#runsql) that's the equivalent, but with `IF NOT EXISTS`. Does that sound right? Other ideas? Am I worrying needlessly?

- The creation of the new `cached_is_dark` field on `InternetArchiveItem`, with a default of `False`. I remember doing something like that once before with a field on `Link`, and it being unbelievably slow to deploy... but it was back when we were on an old MySQL. Hopefully not a problem here? If we think it is, we could make the field nullable here, populate it out-of-band, and then switch to non-nullable later.

- Hopefully not the creation of the nullable field `status` on `InternetArchiveFile`, which I plan to populate out-of-band via celery tasks ([queuing](https://github.com/harvard-lil/perma/pull/3202/files#diff-7d26136c89ca9052be7fa8d0f7a96b8fed12581e8338ef91078ec434eb8a06feR1713), [doing](https://github.com/harvard-lil/perma/pull/3202/files#diff-7d26136c89ca9052be7fa8d0f7a96b8fed12581e8338ef91078ec434eb8a06feR1737))... but 🤷‍♀️ 

Any thoughts are welcome.

If we decide to try this as is, I think the plan would be:

- create the `isempty` index

- try creating the GiST index and see if it's slow; drop if fast; revise this PR before merging if slow

- merge and deploy

- make sure some IA workers are available

- run the status population enqueuing task from the shell:
```
In [1]: from perma.models import InternetArchiveFile

In [2]: from perma.tasks import queue_batched_tasks, populate_internet_archive_file_status

In [3]: queue_batched_tasks(
   ...:     populate_internet_archive_file_status,
   ...:     InternetArchiveFile.objects.filter(status__isnull=True)
   ...: )
```

- run the individual item enqueuing task from the shell:
```
In [4]: from perma.tasks import queue_backfill_of_individual_link_internet_archive_objects

In [5]: queue_backfill_of_individual_link_internet_archive_objects.delay()
```


rebeccacremona:
Thanks to @jcushman for showing me the pattern for [batched celery tasks](https://github.com/harvard-lil/perma/pull/3202/files#diff-7d26136c89ca9052be7fa8d0f7a96b8fed12581e8338ef91078ec434eb8a06feR1713) and for the idea of adding `status` to `InternetArchiveFile`, and thanks to the entire team for help and sage advice about indexes and db performance.

perma_web/perma/tasks.py
```diff
@@ -1695,3 +1708,38 @@ def backfill_daily_internet_archive_objects(date_string):
...
+
+
+@shared_task(acks_late=True)
+def populate_internet_archive_file_status(pks):
+    """
+    We created a large number of InternetArchiveFile objects before realizing we
+    needed a 'status' field. This one-time task populates that field for the existing
+    items in a database-friendly way.
+    """
+    count_updated = InternetArchiveFile.objects.filter(pk__in=pks).update(status='confirmed_present')
```
matteocargnelutti: **Question:** Could the filter be on both `pk` and `status is null` ? 

If that were the case this task:
- Could be interrupted / resumed easily _(it'll only run on the subset that wasn't processed before it was interrupted)_, in case it ends up being slow or crashing. 
- Can't accidentally override `status` at a later point in time if already set to a different value

rebeccacremona: Good idea! Will do!

matteocargnelutti (approved):
This is awesome, @rebeccacremona 🤘 🎸 .
Thanks for the detailed PR notes 😄 , really helped me with this review.

Will follow up on Slack re _"how long will the creation of fields / indexes take"_.

rebeccacremona:
All the things that I was worried would be slow turn out to be plenty fast, with modern postgres: we checked on a replica of prod.
```
perma=# CREATE INDEX IF NOT EXISTS "empty_span_idx" ON "perma_internetarchiveitem" ("identifier", (isempty("span")));
CREATE INDEX
Time: 3287.045 ms (00:03.287)

perma=# CREATE INDEX "perma_inter_span_bafa26_gist" ON "perma_internetarchiveitem" USING gist ("span");
CREATE INDEX
Time: 1705.731 ms (00:01.706)

perma=# ALTER TABLE "perma_internetarchivefile" ADD COLUMN "status" varchar(19) NULL;
ALTER TABLE
Time: 16.520 ms

perma=# ALTER TABLE "perma_internetarchiveitem" ADD COLUMN "cached_is_dark" boolean DEFAULT false NOT NULL;
ALTER TABLE
Time: 17.611 ms
```

Thank you, postgres 11, for letting us [add fields with non-null defaults without rewriting the table](https://www.depesz.com/2018/04/04/waiting-for-postgresql-11-fast-alter-table-add-column-with-a-non-null-default/)!

PR #3205, opened 2022-11-02
Load queryset using custom SQL that is faster.

rebeccacremona:
We tried running this draft of [task to make individual InternetArchiveItem objects](https://github.com/harvard-lil/perma/blob/v0.374/perma_web/perma/tasks.py#L1544) for the `Links` we didn't get to the first time (because the enqueuing task used up too much RAM and crashed after processing ~1.6 million records).

It failed, because the initial database query is impossibly slow.

I did not find a way to make the ORM's query efficient, but I did find equivalent SQL that runs fast locally against a dump of our production data. This PR updates the task to use that raw SQL.

There may be a way to force the Django ORM to return this SQL; I'm not sure. (Sounds like a fun riddle....). Since we don't plan to run this task often (optimally, just once), I'm not going to worry about it too much.

There might also be a better way to get a queryset from raw sql than using `Link.objects.filter(guid__in=RawSQL(sql, []))`. I messed around with `Link.objects.raw`, but couldn't seem to get it working... Again, since this seems fine locally against the dump of production data, I'm not going to worry too much about it.

But, if people have ideas, do let me know 🙂.

### DIFF
--- a/perma_web/perma/tasks.py
+++ b/perma_web/perma/tasks.py
@@ -1548,13 +1548,46 @@ def queue_backfill_of_individual_link_internet_archive_objects(limit=None):
     via dedicated models, rather than using fields on the Link object. This task queues up the
     creation of IA Django model objects for our legacy individual Items.
     """
-    # get all Links we think may have been uploaded to IA,
-    # and then filter out the ones we have already processed
-    links = Link.objects.exclude(
-        internet_archive_upload_status='not_started'
-    ).exclude(
-        internet_archive_items__span__isempty=True
-    )
+    # Get all Links we think may have been uploaded to IA,
+    # and then filter out the ones we have already processed.
+    # Do so with our own SQL, because that generated from the
+    # intuitive ORM query proved to be impossibly slow.
+    # links = Link.objects.all_with_deleted(
+    #     internet_archive_upload_status='not_started'
+    # ).exclude(
+    #     internet_archive_items__span__isempty=True
+    # )
+    from django.db.models.expressions import RawSQL  # noqa
+
+    sql = '''
+        SELECT
+          "perma_link"."guid"
+        FROM
+          perma_link
+          LEFT JOIN (
+            SELECT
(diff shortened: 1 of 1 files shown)


bensteinberg (approved):
Wow nice!

lizadaly:
Nice work!

I don't think you should make any changes, but some modern SQL style kind of things for future raw SQL adventures:

I'd drop all the double quoted table and column identifiers. Django was just being safe when it generated those, but it reduces readability. Django won't have generated any identifiers that actually require quoting, but it can come up with legacy databases.

Nested selects like your left join can usually be written as a common table expression (CTE) that read more like functions:

```sql
WITH links_with_individual_items AS ( query )

SELECT perma_link.guid
FROM
    perma_link
LEFT JOIN links_with_individual_items ON
    perma_link.guid = links_with_individual_items.guid
...
```

which you can think of as a lazy evaluation of the CTE.

I'm not sure without trying whether:

```sql
AND NOT (perma_link.internet_archive_upload_status = 'not_started')
```

is equivalent to

```sql
AND perma_link.internet_archive_upload_status != 'not_started'
```

but if it is I usually find that construction easier to reason about.

Totally doesn't matter, but you can use multiple inner join columns in a single join statement:

```sql
INNER JOIN "perma_internetarchivefile" ON (
  "perma_link"."guid" = "perma_internetarchivefile"."link_id"
)
INNER JOIN "perma_internetarchiveitem" ON (
  "perma_internetarchivefile"."item_id" = "perma_internetarchiveitem"."identifier"
)
```

could be

```sql
INNER JOIN perma_internetarchivefile ON 
   perma_link.guid = perma_internetarchivefile.link_id AND
   perma_internetarchivefile.item_id = perma_internetarchiveitem.identifier
)
```

but again everything here is probably irrelevant to the query planner and is just about human preference.

rebeccacremona:
@lizadaly thank you so much! this is super good to know

PR #3207, opened 2022-11-03
Don't track .dump

rebeccacremona:
We've been experimenting with running against production-like data locally. That has involved passing around and ingesting large `*.dump` files with `pg_restore`. 

For convenience, this PR adds `*.dump` to .gitignore, so we don't have to be careful about accidentally committing them to the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,8 @@ ghostdriver.log
 
 *.todo
 
+*.dump
+
 # node
 node_modules
 npm-debug.log


lizadaly:
Do you want both `*.dump` and `*.dmp`?

bensteinberg:
There is no `dmp`

lizadaly (approved):
Dump `.dump`, got it.

rebeccacremona:
Weeee, off to a great start, today 😅.

Switched to `.dump` everywhere.

PR #3208, opened 2022-11-04
Suppres logging and fix stored archive content types

mcharters:
After updating to a more recent version of perma I found our created archives were no longer playing back - after looking at changes to `S3MediaStorage` it looked like `AzureMediaStorage` needed some tweaks to modify archive content types and encodings too.

I also removed some HTTP logging from the server output.


matteocargnelutti (approved):
LGTM 👍 ! Thanks.

PR #3210, opened 2022-11-10
Add Link/File-level metadata to "daily" digest IA items.

rebeccacremona:
### Context

### This PR

2949 "daily" digest Items were created on Perma.cc's behalf by the Internet Archive team, who used their own scripts to produce them from the "individual" Items we've created ourselves over the past years. These daily Items do not presently have any metadata about the Perma archives they contain: they have only the creation date (not timestamp), GUID, and WARC file.

This PR includes tasks for adding details like the target URL and the capture timestamp to the Item's file-level metadata, which is a stored in a generated "\<identifier\>_files.xml" file associated with the Item. (See https://archive.org/developers/md-write.html#targets.) The best I can tell, that metadata is not searchable and is not presently displayed in the IA GUI... but we think it ought to be present regardless.

### Deploying

These tasks do two things we haven't tried before:

- using `ia_session.get_tasks_api_rate_limit` and `ia_session.s3_is_overloaded` to see if we are approaching any rate limits before issuing scheduling requests to IA;

- re-queueing tasks that fail or are skipped due to rate-limiting automatically.

We'll have to see how that works out! I did my best to predict how the IA API will respond in all situations, but I may have missed some cases: the docs don't provide an exhaustive set of possible return values.

If my handling of the rate-limiting proves insufficient, or if we see tasks re-queuing themselves ad infinitum due to unhandled errors or the like, let's just clear the IA celery queue and start fresh after I push the necessary fixes: we should be able to pick up where we left off.

After this code is merged, I propose to run a single task via the Django shell, and inspect the results in IA and the Django admin after everything is done:
```
from perma.models import InternetArchiveFile
from perma.tasks import add_metadata_to_existing_daily_item_files

add_metadata_to_existing_daily_item_files([991])
```

[991](https://perma.cc/admin/perma/internetarchivefile/991/change/) is one of the two I've been experimenting with locally): you can see dev-style metadata for it (and another link) at this moment in https://ia804600.us.archive.org/7/items/daily_perma_cc_2021-10-01/daily_perma_cc_2021-10-01_files.xml

image

...that should get updated to prod-style metadata when we run this task.

Presuming that looks good, I propose to then run a test batch of 1,500: our rate limit for `modify_xml.php` tasks is 1000, I think, so with any luck, this will give us a change to either gently test out skipping/re-scheduling code, or see if, under current conditions and with our current concurrency settings, we seem to be scheduling work at a manageable clip.

```
from perma.models import InternetArchiveFile
from perma.tasks import queue_batched_tasks, add_metadata_to_existing_daily_item_files

files = InternetArchiveFile.objects.filter(
    item__span__isempty=False,
    status='confirmed_present',
    cached_submitted_url__isnull=True
)[:1500]
queue_batched_tasks(add_metadata_to_existing_daily_item_files, files, batch_size=100)
```

I think we should plan to keep the concurrency low: these tasks will be unusually quick for us to schedule, but will still take time for them to process.

(I had hoped to schedule a smaller number of batch updates using their [multi-write](https://archive.org/developers/md-write.html#multi-target-writes) API, but alas, there is no wrapper for it in the `internetarchive` python package, and I don't want to try and do this one thing without the package.)

### DIFF
--- a/perma_web/perma/models.py
+++ b/perma_web/perma/models.py
@@ -33,6 +33,8 @@ from django.db.models import Q, Max, Count
 from django.db.models.functions import Now
 from django.db.models.query import QuerySet
 from django.contrib.postgres.indexes import GistIndex
+from django.template.defaultfilters import truncatechars
+from django.urls import reverse
 from django.utils import timezone
 from django.utils.functional import cached_property
 from django.views.decorators.debug import sensitive_variables
@@ -47,7 +49,8 @@ from .utils import (Sec1TLSAdapter, tz_datetime,
     prep_for_perma_payments, process_perma_payments_transmission,
     pp_date_from_post,
     first_day_of_next_month, today_next_year, preserve_perma_warc,
-    write_resource_record_from_asset, user_agent_for_domain)
+    write_resource_record_from_asset, user_agent_for_domain,
+    protocol, remove_control_characters)
 
 
 logger = logging.getLogger(__name__)
@@ -2104,6 +2107,8 @@ class InternetArchiveFile(models.Model):
     cached_submitted_url = models.TextField(null=True, blank=True, default=None)
     cached_perma_url = models.TextField(null=True, blank=True, default=None)
 
+    tracker = FieldTracker()
+
     class Meta:
         verbose_name = "Internet Archive File"
         unique_together = (("link", "item"),)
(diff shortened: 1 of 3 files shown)


bensteinberg (approved):

matteocargnelutti (approved):
Still waiting for tests to go through but: LGTM 🤘  ! 
Plan for running this on prod makes a lot of sense to me as well.  
Appreciate the detailed notes / comments as always.

PR #3211, opened 2022-11-10
Add Link/File-level metadata to "daily" digest IA items, Part II

rebeccacremona:
Our test run of https://github.com/harvard-lil/perma/pull/3210 went great! But, we found a few items WERE indeed re-queuing themselves ad infinitum....

This PR fixes the cause for (at least most of) that looping: it turns out IA was stripping repeated whitespace from the "title" field, and we didn't know that...

And, it stops things from re-queuing forever, instead keeping track of how many times we've repeated for a given file, and sending us an error email if we reach the limits set in `settings`.

By default, I'm not setting a retry limit for rate-limiting errors: we have yet to see what that looks like, and whether we'd actually prefer the queue to babysit itself indefinitely.

I set the error rate limit to a round dozen, for no particular reason.


rebeccacremona:
Actually, I should rename these settings, hold one.

perma_web/perma/tasks.py
```diff
@@ -1902,12 +1925,21 @@ def confirm_added_metadata_to_existing_daily_item_files(file_ids):
...
         expected_metadata = InternetArchiveFile.standard_metadata_for_link(link)
         try:
-            assert expected_metadata.items() <= ia_file.metadata.items()
+            assert expected_metadata.items() <= ia_file.metadata.items(), f"{expected_metadata.items()} != {ia_file.metadata.items()}"
         except AssertionError:
             # modify_xml tasks can take some time to complete;
             # the task for this link appears not to have finished yet.
             # we need to check again later.
-            file_ids_to_check_again.append(file_id)
+            print(previous_attempts)
```
bensteinberg: Do you want to leave the print statement here?

rebeccacremona: Nope. Thanks!

perma_web/perma/tasks.py
```diff
@@ -1871,12 +1887,19 @@ def add_metadata_to_existing_daily_item_files(file_ids):
...
+            for file_id in attempts_dict:
+                attempts_dict[str(file_id)] = attempts_dict[str(file_id)] + previous_attempts[str(file_id)]
+
+        add_metadata_to_existing_daily_item_files.delay(file_ids_to_retry, attempts_dict)
         logger.info(f"Re-queued 'add_metadata_to_existing_daily_item_files' for {len(file_ids_to_retry)} InternetArchiveFiles.")
 
 
-@shared_task(acks_late=True)
-def confirm_added_metadata_to_existing_daily_item_files(file_ids):
+@shared_task(acks_late=True, previous_attempts=None)
```
rebeccacremona: This is a mistake too. Sloppy, sorry!

bensteinberg (approved):
Looks great

perma_web/perma/tasks.py
```diff
@@ -1837,7 +1837,15 @@ def add_metadata_to_existing_daily_item_files(file_ids):
             rate_limit_approaching = True
         if s3_is_overloaded or rate_limit_approaching:
             logger.warning(f"Skipped add metadata task for {file_id} (IA Item {ia_item.identifier}, File {link.guid}) due to rate limit.")
-            file_ids_to_retry.append(file_id)
+            retry = (
+                not settings.INTERNET_ARCHIVE_RETRY_FOR_RATELIMITING_LIMIT or
+                not previous_attempts or
+                (settings.INTERNET_ARCHIVE_RETRY_FOR_RATELIMITING_LIMIT > previous_attempts[file_id] + 1)
```
bensteinberg: Does this need to index `previous_attempts` with `str(file_id)` like you do elsewhere?

rebeccacremona: Gulp: yes. Thanks for the careful proofreading. I clearly fired this off too quickly.

bensteinberg: np -- I was just trying to figure out why you weren't using the raw IDs as dictionary keys. I suppose that integer dict keys are probably a code smell.

bensteinberg: Ah you explained it in the commit message: "Dict key is a string (required by celery task format)"

lizadaly: mypy's also helpful here for catching those sorts of things, assuming you defined the dict as having keys of type string.

rebeccacremona: Yeah! I was passing in the integers initially, and couldn't figure out why I kept getting key errors... Hence that print statement you found, actually. It turned out that the keys of `previous_attempts` were strings, despite my having passed in ints.

I wasn't sure if it was better to explicitly convert it to a string once, and then use that var everywhere, or cast each time... My thought was, I was equally likely to use the wrong var as I was to forget to cast, so six of one half dozen of another...

Kind of a bummer the ints don't stick around, I'm guessing it's a `pickle` thing.

PR #3213, opened 2022-11-14
Add Link/File-level metadata to "daily" digest IA items, Part III

rebeccacremona:
### The Story

We let our current [draft of the task](https://github.com/harvard-lil/perma/pull/3211) to add Perma Link/file-level metadata to "daily" digest IA items run over the weekend. It did decently, but ran into some difficulties:

- API requests ended up taking much longer than in our test runs, so virtually every batch of 100 timed out at some point, midway

- While the rate-limiting code seems to have worked (we didn't receive any emails indicating we hit the limit, or any tasks failed to be accepted for scheduling), we did receive error messages stating that we sometimes failed to obtain a necessary lock on the Item we wanted to update.

- We realized that zero `confirm_added_metadata_to_existing_daily_item_files` would run until every initially-queued `add_metadata_to_existing_daily_item_files` task had finished... which made this process less easy to stop and resume, and less easy to inspect.

So, we purged the queue and stopped editing metadata, and instead queued up `confirm_added_metadata_to_existing_daily_item_files` for all the files we processed over the weekend.

```
possibly_done = InternetArchiveFile.objects.filter(
    id__lt=261600,
    item__span__isempty=False,
    status='confirmed_present',
    cached_submitted_url__isnull=True
)
```

That's been running all day with success, but also some problems:

- at our desired concurrency rate, the IA API started timing out on all our requests

- at a reduced concurrency rate, things are succeeding, but we can see in the logs that the IA library's built-in retry mechanism is kicking in a lot: a lot of requests are still timing out

### This PR

This PR adapts the tasks to what we've learned:

- If Celery's soft time limit is hit, it re-queues any files it didn't get to before the limit. (We'll also plan to run in batches smaller than 100, going forward).

- If an API requests time out, it re-queues the file instead of erroring out and emailing us.

- If the Internet Archive API reports it couldn't get a necessary write lock, it just re-queues the file and prints a WARN-level message to the log instead of re-queuing it and emailing us.

- It makes fewer Internet Archive API requests: before, we loaded the IA item's metadata fresh for every file. Now, we cache that, reusing the same object to schedule requests when possible. If a single task is processing files associated with many different IA items, this will increase RAM usage, but we don't anticipate ever setting up tasks that way: the files in a batch generally belong to at most two or three items. (If it becomes a problem, we can add code to `del` the cache if it gets too large.)

- It sets the tasks to run on two separate queues: `ia` if task attempts to alter something on Internet Archive, `ia-readonly` if the task only reads and reacts to information from IA. This setup will let us use different concurrency settings for different kinds of tasks (hopefully allowing us to churn through less-expensive read-only tasks faster), and will let us run both `add..` and `confirm...` tasks at the same time, making it easier to track what has been completed and what hasn't, and making it easier to stop/resume when necessary.

The diff looks more extensive than it really is, because of the new wrapping `try/except SoftTimeLimitExceeded`.

### Deploying and order of operations

@bensteinberg has already set things up so our celery workers will listen to both queues 🙌 

We could consider purging the queue when this lands and re-queuing up the `confirm....` tasks for files processed over the weekend: this may be a lot faster, if skipping the routinely slow/often-timing-out `ia_session.get_item` lets us get on with things and doesn't just result in a different timeout later in the code. Or we could let what's running right now finish. Either way.


bensteinberg (approved):

rebeccacremona:
I also threw in a settings change: let's not retry 12 times just yet: 2 is plenty. Let's up again once we know this is all working better.

PR #3215, opened 2022-11-16
Spruce up admin stats page a bit

rebeccacremona:
Our internal, admin-only stats page has been assembled on-the-fly as we've needed things, and we haven't put a lot of effort into its styles. The status quo is relatively hard to work with: its content jumps around as it loads, as well as every two seconds when it updates.

This PR makes a few adjustments, nothing spectacular, intended to improve usability.

- It splits the page into a number of tabs (thanks Bootstrap!), so that sections' changes in height don't affect each other;

- it adds css to stop content from bleeding into other content: I think the lists of currently-running celery tasks will be much easier to read this way;

- it adds a button that lets you toggle the auto-refreshing of the list of running celery tasks on and off at will, so that you can read without the data changing;

- and a few small css tweaks.

It also adds information on where we stand with Internet Archive rate limits, which I've been watching pretty closely as we test out our new tasks.

We could do a lot more with this page: data probably shouldn't refresh in the background if a given tab isn't active; we probably could do much better with web components and modern JS's support for reactivity instead of handlebars templates that clobber entire divs... but I think this small update will still be worth it.


rebeccacremona:
Screen Shot 2022-11-16 at 11 46 07 AM
Screen Shot 2022-11-16 at 11 43 18 AM
Screen Shot 2022-11-16 at 11 46 12 AM
Screen Shot 2022-11-16 at 11 43 25 AM
Screen Shot 2022-11-16 at 11 43 29 AM

bensteinberg (approved):
🤘

PR #3218, opened 2022-11-16
Add Link/File-level metadata to "daily" digest IA items, Part V

rebeccacremona:
Just a few more tweaks, after observing a day's worth of tasks run with [the last PR](https://github.com/harvard-lil/perma/pull/3216)'s code.

- We want to be able to control whether we get error emails, should tasks retry as many times as they are allowed but still fail: if IA suddenly goes down overnight, we don't want a cavalcade.

- Some logging messages were a little unclear, and we wanted more info after SoftTimeLimits were exceeded.

- We wanted more feedback, after hitting the "refresh" button on the rate limits tab of the admin stats page.

- A few CSS/rendering fixes.

After this, we plan to up the order of magnitude of tasks we are queuing by another factor.


bensteinberg (approved):

rebeccacremona:
The status indicator:

https://user-images.githubusercontent.com/11020492/202299840-eb7597d5-a2c9-4d42-b965-487e6bcefc98.mov

PR #3222, opened 2022-11-18
Add Link/File-level metadata to "daily" digest IA items, Part VIII

rebeccacremona:
We processed ~100K InternetArchiveFiles in about 10hrs with the [latest version of the task](https://github.com/harvard-lil/perma/pull/3221) to add file/Perma Link-level metadata to the existing "daily" digest Internet Archive items, and we're pretty happy with the way it ran!

This PR makes a few tweaks to logging. 

After this, we are hoping to run the whole remaining set, which will take several days, but hopefully not more than a week or two, depending on how many have to retry due to (seemingly unavoidable) dropped network connections.


bensteinberg (approved):
🎉

PR #3235, opened 2022-12-05
Update Chromium

bensteinberg:
https://chromereleases.googleblog.com/2022/12/stable-channel-update-for-desktop.html

Note that we have skipped over 108.0.5359.71-2~deb11u1, the first release of this major version -- the chromium Debian package was only out for a day. https://chromereleases.googleblog.com/2022/11/stable-channel-update-for-desktop_29.html


matteocargnelutti (approved):
Tests pending, but LGTM 👍

Issue #3045, opened 2022-03-07
Ordering of organization root folders in folder tree


rebeccacremona:
What is determining what order these come back in? It seems to be repeatable, but I don't immediately see how they are sorted: it's not by org name, org creation timestamp, folder id....

I could have sworn this used to be alphabetical.......

kilbergr:
Hi @rebeccacremona I remember discussing this with you but want to make sure I'm capturing.

I believe this is an issue exclusively on the admin site. I took a picture having signed in as an admin running the site locally:
<img width="231" alt="Screen Shot 2022-10-25 at 12 01 17 PM" src="https://user-images.githubusercontent.com/7401870/197824226-c5690d40-e49c-4da0-99ce-aabb817f4fdc.png">

I see the `perma_link_folders` and `perma_folders` models. I believe this ticket is in reference to the latter. 

From what I understand, an example would be:
admin user id `1` with email `test_admin_user@example.com` and root folder `22`
a `perma_folders` object with id `22` called `Empty root folder`

Which of the following is what this ticket is targeting:

If there were more than 1 root folder associated with the admin user `1`, you'd want those returned alphabetically (these folders not represented in the image)

OR

Or is it within the various folders associated with admin user `1`, those folders should be returned alphabetically (these folders represented in the image)

rebeccacremona:
> I believe this is an issue exclusively on the admin site.

With nothing expanded, that folder tree shows users [top level folders](https://github.com/harvard-lil/perma/blob/69724bd37fc9fe02c7d0855027ea3130cdd912ff/perma_web/perma/models.py#L863). "Personal Links" is always first. If a user is sponsored, "Sponsored Links" is next. Then comes the [shared folder](https://github.com/harvard-lil/perma/blob/69724bd37fc9fe02c7d0855027ea3130cdd912ff/perma_web/perma/models.py#L1175) of every org the user is allowed to see. Lots of users can see multiple orgs: admin users can see all orgs; registrar users can see all the orgs associated with the registar; org users can see any orgs to which they have specifically been added. 

The org folders are not being returned in any useful order. (I swear `self.get_orgs()` used to come back alphabetically by `name`, but there's no evidence that was the case.) I think we need to change the various queries there to sort asc by name.

rebeccacremona:
> I see the perma_link_folders and perma_folders models

[...]

PR #3236, opened 2022-12-06
Alphabetical ordering of shared org folders

kilbergr:
This PR links to this issue: https://github.com/harvard-lil/perma/issues/3045 and orders folders as Personal Links, Sponsored Links (if sponsored), and finally alphabetical shared org folders.
It looks like this (for an unsponsored user):
Screen Shot 2022-12-06 at 5 24 43 PM


matteocargnelutti (approved):
LGTM 😄!

I am less familiar with this part of the codebase and am not sure about the indexing situation here, and how it could impact performance on this specific query. 

Cheers.

bensteinberg:
It looks like your PRs get their titles directly from the branch name, or maybe it's the other way around. In any case, you might want to name PRs with an evocative title. When I push a fork from the command line and create the PR in the GitHub web UI, it names the PR with the first commit message, which is sometimes but not always right.

rebeccacremona:
Performance looks fine to me. I checked locally against a prod-like DB, as an admin (who therefore has the max number of top-level folders), and it only took between 15-30ms, even though it isn't presently using an index:
```
Hash Right Join  (cost=94.27..4216.36 rows=2583 width=135) (actual time=0.819..23.508 rows=2583 loops=1)
  Hash Cond: (perma_folder.id = perma_organization.shared_folder_id)
  ->  Seq Scan on perma_folder  (cost=0.00..3716.41 rows=154541 width=78) (actual time=0.002..9.048 rows=154541 loops=1)
  ->  Hash  (cost=61.98..61.98 rows=2583 width=57) (actual time=0.796..0.796 rows=2583 loops=1)
        Buckets: 4096  Batches: 1  Memory Usage: 274kB
        ->  Seq Scan on perma_organization  (cost=0.00..61.98 rows=2583 width=57) (actual time=0.004..0.373 rows=2583 loops=1)
              Filter: (NOT user_deleted)
              Rows Removed by Filter: 315
Planning Time: 0.495 ms
Execution Time: 23.639 ms
```

That seems fine!

PR #3238, opened 2022-12-15
Upload files to existing "daily" digest IA items, Part I

rebeccacremona:
### Context

### This PR

There are some 162,082 Perma Links that we would have expected to find uploaded to existing "daily" Internet Archive items, but whose WARCs are not present. For example, [https://perma.cc/06N5Qy9rxNE](https://perma.cc/06N5Qy9rxNE)'s WARC is not included in https://archive.org/download/daily_perma_cc_2013-11-13.

This PR adds Celery tasks for uploading those links to the appropriate daily item.

It does -NOT- create new items in IA, if the appropriate daily item does not already exist.

This is intended as a gentle way to try out our upload code and test parallelization and the handling of rate-limiting. It will likely require tweaking, once we see how it works under real conditions.

### Deploying

There's nothing special required for deployment, though we may want to have another look at all the IA-related settings and make sure we like them. There are now a number of configurable retry limits that could want tweaking.

Before running against any sizeable number of files, we probably want to let IA know that we are resuming uploads, experimentally.

To queue uploads, in the Django shell, run:
```
from perma.tasks import upload_missing_files_to_internet_archive
upload_missing_files_to_internet_archive(limit)
```
where `limit` is the maximum number of links you want to enqueue.


bensteinberg (approved):
:fire:

PR #3239, opened 2022-12-15
Upload files to existing "daily" digest IA items, Part II

rebeccacremona:
We ran a batch of 10 links against the new tasks added in [Part I](https://github.com/harvard-lil/perma/pull/3238). They went off without a hitch!

But all 10 confirmation tasks failed: when the uploads haven't "finished" yet, things look a bit different than I expected... and evidently, I never ran into that when testing locally.

This PR tweaks the confirmation task accordingly. For style, we check to see that the given file "exists" before checking its metadata... though mere existence is insufficient in this case: it can definitely still be a work in progress at that point. For substance: when a metadatum is absent, we pass the empty string on to the validation code instead of `None`, so that it merely detects a mismatch between expectations and reality, instead of borking.

Unless something else is wrong, the confirmation tasks will now re-queue until the upload is indeed confirmed, or until the retry limit is reached.

### DIFF
--- a/perma_web/perma/tasks.py
+++ b/perma_web/perma/tasks.py
@@ -2419,7 +2419,7 @@ def confirm_file_uploaded_to_internet_archive(file_id, attempts=0):
         for k, v in expected_metadata.items():
             # IA normalizes whitespace idiosyncratically:
             # ignore all whitespace when checking for expected values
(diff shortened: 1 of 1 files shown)


bensteinberg (approved):

lizadaly (approved):

bensteinberg:
I'm going to take the liberty of merging and deploying this.

PR #3247, opened 2022-12-22
Upload files to existing "daily" digest IA items, Part VI

rebeccacremona:
Subsequent to [Part V](https://github.com/harvard-lil/perma/pull/3245), this is now running steadily. We are monitored to see if too many confirmation tasks fail; if they do, we should consider running less frequently than every 15 min, and could also consider reducing the allowed number of retries (if we aren't seeing the retries succeed, then what's the point of trying more than once or twice?).

As a result, we can see that this util for checking ratelimiting is too slow, at this scale, because it is checking too many buckets.

We're going to try capping at 100, and see if that reduction is sufficient.

### DIFF
--- a/perma_web/perma/utils.py
+++ b/perma_web/perma/utils.py
@@ -719,7 +719,7 @@ def patch_internet_archive(ia_module):
     ia_module.ArchiveSession.get_s3_load_info = get_s3_load_info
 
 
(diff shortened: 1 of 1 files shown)


bensteinberg (approved):

rebeccacremona:
> also consider reducing the allowed number of retries
We decided to do this to: we aren't seeing any benefit from large numbers of retries.

PR #3259, opened 2023-02-13
Upload files to and delete files from "daily" digest IA items, Part I

rebeccacremona:
### Context

### Previous Work

Earlier work [added missing files](https://github.com/harvard-lil/perma/pull/3238) to the daily digest Items created on our behalf by the IA team. That worked, with the exception of [4 Items where we lack sufficient permissions](https://github.com/harvard-lil/perma/pull/3249); to be resolved offline.

### This PR

This PR makes it possible to upload new files to IA, regardless of whether the appropriate daily IA Item has been created already or not.

It also makes it possible to delete files that have already been uploaded to daily Items.

It's a ton of code, which makes it hard to review ☹️. I apologize for that...

#### Included tasks ####

A given link can be uploaded by calling `upload_link_to_internet_archive.delay(guid)`and deleted by calling `delete_link_from_daily_item.delay(guid)`. (The deletion task specifies that we are deleting from a "daily item" because we have work to do in the future, to delete the files from the legacy "individual" items; we can rename afterwards.)

Both upload and deletion tasks take IA some time to process, even after the API returns a 200. Sometimes, requests are completed quickly, but I believe they may take arbitrarily long. (Today, it seems that requests are being processed after about a 12 minute delay.) So, the tasks to confirm whether upload/deletion attempts were successful (`confirm_file_uploaded_to_internet_archive` and `confirm_file_deleted_from_daily_item`) are not scheduled by the upload/deletion tasks themselves. Instead, celerybeat is configured to `queue_file_uploaded_confirmation_tasks` and `queue_file_deleted_confirmation_tasks` every 15 minutes.

I have not, at this point, written any tasks that attempt to keep us up-to-date going forward: all this is still for working through the backlog.

You can schedule missing uploads for a particular day using `queue_internet_archive_uploads_for_date.delay(date_string, limit=None)`, or for a range of days using `queue_internet_archive_uploads_for_date_range.delay(start_date_string, end_date_string)`, where `date_string` is in the same format as the IA Items' identifiers: `YYYY-MM-DD`.

You can schedule the deletion of all no-longer-eligible-for-IA files using `queue_internet_archive_deletions.delay(limit=None)`. There were so few such files, it didn't seem worth it to restrict by day.

The config for all this is finicky, but I've _tried_ to set it up so that all tasks that actually upload and delete are run on the IA queue, and everything else is run on the read-only IA queue, but I tend to make a lot of mistakes setting this up. A second set of eyes would be appreciated; we should also keep an eye out, when running tasks, to see if they end up on the expected queues.

#### Derive and `cascade_delete`

Uploads do not trigger `derive.php` on the parent Item: we want to do this later, in consultation with IA about their preferences. (Much of the point of this entire project is to avoid overwhelming their system with too many expensive calls to `derive.php`).

Deletions don't trigger `derive.php` either, and the `cascade_delete` flag is set to `False`. We don't yet know how these differ, if at all: `cascade_delete=False` feels safest for the time-being.

#### Concerns about concurrency

It is not possible to create an IA Item outside of the upload process: one can only issue a request for a file to uploaded to a given identifier. IA creates an Item with that identifier, if it doesn't already exist, and otherwise reuses the existing Item.

During limited local testing, this appears to result in the scheduling of two separate `archive.php` tasks, whenever a new Item was created, but only a single `archive.php` task when the Item already exists. 

I am somewhat concerned that if many uploads are requested in parallel, all before an Item with a given identifier is created, redundant `archive.php` tasks will be queued... possibly in a way that results in errors. I think we'll just have to see if their system is equipped to handle the situation.

If not, we'll want to reconsider one of the other designs considered during the preparation of this PR, one which prepares the Item separately using a placeholder file, and which only attempts to upload links' warcs when Items are confirmed to exist.

### Deploying

Due to a bug fixed in this PR, a number of current `InternetArchiveItem` objects have non-zero values for `tasks_in_progress`. We'll want to reset everything to zero, before running any more tasks.

We probably should test this by first updating the IA-specific Grafana dashboard to track these log messages, and then queuing a single day's links to be uploaded, with a small limit (say, 10). If that goes well, then we should queue up the rest of that day's links, and a second day's. If that goes well, we should queue some small date range, and etc., adjusting concurrency and the appropriate schedule for the celerybeat "confirmation" task accordingly.

We can probably test the deletion task in parallel. 

TODO: I will check to see how many files presently need to be deleted. I will also find out the complete date range of the backlog.


perma_web/perma/models.py
```diff
@@ -2077,6 +2085,19 @@ def __str__(self):
     def datetime(cls, datetime_string):
         return datetime.strptime(datetime_string, '%Y-%m-%d %H:%M:%S').replace(tzinfo=timezone.utc)
 
+    @classmethod
+    def standard_metadata_for_date(cls, date_string):
+        iso = datetime.strptime(date_string, '%Y-%m-%d').date().isoformat()
```
lizadaly: Does this transformation actually change the string? Isn't `%Y-%m-%d` already `isoformat()`?

```python
>>> datetime.strptime('2020-01-01', '%Y-%m-%d').date().isoformat()
'2020-01-01'
```

perma_web/perma/models.py
```diff
@@ -2024,6 +2027,11 @@ class InternetArchiveItem(models.Model):
     - not contain more than 10,000 files.
 
     (From documentation archived at https://perma.cc/S4WC-64AF)
+
+    Expect one InternetArchiveItem object for every item in the perma_cc Internet Archive collection.
+
+    **If the 'cached_title' field is empty, it means we have attempted to create that item, but have
+    not yet confirmed whether that attempt succeeded or failed; failed attempts are not uncommon.**
```
lizadaly: Is there a way to be more explicit about this condition than inferring the absence of a value?

perma_web/perma/tasks.py
```diff
@@ -2204,21 +2204,65 @@ def upload_link_to_internet_archive(link_guid, attempts=0, timeouts=0):
...
+        span=(start, end),
+    )
+
+    # Check the Internet Archive-related status of this link
+    perma_file = InternetArchiveFile.objects.filter(
+        item_id=identifier,
+        link_id=link_guid
+    ).first()
+    if perma_file:
+        if perma_file.status == 'confirmed_present':
```
lizadaly, line 2226: Using an [enum](https://docs.python.org/3.9/library/enum.html) for status values like this would reduce the likelihood of typos putting things into invalid statuses.

perma_web/perma/tasks.py
```diff
@@ -2204,21 +2204,65 @@ def upload_link_to_internet_archive(link_guid, attempts=0, timeouts=0):
...
+            logger.info(f"Not uploading {link_guid} to {identifier}: our records indicate it is already present.")
+            return
```
lizadaly, line 2228: I would expect this function to return a result; perhaps a boolean representing success or failure, or your custom task status. A return value from a task becomes available as [task.result](https://docs.celeryq.dev/en/stable/reference/celery.result.html#celery.result.AsyncResult.result) that celery can use as part of a [workflow](https://docs.celeryq.dev/en/stable/userguide/canvas.html) to determine whether a task retries or moves on to the next thing in the chain. If you persisted this with a celery result backend it's also queryable. 

Lastly, it would be helpful for a human reader to better understand the control flow; the way some blocks early exit and others don't is a little confusing to read.

perma_web/perma/tasks.py
```diff
@@ -2204,21 +2204,65 @@ def upload_link_to_internet_archive(link_guid, attempts=0, timeouts=0):
...
+        elif perma_file.status == 'deletion_attempted':
+            # If we find ourselves here, something has gotten very mixed up indeed. We probably need a human to have a look.
+            # Use the error log, assuming this will happen rarely or never.
+            logger.error(f"Please investigate the status of {link_guid}: our records indicate a deletion attempt is in progress, but an upload was attempted in the meantime.")
```
lizadaly, line 2232: If this is rare and serious, should it throw an exception instead?

perma_web/perma/tasks.py
```diff
@@ -2246,146 +2290,98 @@ def upload_link_to_internet_archive(link_guid, attempts=0, timeouts=0):
...
+
+            response = ia_item.upload_file(
+                body=temp_warc_file,
+                key=InternetArchiveFile.WARC_FILENAME.format(guid=link_guid),
+                metadata=InternetArchiveItem.standard_metadata_for_date(date_string),
+                file_metadata=InternetArchiveFile.standard_metadata_for_link(link),
+                access_key=settings.INTERNET_ARCHIVE_ACCESS_KEY,
+                secret_key=settings.INTERNET_ARCHIVE_SECRET_KEY,
+                queue_derive=False,
+                retries=0,
```
lizadaly, line 2324: I assume something about this problem led to a decision to not leverage celery's built-in retry logic, but if this needs to be revisited, or you're designing a data sync in the future, I think a lot of heartache can be avoided by thinking of a sync as a series of small, discrete tasks managed by an orchestration system that handles the logic of retries with exponential backoff, task chaining, and task monitoring.

There are lots of good choices, all in Python—easiest would probably be to start with celery itself. I don't think their docs do a great job of explaining orchestration, but there are other guides: https://dev.to/akarshan/the-curious-case-of-celery-work-flows-39f7 (text), https://www.youtube.com/watch?v=XoMu8vhdc-A (video), and https://av.tib.eu/media/44051 (video).

perma_web/perma/tasks.py
```diff
@@ -2503,3 +2512,365 @@ def confirm_file_uploaded_to_internet_archive(file_id, attempts=0):
...
+            (perma_link.creation_timestamp AT TIME ZONE 'UTC')::DATE = %s
+          )
+          AND (
+            perma_link.user_deleted = False AND perma_link.is_private = False AND perma_link.is_unlisted = False AND perma_link.cached_can_play_back = True
+          )
+    '''
+
+    # Get all links we think should have been uploaded to IA that are not yet associated with a daily InternetArchiveItem.
+    if limit:
+        sql += "LIMIT %s"
```
lizadaly:
```suggestion
        sql += " LIMIT %s"
```

Minor but add whitespace to guard against appending this token directly adjacent to a previous one leading to invalid SQL.

lizadaly (approved):
Nothing blocking trying this out; I left suggestions for future refactors that might make this task more manageable. 

Note there's some CSS in here which doesn't seem related, but maybe it's just an improvement along for the ride?

perma_web/perma/models.py
rebeccacremona: LOL!!! Uh, yup!

perma_web/perma/tasks.py
```diff
@@ -2427,10 +2423,11 @@ def queue_file_uploaded_confirmation_tasks(limit=None):
 @shared_task(acks_late=True)
 def confirm_file_uploaded_to_internet_archive(file_id, attempts=0):
     """
-    This task is enqueued by upload_link_to_internet_archive after it finishes uploading a WARC to
-    IA's S3-like API. It checks to see if the requested upload has been processed and the new WARC
-    is now visibly a part of the expected IA Item; if not, the tasks re-queues itself up to
-    settings.INTERNET_ARCHIVE_RETRY_FOR_ERROR_LIMIT times.
+    This tasks checks to see if a WARC uploaded to IA's S3-like API has been processed
```
bensteinberg:
```suggestion
    This task checks to see if a WARC uploaded to IA's S3-like API has been processed
```

perma_web/perma/tasks.py
```diff
@@ -2427,10 +2423,11 @@ def queue_file_uploaded_confirmation_tasks(limit=None):
...
+    and the new WARC is now visibly a part of the expected IA Item;
+    if not, the tasks re-queues itself up to settings.INTERNET_ARCHIVE_RETRY_FOR_ERROR_LIMIT times.
```
bensteinberg:
```suggestion
    if not, the task re-queues itself up to settings.INTERNET_ARCHIVE_RETRY_FOR_ERROR_LIMIT times.
```

bensteinberg (commented):
Two small bits of copyediting

perma_web/perma/models.py
rebeccacremona: Good idea. I added a [field for the purpose](https://github.com/harvard-lil/perma/pull/3259/commits/d6f5ebd19fe14d6c2ab81e14454af4ef2fac6e86)

perma_web/perma/tasks.py
rebeccacremona: When you use an `enum` for things like this, do you feel comfortable omitting the `else` at the end of switch-style blocks? I find myself still including it... maybe coding against the possibility that in the future, another legal value is added, but we forget to account for it.

rebeccacremona: Thanks for suggesting the enum. I am used to relying on Django's validation of choices to reject invalid values.... but since we only see that if calling `full_clean`, which we don't routinely do, much better idea!

lizadaly: In general I avoid "just in case" code because I'm bad at predicting future bugs. If you're using types it should be impossible to introduce that particular problem:

```python
    from enum import Enum, auto

    class Status(Enum):
        SUCCESS = auto()
        FAILURE = auto()

    status: Status
    status = Status.SUCCESS # this is fine
    status = "fake" # mypy complains here
```

lizadaly: Oh I see, I guess you might forget to handle the new case, but I'd really expect that to be caught by code review.

perma_web/perma/tasks.py
rebeccacremona: The pattern I learned for Celery tasks was, work hard to avoid uncaught exceptions, so that the task itself finishes and is a "success" so far as Celery is concerned. Kind of like... if an uncaught exception is thrown, it's the equivalent of an `E` in a pytest test: something has gone wrong with the code, rather than, the task/test "failed" with an `F`.

It may be that is not the best pattern.

Thanks for the nudge. Will definitely read about that, as well as [returning and storing results](https://github.com/harvard-lil/perma/pull/3259#discussion_r1105850823) as per your other comment.

perma_web/perma/tasks.py
rebeccacremona: 👀 Thanks. I don't know that I/we have taken advantage of Celery's built-in retry logic and orchestration tools before. If the legacy version of these tasks failed, we relied on regularly-scheduled Celerybeat tasks to requeue things, based on our own "status" fields. To be studied!

rebeccacremona:
> Note there's some CSS in here which doesn't seem related, but maybe it's just an improvement along for the ride?

Ah, the CSS is indeed on purpose and quasi-related, and I forgot to remark about it. Our internal /manage/stats page lets us check in on our IA-related rate limit situation in real time: without this CSS tweak, the columns in the "bucket" section below bleed into each other rather than wrapping.

image

rebeccacremona:
> Nothing blocking trying this out; I left suggestions for future refactors that might make this task more manageable.

Thanks for all the feedback, @lizadaly!

I'll keep all the suggestions on using Celery more effectively in mind for future work. Sounds like I've got some studying to do. Looking forward to seeing how you have used Celery in the H2O PDF export task, and comparing it to this pattern.

perma_web/perma/tasks.py
lizadaly: I think it just depends on the semantics of the error—the comment made me think this was a real bad unexpected state that I'd want to be alerted about.

rebeccacremona: I'm a little less sure how this works now that we are using Sentry. In the olden times, writing to `logger.error` like this would send us error emails, just like raising an exception would.

lizadaly: Yeah, I forgot this is still Django's default logging behavior. I wish it weren't, because I think it muddies the waters between an alert and a log entry.

PR #3260, opened 2023-02-14
Delete transitional IA tasks

rebeccacremona:
Over the past several months, I've written a lot of one-time Celery tasks as part of an ongoing effort to reorganize Perma's Internet Archive collection ([internal project description](https://docs.google.com/document/d/1yeiW47MahmTPC4Sh4zunCXgaapBTX8k8DjDb8r-9ifM/edit#heading=h.j7du1d5sm17f); [project board](https://github.com/orgs/harvard-lil/projects/6/views/1)).

This PR deletes most of that code.

We don't need it any more, and I don't think we need it around for easy consultation during further work.


matteocargnelutti (approved):
It's hard to say good bye 😢 👋

PR #3261, opened 2023-02-14
Upload files to and delete files from "daily" digest IA items, Part II

rebeccacremona:
There was a bug in [Part I](https://github.com/harvard-lil/perma/pull/3259); I honestly cannot account for how it slipped through. I must have introduced it somehow during refactoring, and thought I tested, but did not.

Raw querysets consist of model objects, with the fields not retrieved by the SQL [deferred](https://docs.djangoproject.com/en/4.1/topics/db/sql/#deferring-model-fields). So, this query does not return data of the same shape as `values_list('guid', flat=True)`.

Pass the guid to the task, not the whole link object.


bensteinberg (approved):
👍

PR #3267, opened 2023-02-16
Make output keys match

bensteinberg:
This should fix a KeyError that we saw yesterday, #3266.


lizadaly (approved):

rebeccacremona (approved):
Thanks! Interesting: I wonder why we ended up in this branch, just then. UPDATE: it was because the "global" capacity was exceeded, which is to say, they were swamped by people other than us.

PR #3268, opened 2023-02-16
Add deployment sentinel

bensteinberg:
As part of draining the queue of tasks before deployment, we suppress the addition of new capture jobs. The addition and removal of the sentinel is handled by a salt orchestration.

### DIFF
--- a/perma_web/api/views.py
+++ b/perma_web/api/views.py
@@ -27,6 +27,9 @@ from .serializers import FolderSerializer, CaptureJobSerializer, LinkSerializer,
 from django.conf import settings
 from django.urls import reverse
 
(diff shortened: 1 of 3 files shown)


bensteinberg:
I thought of putting this in `settings_prod`, but it seemed simplest to do it this way, and not have to check for the presence of the setting.

lizadaly (approved):
Do you want to warn if it’s being skipped anywhere so you can audit if it’s getting stuck or stale?

rebeccacremona:
> I thought of putting this in `settings_prod`

I like it in `settings_common`: we can even write a test for this later if we want!

PR #3269, opened 2023-02-21
Upgrade Django

rebeccacremona:
We have `DJANGO_ALLOW_ASYNC_UNSAFE=true` set in [CAP](https://github.com/harvard-lil/capstone/blob/78c1815a370bfef8d46c3c570af7706f57b5111c/capstone/docker-compose.yml#L47)'s docker-compose.yml, but [H2O](https://github.com/harvard-lil/h2o/blob/308c51ea50d073e8943afe33f3485d0c11ee94c5/web/conftest.py#L42) does it in the test config. I went with H2O's style.

### DIFF
--- a/perma_web/api/urls.py
+++ b/perma_web/api/urls.py
@@ -1,6 +1,7 @@
 from django.conf import settings
-from django.conf.urls import url, include
+from django.conf.urls import include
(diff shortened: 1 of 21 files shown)


lizadaly (approved):
🎉

PR #3275, opened 2023-02-27
Replace Fabric with Invoke

bensteinberg:
Since we only use Fabric to run things locally, we can switch to Invoke, removing a few dependencies. 

Notable changes for developers who use Invoke tasks include replacing underscores with hyphens: `invoke dev.init-db` instead of `fab dev.init_db` -- and boolean arguments are really boolean now: `invoke run --debug-toolbar` instead of `fab run:debug_toolbar=True`

When writing new Invoke tasks, the two things to remember are including the context as the first argument, and that arguments are not necessarily strings, as they are in Fabric3 -- see https://docs.pyinvoke.org/en/stable/concepts/invoking-tasks.html#type-casting

See also https://github.com/harvard-lil/h2o/pull/1916 and https://github.com/harvard-lil/perma-payments/pull/165. There's some discussion at the latter about whether and how to change the names of files in which Invoke and Celery look for tasks. The short version is that we are going to let the less-configurable Invoke have `tasks.py`, or `tasks/` in this case, and I've chosen to give Celery `celery_tasks.py`.

I will need to update the `CHECK_IA_PENDING` and `CHECK_LINKS_WITHOUT_CACHED_PLAYBACK_STATUS` cron jobs in salt before deploying this to production.

This might be an opportunity to remove obsolete tasks.


matteocargnelutti (approved):
Nice -- Works as expected on my machine (although I had a brief pip hiccup) 👍 .

rebeccacremona (approved):
🙏 Thanks for doing this!!

Yeah, maybe we can follow up soon with a PR to remove a bunch of the legacy tasks. Maybe also condense into a `tasks.py` file instead of retaining the old directory and dev/ structure. Nice clean diff this way, though.

PR #3286, opened 2023-03-16
Upload files to and delete files from "daily" digest IA items, Part VII

rebeccacremona:
TLDR: conditionally queue more Internet Archive upload jobs on a schedule, depending on how many are presently inflight

### Context

Recall that we are reorganizing Perma's [Internet Archive collection](https://archive.org/details/perma_cc). In the past, we would add an "Item" with a single "File" to the collection for each new Perma Link; going forward, we've been asked to instead create one digest-like Item per day, with Files for each of the Perma Links created on that day.

### Previous Work

We are pretty far along in the process. Earlier work [added missing files](https://github.com/harvard-lil/perma/pull/3238) to the daily digest items created on our behalf by the IA team. That worked, with the exception of [4 items where we lack sufficient permissions](https://github.com/harvard-lil/perma/pull/3249); to be resolved offline.

This PR is the latest in a series that [creates new daily items](https://github.com/harvard-lil/perma/pull/3259) and uploads files to them, rather than just adding files to already-existing items. The code as originally designed worked, as such, but needed tuning after we saw a) how Internet Archive handled requests made in parallel and b) got some hands-on experience with the capacity of Internet Archive's internal queuing system, observing how our upload rate and queuing strategy interacted with their usage limits and variable processing speed.

Our initial experiments [flooded the system](https://github.com/harvard-lil/perma/pull/3263) pretty quickly; we paused to design a more thoughtful queuing approach, one that can adapt to changing conditions and queue jobs without an excessive amount of churn (failures and retries) or duplication (adding a redundant upload requests to the queue for jobs that just haven't been processed yet due to slower-than-normal rates). 

### This PR

This PR attempts to automate a strategy that worked well during manual testing. 

We found that queuing up to 100 files at a time for up to 4 separate items was routinely fine: sometimes other people were using the API at the same time and the `global` task limit was exceeded, but we did not approach our API key's limit or per-Item/per-bucket limit, with Celery's concurrency set to 4. Batches of tasks queued using that strategy generally completed in 30 minutes or less; occasionally, batches took less than 8 min.

Here, I schedule `conditionally_queue_internet_archive_uploads_for_date_range` to run every 15 minutes. It checks to see how many upload tasks are in progress or are in the queue, and then queues some more, up to the ~400 at a time limit we saw seemed to work okay manually (`settings.INTERNET_ARCHIVE_MAX_SIMULTANEOUS_UPLOADS`). It spreads itself out over a date range (`start_date_string`, `end_date_string`), taking care that not more than ~100 tasks (`daily_limit`) are inflight for any particular day/bucket/item at a time.

If the Celery queue is not empty when it runs, it will queue up at least some duplicative tasks. I would like for that not to be the case, but I can't think of a strategy that doesn't involve keeping a more closer watch on what is and isn't in the celery queue than we presently do... which I'm hoping just doesn't prove necessary. Hopefully the amount of churn will be small or manageable. If there is a Part VIII in this series, it will probably be because this hope doesn't pan out, and decreasing the task's frequency (say to every half hour) doesn't get us the performance we need. (We could consider a different strategy now, when we are working through the backlog, than we'd need in the future, when we are staying on top of things... for instance, for now, adjusting the date range to non-overlapping ranges on subsequent runs, or etc. etc.)

### While I was in there

I made a few small tweaks to the relevant Django admin pages and our internal stats pages while I was working, all intended to help us keep a closer eye on this while it's running. (I'll update the Grafana dashboard too.)

### Deploying

I do not have the scheduled celerybeat task enabled at present; I want to try a few batches out in the shell on our own time first, e.g. 
```
from perma.celery_tasks import conditionally_queue_internet_archive_uploads_for_date_range

conditionally_queue_internet_archive_uploads_for_date_range('2023-01-01', '2023-01-30', limit=200)

conditionally_queue_internet_archive_uploads_for_date_range('2023-01-01', '2023-01-30',  daily_limit=10)

conditionally_queue_internet_archive_uploads_for_date_range('2023-01-01', '2023-01-30')
```

If that seems to go well, we can toggle and see how it goes. 

We can adjust the date range by setting `IA_UPLOAD_START_DATESTRING` and `IA_UPLOAD_END_DATESTRING` environment variables. I have it defaulting to the entire length of the backlog. I'd appreciate a second person helping reasoning through whether there is any reason to run with a smaller subset. I don't think so?


bensteinberg (approved):
This looks great. Thanks! I agree that the range from `IA_UPLOAD_START_DATESTRING` to `IA_UPLOAD_END_DATESTRING` can default to the entire length of the backlog.

PR #3289, opened 2023-03-21
Upload files to and delete files from "daily" digest IA items, Part X

rebeccacremona:
This PR fixes an oversight in this series: the queuing function assumes that at least one link will always be queued. If it isn't, [this logging line](https://github.com/harvard-lil/perma/compare/develop...rebeccacremona:completed-days?expand=1#diff-d8b159c0525b4635c3804b94e55f7d489d32862fd98bc145c471bcd4a1ac5652L2188) throws `IndexError`s. 

Instead of just fixing the log line, I decided to add special handling for that case. When no eligible links are found for a given day, we provisionally mark that day's InternetArchiveItem "complete." 

That metadatum won't necessarily be true, going forward: planned future work, which will handle Perma Links that are toggled from public to private and vice versa, will be toggling this field and triggering the appropriate work... But, it is useful in the time being: we can use it to determine where in the backlog to start queuing, when `conditionally_queue_internet_archive_uploads_for_date_range` runs. 

This PR arranges for `conditionally_queue_internet_archive_uploads_for_date_range` to look for the most recent 'complete' day, and start the day after that.

### Deploying

Before this is deployed, I will go into the Django admin or the shell, and mark the 4 days we finished working on last night "complete".

### DIFF
--- a/perma_web/perma/celery_tasks.py
+++ b/perma_web/perma/celery_tasks.py
@@ -2282,8 +2282,9 @@ def conditionally_queue_internet_archive_uploads_for_date_range(start_date_strin
             bucket_limit = min(daily_limit, to_queue - total_queued) - in_flight_for_this_day
             if bucket_limit > 0:
                 count_queued = queue_internet_archive_uploads_for_date(date_string, bucket_limit)
-                total_queued += count_queued
-                queued.append(f"{date_string} ({count_queued})")
+                if count_queued:
+                    total_queued += count_queued
+                    queued.append(f"{date_string} ({count_queued})")
         else:
             break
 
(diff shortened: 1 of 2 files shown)


bensteinberg (approved):

rebeccacremona:
I might need to move the code about timing the query; checking. Actually, this might cause the query to be evaluated twice. Checking that as well.

rebeccacremona:
Yup, I need to do both of those things. Hold one.

rebeccacremona:
Okay, addressed. Still LGTY, @bensteinberg?

bensteinberg (approved):

PR #3290, opened 2023-03-21
Upload files to and delete files from "daily" digest IA items, Part XI

rebeccacremona:
With the introduction of [Part X](https://github.com/harvard-lil/perma/pull/3289), this seems to be plugging away steadily. We have not yet seen the system respond to the triggering of the global rate limit, but we can see it responding to IA's rate of processing speeding up and slowing down.

So far, there has been plenty of head room: this PR ups the number of simultaneously allowed uploads from 400 to 499, our API key's explicit limit. I think other protection mechanisms should make sure there is still enough wiggle room to account for derivative tasks, but as usual, we'll have to see how that plays out in real life. To be avoided if possible: the celery queue still having uploads in progress, at the quarters of the hour.

This PR also makes a few tweaks to make monitoring more pleasant: a shuffling of the /manage/stats page, and a few small tweaks to log messages.

image


perma_web/perma/settings/deployments/settings_common.py
```diff
@@ -471,7 +471,7 @@
 INTERNET_ARCHIVE_ACCESS_KEY = ''
 INTERNET_ARCHIVE_SECRET_KEY = ''
 # Rate limiting
-INTERNET_ARCHIVE_MAX_SIMULTANEOUS_UPLOADS = 400  # manual testing suggests IA can usually handle 400 at a time without getting swamped
+INTERNET_ARCHIVE_MAX_SIMULTANEOUS_UPLOADS = 499  # as of 3/21, this is our "accesskey_ration"
```
bensteinberg:
```suggestion
INTERNET_ARCHIVE_MAX_SIMULTANEOUS_UPLOADS = 499  # as of 2023-03-21, this is our "accesskey_ration"
```

bensteinberg (approved):

PR #3300, opened 2023-04-03
Make the GUID fuzzy match in the admin speedy.

### LINKED ISSUE #2951: "By GUID" filter in Link admin is slow/intensive
find out why. is it triggering counts?

rebeccacremona:
Fixes https://github.com/harvard-lil/perma/issues/2951.

Before:
image

After:
image


bensteinberg (approved):
Nice!

perma_web/perma/settings/deployments/settings_common.py
```diff
@@ -137,6 +137,7 @@
     # built in apps
     'django.contrib.auth',
     'django.contrib.contenttypes',
+    'django.contrib.postgres',
```
lizadaly, line 140: Is this an app in the Django sense?

rebeccacremona: Weird, right? It's required, for a person to use `OpClass` https://docs.djangoproject.com/en/4.1/ref/contrib/postgres/indexes/#opclass-expressions

lizadaly: TIL! 👍

Issue #3296, opened 2023-03-29
Django's built-in "last_login" field on user not being updated


rebeccacremona:
Since [this deployment in 2020](https://github.com/harvard-lil/perma/compare/v0.210...v0.211), which included our upgrade from Django 1 to Django 2, the "last_login" field, part of Django's `AbstractBaseUser` model, updated via signal... has not been updated or populated for any user.

I have not yet diagnosed why. Our [pre-save signal for Links](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/signals.py#L7) is working, so it is not the case that all signals are broken.

A few months after that deployment, we added [django-axes](https://github.com/harvard-lil/perma/pull/2860), so if we want, we can backfill the appropriate value of this field for any user who has logged in since 2020-10-15 16:40.

rebeccacremona:
Okay, this happened during the upgrade to [Django 2.1](https://github.com/django/django/pull/9712/files), because of the change introduced to Django by [this PR](https://github.com/django/django/pull/9712/files).

It only registers the `update_last_login` signal if the user model's `last_login` is an instance of `django.db.models.query_utils.DeferredAttribute`.

But, we are using [Django Model Utils' `FieldTracker`](https://django-model-utils.readthedocs.io/en/latest/utilities.html#field-tracker) to [track changes made to our user objects' fields](https://github.com/harvard-lil/perma/blob/v0.426/perma_web/perma/models.py#L841).... so, for us, 
```
last_login_field = getattr(get_user_model(), 'last_login', None)
```
is a `model_utils.tracker.DescriptorWrapper`, not a `django.db.models.query_utils.DeferredAttribute`.

Possible fixes:
- register the signal ourselves
- exclude `last_login` from the fields being tracked
- swap in LIL's [custom field-tracking code](https://github.com/harvard-lil/h2o/blob/v20230404-01/web/main/models.py#L107), which Jack wrote because "those packages can be error-prone in hard-to-diagnose ways", if it works
- maybe other ideas too

I'll see what seems to be cleanest.

rebeccacremona:
See https://github.com/jazzband/django-model-utils/issues/423 and https://github.com/jazzband/django-model-utils/issues/497

rebeccacremona:
[...]

PR #3302, opened 2023-04-04
Backfill last_login from axes logs

rebeccacremona:
Now that https://github.com/harvard-lil/perma/issues/3296 is fixed going forward, this PR fixes our data, to the extent that we can: it looks in the axes login records and finds the most recent one for each user, and puts that date into `last_login`. There's a gap of a few months where we just don't have data... but c'est la vie.

Run locally against ~80K records in a prod-like database, this took 1.5s... so I decided to do it cleanly as a data migration. So, if it takes a long time, that means we'll have a slower deployment than usual... but I think even if it ends up being much slower, adding a minute or two while we are in maintenance mode is perfectly fine.

I'm not sure where the `print`s will be visible, because this will happen as part of a Salt orchestration. @bensteinberg do you know? if you would recommend using the `logger` instead of printing for any reason, happy to switch.


matteocargnelutti (approved):
(Tests still running and small lint tweak needed but ...) NICE!

bensteinberg:
> I'm not sure where the `print`s will be visible

I think if they are sent to stdout when we run `manage.py migrate`, we'll be able to see them by looking at the output of the perma-migrate state run -- and we can confirm this when we deploy to stage.

PR #3303, opened 2023-04-05
Create extensions only if possible

bensteinberg:
I offer this for discussion wrt our current problem with perms. This change, or something like it, will require extension creation to be out-of-band in RDS and similar environments.


bensteinberg:
(Flailing; almost there.)

rebeccacremona (commented):
Neat. This seems like a fine strategy to me! I defer to you and the others re: deciding.

rebeccacremona:
See https://github.com/harvard-lil/perma/issues/3308

Issue #3308, opened 2023-04-05
Install postgres extensions locally and in prod


rebeccacremona:
I recently introduced a [new index](https://github.com/harvard-lil/perma/pull/3300/files) to the `LinkUser` table, one that relies on two postgres extensions, `btree_gin` and `pg_trgm`. 

Django includes helpers for [installing the extensions](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/migrations/0016_link_guid_case_insensitive_idx.py#L20-L21) during a migration, but in our version of postgres, the extensions need to be installed by a superuser. That works fine in development and CI, but not in production: the production database user is not, and should not be, a superuser. 

More recent versions of postgres have the concept of "trusted extensions", which can be installed by any users with `CREATE` privileges... but upgrading to a newer postgres isn't on the table right now... especially given how unimportant this new index is.

We considered a number of workarounds. (See https://hlslil.slack.com/archives/C03CYM6TF3J/p1680640861403829 for group discussion, which was picked up later by @rebeccacremona and @bensteinberg in DMs and live conversation.) Among them:

- In keeping with the pythonic value that it's easier to ask for forgiveness than permission, wrap the installation with try/except such that failures aren't fatal.... and instead, fall through to the fatal error that would be thrown when the index creation is attempted, if the extensions aren't installed. See https://github.com/harvard-lil/perma/pull/3303, https://github.com/harvard-lil/perma/pull/3304, and https://github.com/harvard-lil/perma/pull/3305. That approach proved surprisingly finicky: migrations are intricate, and to make this work as desired, we'd have to further study how they use transactions.

- Instead of doing that, we thought of changing the migration to be conditional based on a tier-level setting, so that it didn't attempt the installation at all in staging and production. See https://github.com/harvard-lil/perma/pull/3307/files.

- We considered install the extension via docker config by customizing the initdb step, as in https://stackoverflow.com/questions/55741735/docker-compose-and-postgres-extensions. But, we didn't like that this would require us to throw away our existing volumes in dev. (Which we totally can do... it's just annoying, and means we lose a certain amount of useful-to-us captures.)

- Or, we could make our own Dockerfile for postgres, and install it there.

- We might also be able to get something working by using docker entrypoint script.

- We considered adding logic to the [command that devs run during local installation](https://github.com/harvard-lil/perma/blob/develop/perma_web/tasks/dev.py#L111) already, and simply explaining to devs with existing installations what do to.

[...]

PR #3309, opened 2023-04-06
Remove custom logic for where to install pg extensions

rebeccacremona:
[This custom logic did not work](https://github.com/harvard-lil/perma/issues/3308#issuecomment-1498072179): while we did get further in the migration, the creation of the index still failed. This PR removes it.

In the meantime, @bensteinberg was able to find a solution on the DB-side. If this PR deploys successfully on stage, which has been migrated back to 0015, then we can close the issue.


bensteinberg (approved):
Sorry about the wheel-spinning yesterday -- I think what I should have realized earlier is that the extensions need to belong to a database (and schema), not just the database instance.

PR #3315, opened 2023-04-12
Don't mark an IA item complete if it's yesterday's

bensteinberg:
We are leaving some IA uploads undone, because we're marking the item complete too early. This does not mark an item complete if it's yesterday's item. 

I'm not sure about the use of `InternetArchiveItem.datetime(f"{date_string} 00:00:00")` as the canonical datetime for the item, but I saw it done [here](https://github.com/harvard-lil/perma/blob/43d389936e8beac8ff02a9c8e0e1ba52bb17d067/perma_web/perma/celery_tasks.py#L1627).

### DIFF
--- a/perma_web/perma/celery_tasks.py
+++ b/perma_web/perma/celery_tasks.py
@@ -2266,7 +2266,7 @@ def queue_internet_archive_uploads_for_date(date_string, limit=100):
         try:
             item = InternetArchiveItem.objects.get(identifier=identifier)
             # Don't mark an item complete if it's yesterday's
-            if timezone.now() - InternetArchiveItem.datetime(f"{date_string} 00:00:00") > timedelta(days=1):
+            if timezone.now() - item.span.lower > timedelta(days=1):
                 item.complete = True
                 item.save(update_fields=['complete'])
                 logger.info(f"Found no pending links: marked IA Item {item.identifier} complete.")
--- a/perma_web/perma/celery_tasks.py
+++ b/perma_web/perma/celery_tasks.py
@@ -2265,9 +2265,13 @@ def queue_internet_archive_uploads_for_date(date_string, limit=100):
         )
         try:
             item = InternetArchiveItem.objects.get(identifier=identifier)
-            item.complete = True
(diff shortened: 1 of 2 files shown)


perma_web/perma/celery_tasks.py
```diff
@@ -2265,9 +2265,13 @@ def queue_internet_archive_uploads_for_date(date_string, limit=100):
         )
         try:
             item = InternetArchiveItem.objects.get(identifier=identifier)
-            item.complete = True
-            item.save(update_fields=['complete'])
-            logger.info(f"Found no pending links: marked IA Item {item.identifier} complete.")
+            # Don't mark an item complete if it's yesterday's
+            if timezone.now() - InternetArchiveItem.datetime(f"{date_string} 00:00:00") > timedelta(days=1):
```
rebeccacremona: Cool! That works. I think slightly more idiomatic, since you have the item here, would be to use `item.span.lower`

```
In [16]: item = InternetArchiveItem.objects.filter(span__isempty=False).first()
In [17]: item.identifier
Out[17]: 'daily_perma_cc_2013-09-12'
In [18]: item.span.lower
Out[18]: datetime.datetime(2013, 9, 12, 0, 0, tzinfo=<UTC>
```

rebeccacremona (approved):
Thanks for this!! One suggested tweak to the line you called attention to.

bensteinberg:
Part of handling this new, small backlog will be to mark as incomplete the small handful of items spuriously marked complete. I'm currently working on something to run in the Django shell.

perma_web/perma/celery_tasks.py
bensteinberg: Thanks!

rebeccacremona:
> Part of handling this new, small backlog will be to mark as incomplete the small handful of items spuriously marked complete. I'm currently working on something to run in the Django shell.

Hmm. Could be we want this field to be editable in the Django admin, while this code is still incomplete. Would that be easier?

rebeccacremona (approved):

PR #3320, opened 2023-04-17
Merge duplicative user accounts

rebeccacremona:
This PR adds two `invoke` tasks: one for merging all duplicative user accounts that we believe can be safely merged, and one to reverse those changes.

They can be run from the command line, similar to:
- `docker compose exec web invoke dev.merge-duplicative-accounts`
- `docker compose exec web invoke dev.unmerge-duplicative-accounts --log-to-file 'merge_reports/reversed.txt'`

Or, for more control, you can import the helper functions and run on particular users within the Django shell:
```
In [1]: from tasks.dev import merge_users_with_only_one_confirmed_account, initialize_csvs, unmerge_accounts

In [2]: from perma.models import LinkUser

In [3]: users = LinkUser.objects.filter(email__iexact='To_Be_Merged@example.COM')

In [4]: initialize_csvs()

In [5]: merge_users_with_only_one_confirmed_account(list(users))

In [6]: unmerge_accounts(106472)
Unmerged accounts [106470, 106471] from user 106472
[<LinkUser: to_be_merged@example.com>, <LinkUser: to_be_merged@example.COM>] from {<LinkUser: To_Be_Merged@example.COM>}
```

I have tested fairly thoroughly locally against a production-like database, and am pretty confident that the merging in fact works, and that the unmerging in fact reverses everything.

I would appreciate a second set of eyes on the logic of the functions that select which user in a given group is the right one to "keep"/be the target of the merge:
- `merge_users_with_only_unconfirmed_accounts`
- `merge_users_with_only_one_confirmed_account`
- `merge_users_with_multiple_confirmed_accounts_but_no_links`
- `merge_users_with_only_one_account_with_links`
- `merge_users_with_multiple_accounts_with_links`

Do they in fact do what their docstrings claim? I think so! But, if I have a mistake, I think it's most likely to be there.

I plan to sanity check the (locally produced) logs of a few merged users from each category with @clare-stanton too.

I considered adding a `dry-run` arg to both tasks... but thought it probably wasn't necessary. Can add if it would make people feel better.

Closes: 
- https://github.com/orgs/harvard-lil/projects/11/views/1?pane=issue&itemId=25018104
- https://github.com/orgs/harvard-lil/projects/11/views/1?pane=issue&itemId=25019521
- https://github.com/orgs/harvard-lil/projects/11/views/1?pane=issue&itemId=25019557

### DIFF
--- a/perma_web/perma/models.py
+++ b/perma_web/perma/models.py
@@ -1116,6 +1116,10 @@ class LinkUser(CustomerModel, AbstractBaseUser):
 
         orgs = set()
         registrars = set()
+        if self.registrar_id:
+            registrars.add(self.registrar_id)
+        else:
+            orgs.update(original_orgs)
         for user in users:
             if user.registrar_id:
                 registrars.add(user.registrar_id)
@@ -1123,9 +1127,9 @@ class LinkUser(CustomerModel, AbstractBaseUser):
                 orgs.update(user.organizations.all())
 
         if orgs or registrars:
-            assert not (orgs and registrars), f"This set of users includes both org and registrar users: {', '.join([user.id for user in users])}."
+            assert not (orgs and registrars), f"This set of users includes both org and registrar users: {self.id}, {', '.join([str(user.id) for user in users])}."
             if registrars:
-                assert len(registrars) == 1
+                assert len(registrars) == 1, f"This set of users includes registrar users from multiple registrars: {self.id}, {', '.join([str(user.id) for user in users])}."
                 new_registrar_id = registrars.pop()
                 if not self.registrar_id:
                     self.registrar_id = new_registrar_id
--- a/perma_web/perma/models.py
+++ b/perma_web/perma/models.py
@@ -1109,6 +1109,64 @@ class LinkUser(CustomerModel, AbstractBaseUser):
         return not self.nonpaying or (self.is_registrar_user() and not self.registrar.nonpaying)
 
 
+    ### merging accounts ###
+
+    def copy_memberships_from_users(self, users):
+        original_orgs = set(self.organizations.all())
+
+        orgs = set()
+        registrars = set()
+        for user in users:
+            if user.registrar_id:
+                registrars.add(user.registrar_id)
+            else:
+                orgs.update(user.organizations.all())
+
+        if orgs or registrars:
+            assert not (orgs and registrars), f"This set of users includes both org and registrar users: {', '.join([user.id for user in users])}."
+            if registrars:
+                assert len(registrars) == 1
+                new_registrar_id = registrars.pop()
+                if not self.registrar_id:
+                    self.registrar_id = new_registrar_id
+                    self.prepend_to_notes(f"Added registrar during the merging of accounts: {new_registrar_id}")
+                    self.save(update_fields=['registrar_id', 'notes'])
+            else:
+                if original_orgs != orgs:
+                    self.prepend_to_notes(f"Added organizations during the merging of accounts: {', '.join([str(o.id) for o in orgs - original_orgs])}")
+                    self.save(update_fields=['notes'])
+                    self.organizations.add(*orgs)
+
+    def soft_delete_after_merge_with_user(self, user):
+        original_email = self.email
+
+        self.email = f"merged_users_{self.id}_and_{user.id}@perma.cc"
+        self.is_active = False
+        self.link_count = 0
+
+        self.prepend_to_notes(f"Original email: { original_email }")
+        if self.registrar_id:
+            self.prepend_to_notes(f"Original registrar: { self.registrar_id }")
+            self.registrar_id = None
+        orgs = list(self.organizations.all())
+        if orgs:
+            self.prepend_to_notes(f"Original orgs: {', '.join([str(o.id) for o in orgs])}")
+            self.organizations.remove(*orgs)
+
+        self.save(update_fields=['email', 'is_active', 'link_count', 'notes', 'registrar_id'])
+        return (original_email, self.email)
+
+    def prepend_to_notes(self, message):
+        if self.notes:
+            self.notes = f"{message}\n\n{self.notes}"
+        else:
+            self.notes = message
+
+    def remove_line_from_notes(self, containing):
+        if self.notes:
+            self.notes = re.sub(f"\n*{containing}.*", '', self.notes)
+
+
 class ApiKey(models.Model):
     """
         Based on tastypie.models: https://github.com/django-tastypie/django-tastypie/blob/master/tastypie/models.py#L35
--- a/perma_web/tasks/dev.py
+++ b/perma_web/tasks/dev.py
@@ -1064,7 +1064,7 @@ def get_and_categorize_duplicative_users():
     multiple_confirmed_several_with_links = set()
 
     #
-    # Identify any groups of users that are not same to merge
+    # Identify any groups of users that are not safe to merge
     #
     for normalized_email, user_group in grouped_duplicative_users.items():
 
--- a/perma_web/tasks/dev.py
+++ b/perma_web/tasks/dev.py
@@ -1119,8 +1119,8 @@ def get_and_categorize_duplicative_users():
             registrar_and_org_mix.add(normalized_email)
 
     exclude_group = any_paid_history | registrar_and_org_mix
-    logger.info(f"MERGING: Found {len(any_paid_history)} users who have purchased subscriptions or bonus links.")
-    logger.info(f"MERGING: Found {len(registrar_and_org_mix)} users who have accounts associated with both registrars and orgs.")
+    logger.warning(f"MERGING: Found {len(any_paid_history)} users who have purchased subscriptions or bonus links.")
+    logger.warning(f"MERGING: Found {len(registrar_and_org_mix)} users who have accounts associated with both registrars and orgs.")
 
     #
     # Organize remaining users by how many accounts associated with their email address have been confirmed.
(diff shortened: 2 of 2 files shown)


perma_web/perma/models.py
```diff
@@ -1109,6 +1109,64 @@ def can_view_usage_plan(self):
...
+        orgs = set()
+        registrars = set()
+        for user in users:
+            if user.registrar_id:
+                registrars.add(user.registrar_id)
+            else:
+                orgs.update(user.organizations.all())
+
+        if orgs or registrars:
+            assert not (orgs and registrars), f"This set of users includes both org and registrar users: {', '.join([user.id for user in users])}."
```
rebeccacremona: Surely that `join` will fail: cast the id to `str`

perma_web/perma/models.py
```diff
@@ -1109,6 +1109,64 @@ def can_view_usage_plan(self):
...
+            if registrars:
+                assert len(registrars) == 1
```
rebeccacremona: I'm using `assert`s throughout; I should catch and record in on of the higher level functions, instead of just borking.

perma_web/perma/models.py
```diff
@@ -1109,6 +1109,64 @@ def can_view_usage_plan(self):
...
+                    self.registrar_id = new_registrar_id
+                    self.prepend_to_notes(f"Added registrar during the merging of accounts: {new_registrar_id}")
+                    self.save(update_fields=['registrar_id', 'notes'])
+            else:
+                if original_orgs != orgs:
+                    self.prepend_to_notes(f"Added organizations during the merging of accounts: {', '.join([str(o.id) for o in orgs - original_orgs])}")
+                    self.save(update_fields=['notes'])
+                    self.organizations.add(*orgs)
+
+    def soft_delete_after_merge_with_user(self, user):
```
rebeccacremona, line 1144: Does this want to be a model method? I am doing the reverse inline in the task... and using bulk operations for increased efficiency. Given that we will only do this once (🤞) and it only takes seconds, I think not worth giving it any thought but.....

Might as well keep it: easier to reference later when making a generalized soft-delete method.

perma_web/tasks/dev.py
```diff
@@ -701,3 +707,496 @@ def populate_folder_cached_path(ctx, batch_size=500):
...
+        writer = csv.writer(csvfile, delimiter='|')
+        writer.writerow(['user id', 'original email', 'normalized email', 'merged with accounts', 'merged at'])
+
+
+def merge_accounts(
+        to_keep,
+        to_delete,
+        copy_memberships=True,
+        transfer_links=False):
+    from perma.models import Link
```
rebeccacremona: Why do we do this with imports in these files? Is it for a reason, or is it a relic of intentions whose shipped has sailed (for instance, a thing for tasks that don't require django to be configured.... or to avoid circular imports in some specific circumstances)?

perma_web/tasks/dev.py
```diff
@@ -701,3 +707,496 @@ def populate_folder_cached_path(ctx, batch_size=500):
...
+
+
+def unmerge_accounts(from_user_id):
+    from perma.models import Link, LinkUser
+
+    user = LinkUser.objects.get(id=from_user_id)
+    if match := re.search(r"\n*Merged with (?P<user_ids>.+)", user.notes):
+        user_ids = [int(uid) for uid in match.group('user_ids').split(', ')]
+    else:
+        print(f"Found no accounts to unmerge from user {from_user_id}.")
```
rebeccacremona: Let's check where these prints will go during the prod run, and see if they would be better written to another report, or written to the application log.

perma_web/tasks/dev.py
```diff
@@ -701,3 +707,496 @@ def populate_folder_cached_path(ctx, batch_size=500):
...
+            merged_with,
+            timezone.now()
+        ])
+
+
+def unmerge_accounts(from_user_id):
+    from perma.models import Link, LinkUser
+
+    user = LinkUser.objects.get(id=from_user_id)
+    if match := re.search(r"\n*Merged with (?P<user_ids>.+)", user.notes):
```
rebeccacremona, line 793: I am enjoying `:=` with `re.search`!

perma_web/tasks/dev.py
```diff
@@ -701,3 +707,496 @@ def populate_folder_cached_path(ctx, batch_size=500):
...
+                from_user_id = int(row['merged with user id'])
+
+                user = LinkUser.objects.get(id=int(row['user id']))
+                user.email = original_email
+                user.is_active = user.is_confirmed
+                user.link_count = int(row['org links transferred']) + int(row['personal links transferred'])
+
+                if match := re.search(r"\n*Original registrar: (?P<registrar_id>\d+)", user.notes):
+                    user.registrar_id = int(match.group('registrar_id'))
+                    user.remove_line_from_notes('Original registrar')
```
rebeccacremona, line 854: Check again that this stanza works.

perma_web/tasks/dev.py
```diff
@@ -701,3 +707,496 @@ def populate_folder_cached_path(ctx, batch_size=500):
...
+                from_user.prepend_to_notes(f"Extracted previously merged user { user.id }.")
+                from_user.save(update_fields=['notes'])
+                from_users.add(from_user)
+
+        LinkUser.objects.bulk_update(to_reverse, ['email', 'is_active', 'link_count', 'registrar_id', 'notes'])
+        for user in from_users:
+            user.refresh_from_db()
+            if match := re.search(r"\n*Added registrar during the merging of accounts", user.notes):
+                user.registrar_id = None
+                user.remove_line_from_notes('Added registrar')
```
rebeccacremona, line 875: Check again that this stanza works too.

perma_web/tasks/dev.py
```diff
@@ -701,3 +707,496 @@ def populate_folder_cached_path(ctx, batch_size=500):
...
+def merge_users_with_multiple_confirmed_accounts_but_no_links(user_list):
+    """
+    Select the account they have logged into most recently, or if they
+    have never logged in, the most recently created confirmed account.
+    Then sync all the registrars/orgs to it and delete the other ones.
+    """
+    to_keep = None
+    to_delete = []
+
+    if all(u.last_login for u in user_list):
```
rebeccacremona: This is wrong. This keeps the account they have logged into most recently only if they have logged into all their accounts; else, goes with the most recent. I think it wants to be `any`, rather than `all`. Give it another think later with a fresh brain.

perma_web/tasks/dev.py
```diff
@@ -701,3 +707,496 @@ def populate_folder_cached_path(ctx, batch_size=500):
...
+    grouped_duplicative_users = defaultdict(list)
+
+    start = time.time()
+    count = 0
+    for user in duplicative_users:
+        count += 1
+        grouped_duplicative_users[user.email.lower()].append(user)
+    end = time.time()
+
+    print(f"Data collected in {end - start} seconds.")
```
rebeccacremona: Maybe can remove some or all these timing checks, now that we know the order of magnitude.

perma_web/tasks/dev.py
```diff
@@ -701,3 +707,496 @@ def populate_folder_cached_path(ctx, batch_size=500):
...
+                any_paid_history.add(normalized_email)
+            if user.registrar_id:
+                registrar = True
+            if user.organization_ids:
+                orgs = True
+        if registrar and orgs:
+            registrar_and_org_mix.add(normalized_email)
+
+    end = time.time()
+    exclude_group = any_paid_history | registrar_and_org_mix
```
rebeccacremona, line 1085: Let's bark more strenuously if there's anybody in the `exclude_group`

perma_web/tasks/dev.py
```diff
@@ -701,3 +707,496 @@ def populate_folder_cached_path(ctx, batch_size=500):
...
+    with open(RETAINED_USERS_CSV, 'r', newline='') as csvfile:
+        csv_reader = csv.DictReader(csvfile, delimiter='|')
+        for row in csv_reader:
+            unmerge_accounts(row['user id'])
+
+@task
+def assert_no_duplicative_accounts(ctx):
+    from perma.models import LinkUser
+    duplicative_users = LinkUser.objects.raw(DUPLICATIVE_USER_SQL)
+    assert not len(duplicative_users)
```
rebeccacremona, line 1170: `assertn't`?

perma_web/perma/models.py
rebeccacremona: Nah.

perma_web/tasks/dev.py
```diff
@@ -701,3 +666,491 @@ def populate_folder_cached_path(ctx, batch_size=500):
...
+        merge_accounts(to_keep, to_delete)
+    else:
+        to_keep = most_recently_logged_into_account
+        to_delete = list(filter(lambda u: u is not to_keep, user_list))
+        merge_accounts(to_keep, to_delete, transfer_links=True)
+
+
+def merge_users_with_multiple_accounts_with_links(user_list):
+    """
+    We should move all links into the account with the most recent login,
```
matteocargnelutti: Not only we should, but we're doing just that 😄

matteocargnelutti (approved):
This is great, @rebeccacremona! 
As briefly discussed on Slack, I think using `invoke` for this makes a lot of sense. 

---

> I would appreciate a second set of eyes on the logic of the functions that select which user in a given group is the right one to "keep"/be the target of the merge:
> - `merge_users_with_only_unconfirmed_accounts`
> - `merge_users_with_only_one_confirmed_account`
> - `merge_users_with_multiple_confirmed_accounts_but_no_links`
> - `merge_users_with_only_one_account_with_links`
> - `merge_users_with_multiple_accounts_with_links`
> 
> Do they in fact do what their docstrings claim? I think so! But, if I have a mistake, I think it's most likely to be there.

Based on my understanding, I also believe they do 👍 !

perma_web/tasks/dev.py
rebeccacremona: LOL whoops.

PR #3324, opened 2023-04-19
Add CSP to replay.perma.cc

matteocargnelutti:
This PR adds a fairly _"loose"_ Content Security Policy to replay.perma.cc.

The intended policy can be summarized as: 
```
Everything's allowed within the <iframe>, as long as it's same-origin.
```

The goal is to help prevent potential _"service worker leaks"_ (meaning: requests meant to be intercepted by the service worker that end up reaching the network) that the playback software might not have accounted for yet by using browser-level safety mechanisms. 

---

**Note:** Will require extensive testing on staging before pushing to prod.

### DIFF
--- a/perma_web/replay/views.py
+++ b/perma_web/replay/views.py
@@ -34,7 +34,11 @@ def iframe(request):
 
     response["Clear-Site-Data"] = '"cache", "storage"'
     response["Cache-Control"] = "no-cache, no-store, must-revalidate"
-    
+
+    # Intended CSP Policy:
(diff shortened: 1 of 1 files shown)


rebeccacremona (approved):
Gave this a kick locally too; I too didn't see any problems, but my local collection isn't as extensive as it once was. Staging has a lot of already-broken captures (since making a capture there is often the first thing we do when a user reports a problem), so if we're planning to really kick the tires there, let's find GUIDs that are known to work before deploying :-)

matteocargnelutti:
Good call, thanks @rebeccacremona !

PR #3328, opened 2023-04-26
Make logins case-insensitive

### LINKED ISSUE #3152: Case-insensitive email addresses

rebeccacremona:
This PR makes login case-insensitive, and ensures at the database level that no "duplicative" accounts can exist. It cannot be deployed until the existing duplicative accounts are merged, using the code in [the former PR](https://github.com/harvard-lil/perma/pull/3320) and a small amount of manual clean up.

- It adds a new field to our user model, "raw_email", and copies over the current contents of the "email" column. 
- Then, it downcases the "email" column and adds a `UNIQUE lower(email)` index to the DB so that, even if something goes wrong at the application layer, no more duplicative accounts can be created. 
- It arranges for the login and password reset forms to downcase the submitted email field during the `clean` phase, so that logins and password resets are case-insensitive. 
- When admins/registrar users/org users add people to groups, it downcases the email they enter, so that the check-for-existing-users step is case-insensitive, and so that, if no existing user is found, the new user is created with a lowercase email address.
- When people sign up by themselves, the email they enter is saved in `raw_email` as is, and a lowercase version is saved in `email`.
- When Perma displays the email address or when it sends mail, it uses the `raw_email` field.

The unit tests are fairly extensive, so I am pretty confident this works as intended. 

However, there is definitely the possibility that I am missing something. Can we think of anything I'm missing? Can we think of anything this might break that wouldn't normally be covered by a test suite?

Tentatively closes https://github.com/harvard-lil/perma/issues/3152

### NB

This code should not be merged until we are ready to deploy it, just in case.

Before we deploy it, Ben and I intend to contribute a few more bits of code:
- a tweak to the account merging code of the aforementioned PR
- code that will email the merged users, after the fact
- some Salt excitement to make the deployment smoother

I am not marking it a draft, in order to request reviews.


rebeccacremona:
I tested the data migrations locally with a prod-like dump, and [found it to be zippy](https://github.com/orgs/harvard-lil/projects/11/views/1?pane=issue&itemId=25019084): like 7 or 8 seconds.

perma_web/perma/models.py
```diff
@@ -848,8 +852,31 @@ class LinkUser(CustomerModel, AbstractBaseUser):
     class Meta:
         verbose_name = 'User'
 
+        constraints = [
```
matteocargnelutti, line 855: This is a great comment

rebeccacremona:
Note that `email` isn't included in [API results](https://github.com/harvard-lil/perma/blob/develop/perma_web/api/serializers.py#L38), which means we don't have to worry about `email`/`raw_email` in the JS or anywhere upstream.

perma_web/perma/models.py
rebeccacremona: Thanks!

perma_web/perma/views/user_management.py
```diff
@@ -1357,7 +1361,7 @@ def limited_login(request, template_name='registration/login.html',
     """
 
     if request.method == "POST" and not request.user.is_authenticated:
-        username = request.POST.get('username')
+        username = request.POST.get('username', '').lower()
         try:
             target_user = LinkUser.objects.get(email=username)
```
matteocargnelutti, line 1366: This type of scenario always makes me unnecessarily nervous, and I tend to throw before doing a db lookup if `username` is not in the format I expect. Mostly a druther.

matteocargnelutti (approved):
💯

bensteinberg (approved):
This is amazing!

perma_web/perma/views/user_management.py
rebeccacremona: This makes so much sense. You know what? I want to audit the whole app, and see where else we do this. I think I'll make a separate ticket rather than addressing here, since as is, this basically keeps the same behavior as before (switching to looking up `''` instead of `None`)

PR #3339, opened 2023-05-25
Add a script for initializing a new local db from a db dump 

### LINKED ISSUE #3314: Add a script for initializing a new local db from a db dump
We do this often enough now that it would be convenient to script it. Something similar to https://github.com/harvard-lil/h2o/blob/develop/docker/init.sh#L47-L49

rebeccacremona:
closes https://github.com/harvard-lil/perma/issues/3314


matteocargnelutti (approved):
Nice!

Issue #2483, opened 2018-09-12
Estimation of queue time off


rebeccacremona:
![image](https://user-images.githubusercontent.com/11020492/45439136-04131600-b687-11e8-8cfe-9d15c3e04930.png)

rebeccacremona:
Way to go, 2018 Becky, for not bothering to explain how you made that happen......

rebeccacremona:
Aha! To reproduce locally, run with `CELERY_TASK_ALWAYS_EAGER = False`, but don't have any workers processing jobs.

rebeccacremona:
Interesting, I think [this line to find out how many workers are running](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/settings/utils/post_processing.py#L97-L102) is not working particularly well, even in prod:
if you inspect the source of /manage/create, you'll see something like
```
        host = "perma.cc",
        workers = 2,
        average = 8.90523;
```
There are a lot more than 2 workers available. Maybe only two workers were ready when uWsgi started? I bet that's it. Locally, I'm always getting 0, no matter how many workers I'm running.

I'll make a quick UI fix for the "Infinity" case (which should never happen except locally)... and we can decide if we want to keep this as part of a more involved UI discussion or not.

PR #3343, opened 2023-05-26
If we can't estimate queue time, just say "Queued."

rebeccacremona:
Fixes it so that we no longer say a capture is "Queued: begins in Infinity minutes." if our count of celery workers returns `0`.

Addresses part of https://github.com/harvard-lil/perma/issues/2483


bensteinberg (approved):
"infinity minutes" is my favorite amount of minutes

PR #3351, opened 2023-06-12
Make a stand-alone form for reporting inappropriate content

rebeccacremona:
Currently, when a user clicks on the “Flag Inappropriate” button on the “Record Details” tray of a Perma Link they are redirected to our [Contact Form](https://perma.cc/contact?flag=T9KH-SZNL) with a pre-filled message.

We decided to extract that into its own form for special handling: see ENG-96 for more discussion about details and design.

### Flow

This PR adds a form with two new fields: a select/dropdown of reasons for reporting, and a text field requesting info on where the user found the Perma Link. 

image
image

If the reporting user is logged in, their email address is pre-populated for them.

If required fields are missing, users are prompted to fill them in, but otherwise, they are directed to the "Thanks!" page.

image

Perma admins are sent an email something like:
```
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit
Subject: [perma-contact] Reporting Inappropriate Content
From: webmaster@localhost
To: webmaster@localhost
Date: Mon, 12 Jun 2023 20:33:16 -0000
Message-ID: <168660199633.1417.8184645293545937667@d807b0a87676>
Reply-To: me@me.com


A user has reported that CMU9-DXVG contains inappropriate material.

https://perma.test:8000/admin/perma/link/?guid=CMU9-DXVG


Reason for Reporting
--------------------
Graphic or Dangerous Content

How did you discover this Perma Link?
-------------------------------------
I saw it on Internet Archive.

----
Troubleshooting Info:
User email: me@me.com
Logged in: false
Affiliations: (none)
Referring Page: https://perma.test:8000/CMU9-DXVG
User agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36
```

### Design decisions

I chose _not_ to have the form perform any checks to see if the GUID in question exists. While no more expensive than a regular 404, I wanted to minimize database usage, should a bot slip past our protections and make a gazillion submissions. But, I only send the email if _something_ is supplied for GUID.

I'm not that worried about it, but it could be that we'd rather check for the existence of the link instead.

### DIFF
--- a/perma_web/perma/forms.py
+++ b/perma_web/perma/forms.py
@@ -2,6 +2,7 @@ from axes.utils import reset as reset_login_attempts
 
 from django import forms
 from django.contrib.auth.forms import SetPasswordForm
+from django.db.models.fields import BLANK_CHOICE_DASH
 from django.forms import ModelForm
 from django.http import HttpResponseRedirect
 from django.urls import reverse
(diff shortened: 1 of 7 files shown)


bensteinberg (approved):
👍

rebeccacremona:
With the optional "message" field removed:
image

PR #3353, opened 2023-06-14
Use WACZ-exhibitor for playback in Perma (behind a feature flag)

rebeccacremona:
See ENG-99.

[wacz-exhibitor](https://github.com/harvard-lil/wacz-exhibitor) was built to help websites embed WARC and WACZ replays in the [best way possible](https://lil.law.harvard.edu/blog/2022/09/15/opportunities-and-challenges-of-client-side-playback/): we’d do well to use it in Perma! 

Our proposed move to WACZ would allow us to fully benefit from its caching smarts, but we will benefit a bit even while Perma is still primarily serving WARC playbacks.

## This PR

This PR adds the new Django setting `PLAYBACK_HOST`, set to `None` by default. 

If defined, the application will send playbacks to that host instead of to the [`replay` Django app](https://github.com/harvard-lil/perma/tree/develop/perma_web/replay) currently served at a subdomain.

Locally,
- run `docker compose up -d` and then `bash make_cert.sh` to prepare your environment, then
- uncomment `PLAYBACK_HOST = 'rejouer.perma.test:8080'` in settings_dev.py

to direct playbacks to an nginx container serving a [slightly tweaked version of `wacz-exhibitor`](https://github.com/rebeccacremona/wacz-exhibitor/commit/9bc60b6bd477a28bb0e04d96fbd7c1d2fd01e739).

You can toggle back and forth without restarting the Django dev server.

Most playbacks should look identical.

### Interstitials

While working to re-implement the [custom interstitials](https://github.com/harvard-lil/perma/pull/3036) that handle any playbacks that should trigger downloads... I discovered that they are currently broken 😢: Safari users do not receive a usable download, and other users receive multiple copies. (Bug to be written up shortly.)

So, this PR does NOT re-implement the more complex interstitials we are currently using in prod, which were introduced specifically for Safari.

In production, when a user clicks the "View/Download File" button, we pry into the `replay-web-page` instance and find the internal URL for the payload. Then, we insert an `<a>` into the DOM with that as `href` and click it. That used to reliably trigger a download in all browsers... but alas: in Safari, that network request is NOT intercepted by the service worker, and falls through to the live web, where it either 404s (`replay.perma.cc`) or 500s (`wacz-exhibitor`). 

I cannot think of any further workarounds... So, I just do the simpler thing, our original strategy: don't stick the playback iframe into the page until the user clicks the "View/Download File" button. That reliably triggers one download in all browsers tested except for Safari.

We probably want to talk about this some more... but not as part of this PR.

### Fallback for old browsers

In production, `replay.perma.cc` is responsible for rendering an "update your browser!" message, if a person with a very old browser loads a Perma Link.

When playbacks are mediated by `wacz-exhibitor`, the perma app takes care of it. (TBD: whether that's something that ought to be handled by `wacz-exhibitor` instead; I think probably, but could be wrong!)

I didn't use an identical strategy: this code will attempt a playback for every browser that can read `script type=module`. That might not be good enough 🤷‍♀️ .... but I honestly don't know. Close enough for jazz? If not, we can readdress in a future PR.

This PR arranges for users to see the "update your browser!" message even when playback is mediated by an interstitial; with `replay.perma.cc` playbacks, the message is hidden behind the interstitial.

### Deploying

I think this is safe to merge as-is, it should be a no-op without a setting of the feature flag, except for some trivial changes to interstitial CSS.

### DIFF
--- a/perma_web/perma/templates/archive/download_to_view_pdf.html
+++ b/perma_web/perma/templates/archive/download_to_view_pdf.html
@@ -1,5 +1,5 @@
 <div id="playback-interstitial" class="record-message">
-  <p class="hidden record-message-primary">Perma.cc can't display this file type on mobile, but you can view or download the archived file by clicking below.</p>
+  <p class="hidden record-message-primary">Perma.cc can’t display this file type on mobile, but you can view or download the archived file by clicking below.</p>
   <p class="hidden record-message-secondary">File type {{mime_type}}</p>
   {% if PLAYBACK_HOST %}
     <button id="download-button" class="hidden btn btn-primary">View/Download File</button>
--- a/perma_web/perma/templates/archive/download_to_view_pdf.html
+++ b/perma_web/perma/templates/archive/download_to_view_pdf.html
@@ -1,7 +1,14 @@
-<div class="record-message">
-  <p class="record-message-primary">Perma.cc can't display this file type on mobile, but you can view or download the archived file by clicking below.</p>
-  <p class="record-message-secondary">File type {{mime_type}}</p>
+<div id="playback-interstitial" class="record-message">
+  <p class="hidden record-message-primary">Perma.cc can't display this file type on mobile, but you can view or download the archived file by clicking below.</p>
+  <p class="hidden record-message-secondary">File type {{mime_type}}</p>
+  {% if PLAYBACK_HOST %}
+    <button id="download-button" class="hidden btn btn-primary">View/Download File</button>
+    <div>
+      {% include "archive/includes/wacz_exhibitor_iframe.html" with interstitial=True  %}
+    </div>
+  {% else %}
     <div>
       {% include "archive/includes/client_side_iframe.html" with interstitial=True target="blank" %}
     </div>
+  {% endif %}
 </div>
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -106,6 +106,26 @@ services:
           - 'perma.minio.test'
 
   #
+  # wacz-exhibitor
+  #
+  wacz-exhibitor:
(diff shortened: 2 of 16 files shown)


perma_web/perma/templates/archive/download_to_view_pdf.html
```diff
@@ -1,7 +1,14 @@
-<div class="record-message">
-  <p class="record-message-primary">Perma.cc can't display this file type on mobile, but you can view or download the archived file by clicking below.</p>
-  <p class="record-message-secondary">File type {{mime_type}}</p>
+<div id="playback-interstitial" class="record-message">
+  <p class="hidden record-message-primary">Perma.cc can't display this file type on mobile, but you can view or download the archived file by clicking below.</p>
```
bensteinberg: To match the one in `templates/archive/iframe.html`:
```suggestion
  <p class="hidden record-message-primary">Perma.cc can’t display this file type on mobile, but you can view or download the archived file by clicking below.</p>
```

bensteinberg (approved):
Looks great. FWIW, `PLAYBACK_HOST` is the setting we used to use for webrecorder, so I'll have to ensure it is completely absent from deployment config.

rebeccacremona:
@bensteinberg can you help me find [this changed character](https://github.com/harvard-lil/perma/pull/3353/commits/fa710e4d1082f93059705ea77858e3275a308334) in [this test](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/tests/test_views_common.py#L181)?

bensteinberg:
Lines 188 and 196, I think

matteocargnelutti:
👏 👏 👏 🚀

PR #3355, opened 2023-06-15
Remove replay app (that was replaced by wacz-exhibitor)

rebeccacremona:
See ENG-124.

We are now comfortable removing the `replay` app: playback via wacz-exhibitor seems to be going great!


bensteinberg (approved):
LGTM! I think there's one more mention of replay.perma.test to remove at the bottom of `install.md`.

rebeccacremona:
Thank you, `replay` app, for your service!

PR #3360, opened 2023-06-23
Update wacz-exhibitor

rebeccacremona:
See ENG-143 and ENG-170.

This PR updates Perma's dev environment to use the latest version of `wacz-exhibitor` and makes the recommended changes to `nginx.conf`.

It also overhauls the `wacz-ehibitor` docker-compose situation... which I think I basically did entirely incorrectly the first time 😂 . I'm not sure what went wrong there... If I got this right, https://github.com/harvard-lil/docker-compose-update-action will now manage the building, tagging, and pushing of the image to the registry. But honestly I'm a little 🤯 about the whole thing; this syntax might not be exactly right.


rebeccacremona:
So far so good... although I guess I should have done `0-nohash` instead of `1-nohash`... Do we care?
image

bensteinberg (approved):
Let's give it a whirl!

PR #3362, opened 2023-06-23
remove partners list and map

clare-stanton:
Removing partners list and map from About page because it is outdated and would be visually overcrowded and not useful if it we were to fully update it.


rebeccacremona (approved):
Hooray, Clare 🎉 🎉 🎉 !

What a fun way to end the week.

I think this is good to go! We can think about whether to delete the database column, clean up the CSS etc. some other time :-)

We have never removed a JS bundle from webpack config before, so we'll just have to see how that plays out during deployment 😂

rebeccacremona:
Ah, thank you, tests, for failing: @clare-stanton the landing page also has a registrar logo carousel, we did not remember. I will remove it too.

image

PR #3363, opened 2023-06-26
Remove map JS

bensteinberg:
This is a follow-up to #3362.


rebeccacremona (approved):
I think probably this will, on merge, cause another commit of built assets, with an updated webpack-stats.json

rebeccacremona:
My above prediction was, of course, wrong: since we DID remove the entrypoint from webpack config in the first PR, the change to the stats file subsequent to that PR was sufficient

PR #3369, opened 2023-07-05
Update Django

bensteinberg:
https://docs.djangoproject.com/en/dev/releases/3.2.20/


bensteinberg:
In connection with another error, I notice https://docs.djangoproject.com/en/3.2/ref/validators/#django.core.validators.URLValidator.max_length -- which is new in 3.2.20, and defaults to 2048 characters.

matteocargnelutti (approved):

PR #3371, opened 2023-07-07
Temporary workaround for mobile PDF interstitials

matteocargnelutti:
@kmukk reported that the mobile interstitial for PDFs isn't working as expected: nothing happens on click.

After looking into the issue, it appears that this might be due to the way we interact with replayweb.page in that context: most definitely TBD, and this will require a proper diagnosis.

In the meantime: since it appears that we can now render PDFs within `<replay-web-page>` on mobile, in concertation with @kmukk and @bensteinberg, we've decided to disable the interstitial for the time being.

Users _should_ therefore now see the PDFs on mobile instead of the interstitial, and still be able to download them if they so choose.

### DIFF
--- a/perma_web/perma/tests/test_views_common.py
+++ b/perma_web/perma/tests/test_views_common.py
@@ -1,7 +1,7 @@
 from django.conf import settings
 from django.core import mail
 from django.urls import reverse
-from django.test import override_settings, Client
+from django.test import override_settings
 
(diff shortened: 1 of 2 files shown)


bensteinberg (approved):
:+1: pending tests

matteocargnelutti:
On it 😉

PR #3388, opened 2023-09-01
Free disk space in CI

bensteinberg:
This is an attempt to free disk space so the GitHub actions runner does not run out; we've seen image build failures like
```
ERROR: failed to solve: failed to create temp dir: mkdir /tmp/containerd-mount4076981537: no space left on device
```
 I'm using the default setup from https://github.com/marketplace/actions/free-disk-space-ubuntu -- we can do something like this manually so as not to have the dependency, if that seems better.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,6 +11,23 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      ### free disk space ###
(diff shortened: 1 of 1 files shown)


rebeccacremona (approved):
Neat! I'll note that this is probably my fault, and I should change the Dockerfile to build a cleaner image.

PR #3389, opened 2023-09-01
Free disk space tweak

bensteinberg:
This is a tweak to #3388, saving 3+ minutes in testing when we don't need to build and push an image. It also fixes a typo I happened to notice while watching output from CI.


rebeccacremona (approved):
Ha, thanks for the typo fix!

PR #3392, opened 2023-09-05
Revised private by policy

rebeccacremona:
See ENG-295.

We have decided to make it simpler for domains to request that new Perma Links of their site be made private. Instead of inspecting robots.txt, the page's `meta` tags, and the requests' headers, we will instead compare the domain of the captured URL to a list, and will add domains to the list (`settings.PRIVATE_BY_POLICY_DOMAINS`) upon request, according to policy.

This PR renames the "keys" of the `Link.private_reason` to be descriptive, rather than "new/old", and adds two new ones: `domain`, for use with the new policy, and `flagged`, for use when a user has reported that a Perma Link contains flagged content (which admins sometimes recorded as private due to `policy`). The `manage/stats` page is updated accordingly.

The data migration is speedy locally with a prod-like database; I'm not worried about it.


perma_web/perma/migrations/0028_auto_20230905_1813.py
```diff
@@ -0,0 +1,58 @@
...
+
+from django.db import migrations, models
+
+
+def fixup_private_reason(apps, schema_editor):
+    """
+    Instead of 'policy' and 'old_policy', use descriptive keys
+    """
+    Link = apps.get_model('perma', 'Link')
+    print("Migrating 'old_policy' to 'meta_generic'.")
```
bensteinberg:
```suggestion
    print("Migrating 'old_policy' to 'meta'.")
```
unless you meant to use `meta_generic`?

perma_web/perma/migrations/0028_auto_20230905_1813.py
```diff
@@ -0,0 +1,58 @@
...
+    print("Migrating 'policy' to 'meta_perma'.")
+    Link.objects.filter(
+        private_reason='policy'
+    ).update(
+        private_reason='meta_perma'
+    )
+
+def reverse_fixup_private_reason(apps, schema_editor):
+    Link = apps.get_model('perma', 'Link')
+    print("Migrating 'meta_generic' to 'old_policy'.")
```
bensteinberg:
```suggestion
    print("Migrating 'meta' to 'old_policy'.")
```

bensteinberg (approved):

perma_web/perma/migrations/0028_auto_20230905_1813.py
rebeccacremona: Thanks. A first draft of this used longer key names... but it turned out that migrating the column to allow more than 10 chars was expensive, so I decided not worth it.

PR #3393, opened 2023-09-08
Tests pass locally with CAPTURE_ENGINE = 'scoop-api'

rebeccacremona:
Differences: 
- Scoop doesn't capture Flash's swf files
- Scoop doesn't capture as many favicons (we knew this, and are fine with it)
- Scoop doesn't capture the fallback `src` for images that use `srcset`: TBD.

I have not toggled `CAPTURE_ENGINE = 'scoop-api'` in settings_testing.py: CI will still test with Perma capture, for the time being.

### DIFF
--- a/perma_web/api/tests/utils.py
+++ b/perma_web/api/tests/utils.py
@@ -39,6 +39,9 @@ def copy_file_or_dir(src, dst):
 
 
 def index_warc_file(warc_file):
+    # Note: warcio will NOT locate the HTTP headers of WARC response records
+    # who use the file:/// protocol.... which is to say, Scoop attachments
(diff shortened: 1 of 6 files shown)


perma_web/api/tests/utils.py
```diff
@@ -39,6 +39,9 @@ def copy_file_or_dir(src, dst):
 
 
 def index_warc_file(warc_file):
+    # Note: warcio will NOT locate the HTTP headers of WARC response records
```
matteocargnelutti, line 42: 🔍  !

matteocargnelutti (approved):
🚀

PR #3398, opened 2023-09-26
Add a few Scoop-specific tests

rebeccacremona:
I could imagine this being more thorough, but I think this should be enough to get us started.

I *do* enjoy playing with mocks.

Note that these do not currently run in CI: we have not yet toggled CI to test against Scoop.

See ENG-320.

### DIFF
--- a/perma_web/api/tests/test_link_resource.py
+++ b/perma_web/api/tests/test_link_resource.py
@@ -82,7 +82,7 @@ class LinkResourceTestMixin():
             'private_reason',
         ]
 
-    def assertRecordsInWarc(self, link, upload=False, expected_records=None, check_screenshot=False):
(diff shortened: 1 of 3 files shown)


perma_web/api/tests/test_link_resource.py
```diff
@@ -686,3 +709,132 @@ def test_should_accept_spaces_in_url(self):
         link = Link.objects.get(guid=obj['guid'])
         self.assertEqual(link.capture_job.status, 'completed')
         self.assertEqual(link.primary_capture.status, 'success')
+
+    def test_custom_title_not_overridden(self):
+        custom_title = 'The spiciest of test pages'
```
bensteinberg, line 714: Nice!

bensteinberg (approved):
LGTM!

PR #3401, opened 2023-10-04
Dependency updates

bensteinberg:
https://urllib3.readthedocs.io/en/latest/changelog.html#id9 (this link is unlikely to be accurate once there's a newer version)

https://pillow.readthedocs.io/en/stable/releasenotes/10.0.1.html and see also https://pillow.readthedocs.io/en/stable/releasenotes/10.0.0.html -- this is a major version upgrade.


bensteinberg:
https://docs.djangoproject.com/en/dev/releases/3.2.22/

rebeccacremona (approved):
LGTM! As with H2O PR, apologies for the slow review.

Now that the logo carousel is gone, we use pillow even less. After this is on stage, I think I will manually test an upload-your-own link of supported and unsupported formats, and a replacement link, just to be extra careful. (Those things are covered in the test suite, but just ot be thorough :-)

PR #3415, opened 2023-10-26
Refactor test_capture_job.py

rebeccacremona:
See ENG-415.

This PR refactors test_capture_job.py:

a) to use pytest fixtures powered by factory boy instead of our [json fixtures](https://github.com/harvard-lil/perma/tree/develop/perma_web/fixtures)

b) to use standard pytest syntax instead of the older [unittest-style syntax](https://docs.pytest.org/en/7.1.x/how-to/unittest.html)

This was easy, because I basically already went through this exercise in https://github.com/harvard-lil/perma-capture/commit/b2fe7addb5c4d8c19769e4b552516726b3829761 and related commits, when building the prototype capture service.

I hope to be able to reuse a lot of that work (as intended), as this refactoring project  proceeds!

### Reviewing

The unified diff is very hard to read, but I think if you click through the individual commits, you'll be able to easily see at a glance that the tests haven't been changed.

### DIFF
--- a/perma_web/api/tests/test_current_user_authorization.py
+++ b/perma_web/api/tests/test_current_user_authorization.py
@@ -1,6 +1,13 @@
 from .utils import ApiResourceTestCase
-from perma.tests.test_capture_job import create_capture_job
-from perma.models import LinkUser, Link
+from perma.models import LinkUser, Link, CaptureJob
+
+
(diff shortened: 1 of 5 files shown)


matteocargnelutti (approved):
🎸

PR #3416, opened 2023-10-27
Update payments image

bensteinberg:


rebeccacremona (approved):
LGTM!

Sorry for the delay: I wanted to make sure the dev server was still working after the Werkzeug update (which you may recall needed followup in the Perma repo), and kept forgetting to check. But it works fine!

PR #3420, opened 2023-10-30
Toggle so that Scoop API capture is the default.

rebeccacremona:


bensteinberg (approved):

rebeccacremona:
Oh right, we run with `docker-compose -f docker-compose.yml up -d` in CI, and the Scoop service is currently in `docker-compose.override.yml`..... To be considered.

PR #3423, opened 2023-10-31
Tweak logging

rebeccacremona:
This is a very small PR to tweak our logging of captures made with the Scoop API.

1) We now store the Scoop job id in a dedicated database field, so that it is available even if the job never advanced [beyond this loop](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/celery_tasks.py#L1036) (one variety of error we have been debugging).

2) We log the Scoop job id and status while in the loop, to make it easier to match up Perma logs and Scoop API logs, when something goes wrong.


bensteinberg (approved):

matteocargnelutti:
🏄 !

matteocargnelutti (approved):

PR #3426, opened 2023-11-01
Update to the latest Scoop REST API in dev and tests

rebeccacremona:
I am not positive the CI move will work... If it doesn't, I will back out that change and address a bit later.


rebeccacremona:
Hooray! It worked!

bensteinberg (approved):
:metal:

PR #3427, opened 2023-11-02
Add capture errors panel to admin stats page.

rebeccacremona:
It should look like this, but with fewer zeroes :-)

image


matteocargnelutti (approved):
🎸

PR #3433, opened 2023-11-14
Check for Docker image updates

bensteinberg:
I'm not absolutely sure about this, but:

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#configuration-options-for-private-registries

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+registries:
+  lil-registry:
+    type: docker-registry
+    url: https://registry.lil.tools
+updates:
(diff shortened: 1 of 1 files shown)


rebeccacremona (approved):
Cool! 

I wonder if you have to tell it to look at both docker-compose.yml and docker-compose.override.yml or if it is smart enough.

I also notice something weird: look, we are using minio from Dockerhub. https://github.com/harvard-lil/perma/blob/develop/docker-compose.yml#L77. I wonder if we had some reason for doing so.... and whether that will mess anything up with this too.

PR #3434, opened 2023-11-15
Remove Perma's capture code (that was replaced by Scoop)

rebeccacremona:
See ENG-403

We are now comfortable removing Perma's own capture code: capture via Scoop seems to be going great!

I decided to keep the `CAPTURE_ENGINE` settings and leave `run_next_capture()` as is for now; let's discuss whether it is more proper/cleaner to remove it.

...

So long, code! You served us well!


bensteinberg (approved):
Amazing! Thanks!

Do you think we'll need to keep `CAPTURE_USER_AGENT` updated? Maybe, but maybe not often.

rebeccacremona:
> Do you think we'll need to keep CAPTURE_USER_AGENT updated? Maybe, but maybe not often.

I'm of mixed minds... I am unsure about the wisdom of continuing to have Perma validate the URL at all... it could be, it would be better to delegate that responsibility to the Scoop API. But we haven't talked about that yet, and whether that would be carving at the joints, or, rather, would be overloading that API.

Issue #3276, opened 2023-02-28
ORM-generated SQL for a particular `COUNT` is slow w/ Django 3.2


rebeccacremona:
After upgrading from Django 2.2 to 3.2, we noticed that [/manage/sponsored-users/](https://github.com/harvard-lil/perma/blob/v0.413/perma_web/perma/urls.py#L145), if visited by admins, caused an unmanageable spike in database CPU usage, repeatedly resulting in 502s across the board.

We eventually figured out that [this count](https://github.com/harvard-lil/perma/blob/236ac15f7a32e5e42296c59042c4f9594a68842a/perma_web/perma/views/user_management.py#L632) was generating different, much much slower SQL in Django 3.2:

```sql
--
-- Django 3.2
--
SELECT COUNT (*) FROM (
    SELECT DISTINCT 
        perma_linkuser.id,
        perma_linkuser.last_name
    FROM 
      perma_linkuser 
    WHERE 
      (
        NOT (
          EXISTS(
            SELECT 
              (1) AS a 
            FROM 
              perma_linkuser U0 
              LEFT OUTER JOIN perma_sponsorship U1 ON (U0.id = U1.user_id) 
            WHERE 
              (
                U1.registrar_id IS NULL 
                AND U0.id = perma_linkuser.id
              ) 
            LIMIT 
              1
          )
        ) AND NOT perma_linkuser.is_active 
        AND NOT perma_linkuser.is_confirmed
      ) 
    ORDER BY 
      perma_linkuser.last_name ASC
) subquery;


Time: 28117.552 ms (00:28.118)

--
-- Django 2.2
--
SELECT COUNT (*) FROM (
    SELECT DISTINCT 
      perma_linkuser.id,
      perma_linkuser.last_name
    FROM 
      perma_linkuser
    WHERE 
      (
        NOT (
          perma_linkuser.id IN (
            SELECT 
              U0.id 
            FROM 
              perma_linkuser U0 
              LEFT OUTER JOIN perma_sponsorship U1 ON (U0.id = U1.user_id) 
            WHERE 
              U1."registrar_id" IS NULL
          )
        ) 
        AND perma_linkuser.is_active = False 
        AND perma_linkuser.is_confirmed = False
      ) 
    ORDER BY 
      perma_linkuser.last_name ASC
) subquery;
 

Time: 42.158 ms
```

[...]

PR #3435, opened 2023-11-16
Retrieve sponsored users with a non-standard strategy for improved performance.

rebeccacremona:
See ENG-181 and https://github.com/harvard-lil/perma/issues/3276.

This page, when viewed as an admin, consistently gives us a hard time.

Every other page of this type reuses the `list_users_in_group` method... which just isn't working out here: the SQL is just not very efficient. So, this PR adds a custom method for retrieving sponsored users and handing them off to the template with the necessary related objects pre-fetched.

There is some duplication of code between `list_users_in_group` and the new method, but I think it's not worth it to restructure at this time: these views are likely being completely redesigned soon anyway.


bensteinberg (approved):
:+1:

PR #3438, opened 2023-11-30
Convert additional tests to pytest syntax

rebeccacremona:
In #3415, I began refactoring our python test suite to:

a) use pytest fixtures powered by factory boy instead of our [json fixtures](https://github.com/harvard-lil/perma/tree/develop/perma_web/fixtures)

b) use standard pytest syntax instead of the older [unittest-style syntax](https://docs.pytest.org/en/7.1.x/how-to/unittest.html)

This PR continues the effort, picking off some more of the easier-to-convert modules.

The entirety of `test_email`, `test_utils`, `test_cloudflare_cache`, `test_celery_tasks`, and `test_urls` are converted to pytest syntax, and are _mostly_ (but not entirely) updated to use pytest fixtures: I kicked the can in a couple places on the more complex ones, figuring they will be easier once we have a richer library of pytest fixtures prepared.

A big chunk of `test_models` is also converted and switched over to pytest fixtures. There is still lots to do, but I think this PR converted every test that involves Perma Payments. @kilbergr is going to take the lead on coverting the rest of that file, in a series of upcoming PRs 🙏 🙌.

The diffs are, again, hard to read, but: none of the tests are materially changed: all the same assertions are made, etc. The only thing that has been changed is syntax.

### DIFF
--- a/perma_web/conftest.py
+++ b/perma_web/conftest.py
@@ -277,3 +277,30 @@ class LinkFactory(DjangoModelFactory):
         ),
         no_declaration=None
     )
+
+
+### fixtures for testing utils ###
+
+@pytest.fixture
+def spoof_perma_payments_post():
+    data = {
+        'encrypted_data': {"timestamp": 1504884268.560902, "desired_field": "desired_field"},
+    }
+    assert 'encrypted_data' in data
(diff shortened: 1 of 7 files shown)


kilbergr (approved):
:shipit:

PR #3439, opened 2023-11-30
first foray into test conversion

kilbergr:
This converts tests in the organization and privacy defaults section and cached folders section.

### DIFF
--- a/perma_web/perma/tests/test_models.py
+++ b/perma_web/perma/tests/test_models.py
@@ -1145,6 +1145,135 @@ def test_get_bonus_packages(prep, mock_datetime, paying_user):
     ]
 
 
+#
+# Organization privacy
+#
+
+def test_new_orgs_are_public_by_default(db):
+    r = Registrar()
+    r.save()
+    o = Organization(registrar=r)
+    o.save()
+    r.refresh_from_db()
+    o.refresh_from_db()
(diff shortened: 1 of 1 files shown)


perma_web/perma/tests/test_models.py
```diff
@@ -1258,27 +1258,27 @@ def test_folders_cached_paths_updated_when_moved(self):
         f3_1 = Folder(parent=f3, name='3_1')
         f3_1.save()
         f3_1.refresh_from_db()
-        self.assertEqual(f'{f1.pk}-{f3.pk}', f3.cached_path)
-        self.assertEqual(f'{f1.pk}-{f3.pk}-{f3_1.pk}', f3_1.cached_path)
+        f'{f1.pk}-{f3.pk}' == f3.cached_path
```
bensteinberg: Do this line and some subsequent ones not require an `assert`?

rebeccacremona (changes requested):
Well done!

Tiny tweak: it looks like these are still methods of the `ModelsTestCase` class, rather than top-level functions. 

Would you mind moving them out of the class?

I think you could group them under two comment headings: "Org privacy" and "Cached folder paths" (or similar), like

```
#
# Related to Perma Payments
#
```

rebeccacremona (approved):

PR #3441, opened 2023-12-04
Track playwright errors

rebeccacremona:
See ENG-490.

We're sporadically seeing a particular error message from Playwright during Scoop captures: not frequently enough to be a cause for concern, but frequently enough that I'd like to count their occurrence, rather than considering them anomalous.

This PR adds that to the list of routine errors we are tracking this way.


matteocargnelutti (approved):
Nice!

bensteinberg (approved):

PR #3445, opened 2023-12-08
Remove unused utilities.

rebeccacremona:
We haven't been using these "convenience methods" or this convenience script for a while. Rather than update them to match current dev flows, I think it's better to just remove them.


matteocargnelutti (approved):
🧹

PR #3446, opened 2023-12-08
Resolve python warnings during tests

rebeccacremona:
I noticed that pytest was reporting a lot of warnings:

image

It turns out that fixing them was no big deal: just a few tweaks and package upgrades. 

There's only one left unresolved, a deprecation warning about a line in `warctools`, which we are still using to produce upload-your-own warcs:

image

That library is not under active development; if it becomes a problem before we switch to WACZ, we can figure out what we want to do. But in the meantime, I'm leaving it.


matteocargnelutti (approved):
🤘

PR #3448, opened 2023-12-11
Upgrade from Django 3.2 to 4.2

rebeccacremona:
Support for Django 3.2 (LTS) will end in April. This PR upgrades us to the next LTS version, 4.2, which will be supported until April 2026.

Default storage and staticfile storage is configured significantly differently in 4.2: instead of a lot of stand-alone settings, everything is configured via key-value pairs in a big dict. So, the settings for stage and prod will have to be reconfigured before deployment.

I had to update a number of supporting packages for compatibility and to resolve deprecation warnings. 

The only remaining warnings are the one we already had, from [warcprox](https://github.com/harvard-lil/perma/pull/3446), and one that I think happens merely because we are patching `django.utils.timezone`: simply mocking it is seems to be enough to trigger the deprecation warning. I think that will go away by itself during the next upgrade.


bensteinberg (approved):
🤘

PR #3449, opened 2023-12-11
Remove unused image-related code and dependencies

rebeccacremona:
The carousel of registrar logos and list of "partners" was [recently removed](https://github.com/harvard-lil/perma/pull/3362) from Perma's landing and about pages.

This PR follows up and removes the remaining code and dependencies that only supported that feature.

That includes legacy "thumbnail"-related code and dependencies that haven't been in use at any point during my tenure at LIL.

I was able to remove some packages from the Docker image:
```
        libjpeg62-turbo-dev \
        imagemagick \
        libmagickwand-dev \
```
We probably want to make a similar move on the stage and prod minions after deployment.

See ENG-173.


bensteinberg (approved):
LGTM! I think in addition to removing libjpeg62-turbo-dev, imagemagick, libmagickwand-dev, I also need to make sure that libpango-1.0-0 and libcairo2 are explicitly installed (they're already there). I'll also remove `GEOCODING_KEY` from production settings.

rebeccacremona:
Thanks! I think you don't have to install `libpango-1.0-0` and `libcairo2`: I just had to do that for the functional tests, which use playwright and chromium. I don't think you have to install any of the packages in that stanza, right?

bensteinberg:
Ah whoops right, thanks.

PR #3451, opened 2023-12-11
Tweak strategy for finding links to upload to IA

rebeccacremona:
While investigating various strategies for monitoring slow or expensive queries, it came to @mdellabitta's attention that the [job to enqueue internet archive upload tasks](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/settings/utils/post_processing.py#L71), which runs every five minutes, was using up more DB juice than we'd like.

This PR takes another swing at making that query less expensive.

First, it uses different logic, depending on the case:
- if asking about a date after 2022-10-03, the last day we ever created "individual" style Internet Archive Items, then it uses a simpler query
- if asking about 2022-10-03 or earlier, it does the joins necessary to ignore "individual" style Internet Archive Items when calculating.

Second, it reorders the queries to use a new index, an index that starts with the "date" portion of `Link.creation_timestamp`. 

Tested locally with a production-like database, performance did improve: I can see the new index being used, and can see times decrease..... but, on my laptop, these queries are never particularly slow. 

New:
image

Orig:
image

Let's see if this helps enough... If it doesn't, we might price out getting some more RAM.

With thanks to @mdellabitta and @bensteinberg for suggestions and moral support when debugging the index creation.


bensteinberg (approved):
👍

PR #3456, opened 2023-12-12
Tidy

rebeccacremona:
This PR just takes care of a few trivial, unrelated things: I gave the remaining codebase a quick read through, to look for additional easy-to-tackle tech debt, and while reading, I saw a few assorted things that could be cleaned up.


matteocargnelutti (approved):
Tests pending but: 🤘  🎸

PR #3458, opened 2023-12-13
Rek sponsored links test

kilbergr:
This makes pytests from unit tests for the remaining models tests that probe the action of moving bonus links around between folders.

I added this line: https://github.com/harvard-lil/perma/pull/3458/files#diff-7647024a9efc6c8bce2a3b02675bfd7410fb12302f16b7c4bd815391088ca407R446 to account for the fact that there was a 'Test Journal' in the JSON test fixtures that was no longer attached to the user.

I added these two lines: https://github.com/harvard-lil/perma/pull/3458/files#diff-fbe1ed0bfd6c4d106680862939ce3882594ac33091fba94e3263812b27bcd52aR1592 and https://github.com/harvard-lil/perma/pull/3458/files#diff-fbe1ed0bfd6c4d106680862939ce3882594ac33091fba94e3263812b27bcd52aR1568 to the test code rather than having it in the fixture to be able to pass the fixture into the test.

### DIFF
--- a/perma_web/conftest.py
+++ b/perma_web/conftest.py
@@ -167,7 +167,7 @@ from dateutil.relativedelta import relativedelta
 from django.db.models import signals
 from django.utils import timezone
 
-from perma.models import Registrar, Organization, LinkUser, Link, CaptureJob
+from perma.models import Registrar, Organization, LinkUser, Link, CaptureJob, Sponsorship
 from perma.utils import pp_date_from_post
 
 
@@ -269,6 +269,36 @@ class LinkUserFactory(DjangoModelFactory):
     is_confirmed = True
 
     password = factory.PostGenerationMethodCall('set_password', 'pass')
+
+
+@register_factory
+class RegistrarUserFactory(LinkUserFactory):
+    registrar = factory.SubFactory(RegistrarFactory)
+
+
+# SponsorshipFactory has to come after RegistrarUserFactory and LinkUserFactory,
+# and before SponsoredUserFactory
+@register_factory
+class SponsorshipFactory(DjangoModelFactory):
+    class Meta:
+        model = Sponsorship
+
+    user = factory.SubFactory(LinkUserFactory)
+    registrar = factory.SubFactory(RegistrarFactory)
+    created_by = factory.SubFactory(
+        RegistrarUserFactory,
+        registrar=factory.SelfAttribute('..registrar')
+    )
+
+
+@register_factory
+class SponsoredUserFactory(LinkUserFactory):
+
+    sponsorships = factory.RelatedFactoryList(
+        SponsorshipFactory,
+        size=1,
+        factory_related_name='user'
+    )
 
 
 @register_factory
--- a/perma_web/conftest.py
+++ b/perma_web/conftest.py
@@ -167,7 +167,7 @@ from dateutil.relativedelta import relativedelta
 from django.db.models import signals
 from django.utils import timezone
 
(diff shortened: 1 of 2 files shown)


perma_web/conftest.py
```diff
@@ -271,6 +271,36 @@ class Meta:
     password = factory.PostGenerationMethodCall('set_password', 'pass')
 
 
+@register_factory
+class RegistrarUserFactory(LinkUserFactory):
+    registrar = factory.SubFactory(RegistrarFactory)
+
+
+# SponsorshipFactory has to come after RegistrarUserFactory and LinkUserFactory,
+# and before SponsoredUserFactory
```
rebeccacremona, line 280: Another way to deal with import problems like this, instead of worrying about the order of definition, is to use a string with the full path to the related fixture, instead of using the class itself, like [here](https://github.com/harvard-lil/perma/blob/develop/perma_web/conftest.py#L317).

E.g. something like `user = factory.SubFactory('conftest.LinkUserFactory)`, if this is above the definition of `LinkUserFactory` in the file.

This is perfect here; no need to change or move!

But since you commented, I thought I'd mention :-)

perma_web/perma/tests/test_models.py
```diff
@@ -1459,178 +1458,154 @@ def test_registrar_most_active_org_this_year(registrar, organization_factory, li
...
+    sponsorship_factory(registrar=registrar)
+    sponsored_folders = Folder.objects.filter(sponsored_by=registrar).prefetch_related('parent')
+    assert registrar.name != new_name and sponsored_folders
+    for folder in sponsored_folders:
+        if folder.parent.is_sponsored_root_folder:
+            assert folder.name == registrar.name
+        else:
+            assert folder.name != registrar.name
+    registrar.name = new_name
+    # You need to save if you use refresh_from_db
```
rebeccacremona, line 1491: Indeed! A person uses `refresh_from_db` only after calling save: the aim is to get a clean copy of the object so that you trust the results. For instance, if you literally typed:
```
registrar.foo = 'bar'
registrar.save()
```
here, and then did `assert registrar.foo == 'bar'`... it would pass, since you are just checking the very object that you created. But, if you were to do:
```
registrar.foo = 'bar'
registrar.save()
registrar.refresh_from_db()
```
and then `assert registrar.foo == 'bar'`, it would fail... because the `Registrar` model HAS no field called `foo`, so the call to save is a no-op, and so a fresh copy of that registrar does not have foo set to bar.

Does that make sense?

perma_web/conftest.py
```diff
@@ -401,6 +437,19 @@ def user_with_links_this_month_before_the_15th(link_user, link_factory):
     return link_user
 
 
+@pytest.fixture
+def complex_user_with_bonus_link(link_user_factory, folder_factory, organization, registrar,
+                                 sponsorship_factory, link_factory, in_subfolder=False):
```
rebeccacremona: Nice!

I have a couple questions about this fixture; maybe let's chat?

kilbergr: sponsorship_factory(registrar=registrar, user=user, created_by=registrar.user)
The created by user has to be the registrar user--this won't work, use debugger to figure out how to access registrar user from registrar.

kilbergr: folder_factory for sponsored folder has to be gotten through sponsorship factory.

rebeccacremona (changes requested):
Cool! Thanks for this! 

As discussed in person, just a few tweaks!

rebeccacremona (approved):
LGTM!

PR #3461, opened 2023-12-19
redoes 3 link count caching tests as pytests

kilbergr:
This is a PR for the link caching tests--translated from unittest to pytest. https://linear.app/harvard-lil/issue/ENG-510/test-link-count-caching-use-pytest-style-tests-instead-of-old-class


perma_web/perma/tests/test_link_count_caching.py
```diff
@@ -1,67 +1,52 @@
...
+def test_link_count_for_orgs(organization, link_user, org_link_factory):
+    """ We do some link count tallying on save. Let's make sure
+    we're adjusting the counts on the orgs """
 
-    ### TESTS ###
+    org_to_which_user_belongs = organization
+    link_user.organizations.set([org_to_which_user_belongs])
+    link_count = org_to_which_user_belongs.link_count
+    link = org_link_factory(created_by=link_user, submitted_url="http://example.com", organization=org_to_which_user_belongs)
+    link.save()
```
rebeccacremona, line 20: I think you can simplify this by making an OrgUserFactory.

Also: I am pretty sure you can make a link associated with an org like this: `link_factory(organization=my_org, ...)`.

In which case, this would look something like:
```
def test_link_count_for_orgs(org_user, link_factory):
    """ We do some link count tallying on save. Let's make sure
    we're adjusting the counts on the orgs """

    ### TESTS ###
    org = org_user.organizations.first()
    link_count = org.link_count
    link = link_factory(created_by=org_user, submitted_url="http://example.com", organization=org)
    link.save()
    # etc.
```

A little clearer, do you think?

perma_web/perma/tests/test_link_count_caching.py
```diff
@@ -1,67 +1,52 @@
...
-        self.assertEqual(link_count + 1, self.regular_user.link_count)
+    org_to_which_user_belongs.refresh_from_db()
+    assert link_count == org_to_which_user_belongs.link_count
 
-        link.safe_delete()
-        link.save()
 
-        self.regular_user.refresh_from_db()
-        self.assertEqual(link_count, self.regular_user.link_count)
+def test_link_count_for_registrars(registrar, link_user_factory, organization_factory, org_link_factory):
```
rebeccacremona: Similarly, I think you can simplify this by doing more setup in the fixtures. I think this wants to use just a `registrar_user` and a `link_factory`. 

You would need to arrange for the `RegistrarFactory` to generate its related orgs, using the same strategy as:
```
@register_factory
class SponsoredUserFactory(LinkUserFactory):

    sponsorships = factory.RelatedFactoryList(
        SponsorshipFactory,
        size=1,
        factory_related_name='user'
    )
```
Let me know if that doesn't make sense!

rebeccacremona (changes requested):
Thanks for working on this!

I have a couple suggestions.

Let me know if they are no fun, and we can merge as-is.

perma_web/perma/tests/test_link_count_caching.py
kilbergr: I used this: https://factoryboy.readthedocs.io/en/stable/recipes.html#simple-many-to-many-relationship to make the OrgUserFactory but you still have to pass in the organization so it's not quite as simple as this. Let me know what you think and I can keep tinkering.

perma_web/perma/tests/test_link_count_caching.py
kilbergr: As I understand it, the reason that works is because there's a `user` field in the `sponsorship` model. Since there's no `organization` field on the `registrar` model, I don't think you can do quite the same thing. I'm reading from this documentation: https://factoryboy.readthedocs.io/en/stable/recipes.html#reverse-dependencies-reverse-foreignkey
Am I misunderstanding?
I will still try to simplify.

perma_web/perma/tests/test_link_count_caching.py
rebeccacremona: The general pattern we're aiming for is to hide away the complexity inside the fixtures, so that you don't have to see it in the test.

Using that post-generation hook definitely sounds like one good approach. Another might be something in the style of the `user_with_links` fixture:

```
@pytest.fixture
def org_user(link_user, organization):
    link_user.organizations.add(organization)
    return link_user

def test_cremona(org_user):
    import pdb; pdb.set_trace()
    assert True

> /perma/perma_web/api/tests/test_link_resource.py(847)test_cremona()
-> assert True
(Pdb) org_user
<LinkUser: user0@example.com>
(Pdb) org_user.organizations.first()
<Organization: Graham, Church and White>
(Pdb) org_user.organizations.first().registrar
<Registrar: Garcia-Matthews>
```

Or, if you wanted to provide the option to pass orgs, and not always have one be randomly generated on the spot, you could do:

```
@pytest.fixture
def org_user_factory(link_user, organization):
    def f(orgs=None):
        if orgs:
            link_user.organizations.set(orgs)
        else:
            link_user.organizations.add(organization)
        return link_user
    return f

@pytest.fixture
def org_user(org_user_factory):
    return org_user_factory()
```

So then, in the test itself, you don't have to worry about constructing an org user, you can just summon one from the ether.

I remember really struggling with this pattern when Jack showed it to me, and then really enjoying it. Your mileage may vary 😄. If this doesn't strike you as a good idea, please let me know :-)

perma_web/perma/tests/test_link_count_caching.py
rebeccacremona: It works because there is a `registrar` `ForeignKey` field on the `Organization` model :-)

Example implementation:

```
@register_factory
class RegistrarWithOrgsFactory(RegistrarFactory):

    organizations = factory.RelatedFactoryList(
        'conftest.OrganizationFactory',
        size=3,
        factory_related_name='registrar'
    )

def test_cremona(registrar_with_orgs):
    import pdb; pdb.set_trace()
    assert True

> /perma/perma_web/api/tests/test_link_resource.py(847)test_cremona()
-> assert True
(Pdb) registrar_with_orgs.organizations.all()
<OrganizationQuerySet [<Organization: Dean-Daniel>, <Organization: Green Inc>, <Organization: Jones Inc>]>
(Pdb) for org in registrar_with_orgs.organizations.all(): print(org.registrar)
Johnson, Robinson and Adams
Johnson, Robinson and Adams
Johnson, Robinson and Adams
(Pdb) registrar_with_orgs.name
'Johnson, Robinson and Adams'
```

Make sense? Happy to chat more if not!

There's probably no need to make a whole separate `Factory` subclass; probably harmless to add it to the regular one, though I haven't tried it and haven't thought it through. 

A fun trick might be, working out how to optionally pass in how many orgs to create, with a default of one.

perma_web/perma/tests/test_link_count_caching.py
kilbergr: Okay, I see what you're getting at now. Thanks for the explainer.

perma_web/perma/tests/test_link_count_caching.py
kilbergr: I tried adding as a param like this:

```
@register_factory
class RegistrarFactory(DjangoModelFactory):
    class Meta:
        model = Registrar
        exclude = ('num_orgs',)

    name = factory.Faker('company')
    email = factory.Faker('company_email')
    website = factory.Faker('url')
    num_orgs = 1

    # Default to "approved" in the fixtures for convenience
    status = 'approved'

    organizations = factory.RelatedFactoryList(
        'conftest.OrganizationFactory',
        size=num_orgs,
        factory_related_name='registrar'
    )
```

Unsuccessfully.

rebeccacremona: Ack! I should not have tempted fate and used the word "fun".

Here's what I expected would work, which is almost exactly what you came up with, except instead of looking straight at the value of `num_orgs`, it makes use of `SelfAttribute`, which is how Factory Boy knows to look and see if you passed in an overriding value for `num_orgs` when you called the factory:
```
@register_factory
class RegistrarFactory(DjangoModelFactory):
    class Meta:
        model = Registrar

    class Params:
        num_orgs = 1

    name = factory.Faker('company')
    email = factory.Faker('company_email')
    website = factory.Faker('url')

    # Default to "approved" in the fixtures for convenience
    status = 'approved'

    organizations = factory.RelatedFactoryList(
        'conftest.OrganizationFactory',
        size=factory.SelfAttribute('num_orgs'),
        factory_related_name='registrar'
    )
```
But!!!!!! Turns out there is an [open issue](https://github.com/FactoryBoy/factory_boy/issues/767) about exactly this: the thing I expected to work.... does not work. Which is no fun. I put a 👍 on the issue in question.

There are workarounds, but meh, let's just hope we never need them.

Sorry for the misadventure!

rebeccacremona (approved):
🔥 🤘 🚀

perma_web/perma/tests/test_link_count_caching.py
kilbergr: I saw this same issue in my travels! And another related one...anyway it was no big deal. Thanks for the thorough review!

PR #3462, opened 2024-01-02
Send URL validation requests to Scoop API

rebeccacremona:
See ENG-502.

This PR updates Perma's dev environment to use the latest Scoop API and, arranges so that, when `VALIDATE_URL_LOCALLY` is `False`, URLs submitted to Perma for capture are validated by the Scoop API instead of locally by the Perma app.

It toggles the flag in CI, but keeps it `True` for deployments by default: to be discussed. Maybe I should just change it in `settings_common`: so far as I am concerned we are ready to give it a try.


rebeccacremona:
(The tests probably failed because this is not the right [IP range](https://github.com/harvard-lil/perma/pull/3462/files#diff-604a772918d15bfb09c21d50c4a5096dc90bcf6e935388fda578ad41ff71e24bR127) to allowlist in CI)

bensteinberg (approved):

rebeccacremona:
> ## Suspect Issues
> 
> This pull request was deployed and Sentry observed the following issues:
> 
>     * ‼️ **ScoopAPIException: 502: {}** `/api/v1/user/archives/` [View Issue](https://library-innovation-lab-harvard.sentry.io/issues/4808428104/?referrer=github-pr-bot)
> 
> 
> Did you find this useful? React with a 👍 or 👎

Just a single instance, it's fine, Sentry, thanks though.

PR #3464, opened 2024-01-05
Longer validation timeout

rebeccacremona:
We are sometimes seeing `ScoopAPINetworkException` during validation that don't seem like actual network problems... they seem like Perma is giving up before the Scoop API is ready. 

Let's give it some more time.

We can rethink this, and other timeouts, as we see how things pan out over time.


matteocargnelutti (approved):
_Yes_

PR #3467, opened 2024-01-09
Remove logging about URL munging

rebeccacremona:
During the early days of integrating Perma and Scoop, we wanted to make sure we understood what was happening to the formatting of URLs, from the moment a user typed or pasted something into Perma's input field, all the way through the pipeline, to how it showed up in the WARC, the provenance info, and the Scoop API results.

This logging was part of that.

But, this is well-understood now: we don't need the logging anymore.

This PR removes it.

### DIFF
--- a/perma_web/perma/celery_tasks.py
+++ b/perma_web/perma/celery_tasks.py
@@ -124,15 +124,6 @@ def save_scoop_capture(link, capture_job, data):
                 'private_reason'
             ])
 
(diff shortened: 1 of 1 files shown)


bensteinberg (approved):

matteocargnelutti (approved):

rebeccacremona:
We very approve of removing logging :-)

PR #3469, opened 2024-01-10
Rek test link mgmt

kilbergr:
Two things:

1. I am doing this: https://github.com/harvard-lil/perma/compare/develop...kilbergr:perma:rek-test-link-mgmt?expand=1#diff-7647024a9efc6c8bce2a3b02675bfd7410fb12302f16b7c4bd815391088ca407R508 because I can't pass `secure=True` into `super().generic(*args, **kwargs)` because it is already set in kwargs (so I get this error: TypeError: django.test.client.RequestFactory.generic() got multiple values for keyword argument 'secure'). Lmk if this is objectionable.
2. I added a response.status_code check to the last test, which didn't have one: https://github.com/harvard-lil/perma/compare/develop...kilbergr:perma:rek-test-link-mgmt?expand=1#diff-3d65703d1dab85c7f9193746596fd9d55e3dd667ebe8a10aede02e538ae68ec1R67. However, I can't tell from the test's name whether I should be expecting a different response (i.e. does permitted mean you're allowed to delete this link, and therefore this should be a 200 or is permitted a link type?). Do you know? It passes on a 404, which from this: https://github.com/harvard-lil/perma/blob/f5e43c21a39e61bf7b9dd1b63a925acbfc387678/perma_web/perma/views/link_management.py#L45 suggests this is a not deletable link (`request.user.can_delete(link) returns False`.

### DIFF
--- a/perma_web/perma/tests/test_views_link_management.py
+++ b/perma_web/perma/tests/test_views_link_management.py
@@ -10,22 +10,22 @@ from perma.views.link_management import Link
 
 ### create_link function ###
 
-def test_display_after_delete_real_link(perma_client, link_user):
+def test_display_after_delete_real_link(perma_client, link_factory, link_user):
+    new_link = link_factory(created_by=link_user)
     response = perma_client.get(reverse('create_link'),
                             as_user=link_user,
-                            data={'deleted':'ABCD-0003'})
+                            data={'deleted': new_link.guid})
     messages = list(response.context['messages'])
     assert len(messages) == 1
-    assert str(messages[0]) == "Deleted - Wikipedia"
-    assert 200 == response.status_code
+    assert str(messages[0]) == f"Deleted - {new_link.submitted_title}"
 
 def test_display_after_delete_fake_link(perma_client, link_user):
+    assert not Link.objects.filter(guid='ZZZZ-ZZZZ').exists()
     response = perma_client.get(reverse('create_link'),
                                 as_user=link_user,
                                 data={'deleted': 'ZZZZ-ZZZZ'})
     messages = list(response.context['messages'])
     assert len(messages) == 0
-    assert 200 == response.status_code
 
 def test_reminder(perma_client, link_user):
     response_content = perma_client.get(reverse('create_link'),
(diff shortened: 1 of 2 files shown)


rebeccacremona:
sorry for the delay, @kilbergr! i promise to take a look early on Tues!

rebeccacremona (changes requested):
Another apology for the very late review!

>             kwargs['secure'] = True
Yes! This is a perfect way to do it. Well done.

Re: that last test.

So the original tests, when you don't see an explicit `require_status_code` all have an implicit `require_status_code=200`, because of [this default](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/tests/utils.py#L41).

So, if the revised test is getting a 404, that means the test has changed: it's not working.

Indeed, I just popped into the debugger, and confirmed: in the original we are getting a 200:
```
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /perma/perma_web/perma/tests/test_views_link_management.py(67)test_confirm_delete_permitted_link()
(Pdb) response.status_code
200
```
Which is to say, the test confirms that the user successfully deletes the link.

I will pull this code down and confirm why, but just reading through: I am pretty sure it is because, while you have migrated these tests to use the new pytest `link_user` fixture, all the referenced links (7CF8-SS4G, JJ3S-2Q5N) are things set up by the old JSON fixtures.

In the case of `test_confirm_delete_permitted_link` in particular: you are attempting to delete `JJ3S-2Q5N` as a brand new, random user, created in the moment by use of `link_user`/the LinkUserFactory. Correctly, that user does not have permission to delete link `JJ3S-2Q5N`... that user is not associated with that link in any way. For this test to check that behavior, you would need to create a link (using `link`/`LinkFactory`), and then try deleting it as `link.created_by`.

Can you switch the tests that depend on "real", existing links to use links created by the new pytest fixtures?

For the tests that depend on a "non-existent" link, I'd recommend just throwing in a `assert not Link.objects.filter(guid='ZZZZ-ZZZZ').exists()`

rebeccacremona:
(And if you don't want to do these thing.... holler, and I'll send you a commit :-)

kilbergr:
@rebeccacremona that makes a lot of sense. I've made the suggested changes I believe. Let me know if this wasn't what you were thinking. Thank you!

perma_web/perma/tests/test_views_link_management.py
```diff
@@ -1,71 +1,67 @@
 # -*- coding: utf-8 -*-
+from django.urls import reverse
+
 from mock import patch
 
-from .utils import PermaTestCase
+# from .utils import PermaTestCase
```
rebeccacremona:
```suggestion
```

rebeccacremona (approved):
Huzzah! Looks great!

PR #3471, opened 2024-01-24
Rek pytest views auth

kilbergr:
This PR translates test_views_auth from unittest to pytest style. I copied and modified some helpers from perma/tests/utils, namely `submit_form`: https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/tests/utils.py#L84, `randomize_capitalization`, and `log_in_user`: https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/tests/utils.py#L24.

### DIFF
--- a/perma_web/conftest.py
+++ b/perma_web/conftest.py
@@ -686,11 +686,6 @@ def randomize_capitalization(s):
     return ''.join(choice((str.upper, str.lower))(c) for c in s)
 
 
-def log_in_user(client, user, password=TEST_USER_PASSWORD):
-    client.logout()
-    return client.post(reverse('user_management_limited_login'), {'username': user, 'password': password}, secure=True)
-
-
 def submit_form(client,
                 view_name,
                 data={},
--- a/perma_web/conftest.py
+++ b/perma_web/conftest.py
@@ -294,19 +294,19 @@ class LinkUserFactory(DjangoModelFactory):
 
 
 @register_factory
-class RegistrarUserFactory(LinkUserFactory):
-    registrar = factory.SubFactory(RegistrarFactory)
-
-
-@register_factory
-class DeactivatedRegistrarUserFactory(RegistrarUserFactory):
+class DeactivatedUserFactory(LinkUserFactory):
     is_active = False
 
 
 @register_factory
-class UnactivatedRegistrarUserFactory(RegistrarUserFactory):
+class UnactivatedUserFactory(LinkUserFactory):
     is_active = False
     is_confirmed = False
+
+
+@register_factory
+class RegistrarUserFactory(LinkUserFactory):
+    registrar = factory.SubFactory(RegistrarFactory)
 
 
 # SponsorshipFactory has to come after RegistrarUserFactory and LinkUserFactory,
--- a/perma_web/conftest.py
+++ b/perma_web/conftest.py
@@ -2,6 +2,7 @@ import pytest
 import boto3
 from dataclasses import dataclass
 import os
+from random import choice
 import subprocess
 from django.conf import settings
 from django.core.management import call_command
@@ -185,6 +186,8 @@ from perma.utils import pp_date_from_post
 
 
 GENESIS = datetime.fromtimestamp(0).replace(tzinfo=tz.utc)
+# this gives us a variable that we can use unhashed in tests
+TEST_USER_PASSWORD = 'pass'
 
 ### internal helpers ###
 
@@ -287,12 +290,23 @@ class LinkUserFactory(DjangoModelFactory):
     is_active = True
     is_confirmed = True
 
-    password = Password('pass')
+    password = Password(TEST_USER_PASSWORD)
 
 
 @register_factory
 class RegistrarUserFactory(LinkUserFactory):
     registrar = factory.SubFactory(RegistrarFactory)
+
+
+@register_factory
+class DeactivatedRegistrarUserFactory(RegistrarUserFactory):
+    is_active = False
+
+
+@register_factory
+class UnactivatedRegistrarUserFactory(RegistrarUserFactory):
+    is_active = False
+    is_confirmed = False
 
 
 # SponsorshipFactory has to come after RegistrarUserFactory and LinkUserFactory,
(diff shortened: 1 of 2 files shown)


perma_web/conftest.py
```diff
@@ -662,3 +676,58 @@ def one_two_three_dict():
...
+# Helpers
+#
+
+def randomize_capitalization(s):
+    return ''.join(choice((str.upper, str.lower))(c) for c in s)
+
+
+def log_in_user(client, user, password=TEST_USER_PASSWORD):
+    client.logout()
+    return client.post(reverse('user_management_limited_login'), {'username': user, 'password': password}, secure=True)
```
rebeccacremona: Okay, I feel like I should be able to figure this out, but: do you know why we need both `log_in_user` and `attempt_login`? _Do_ we need both? If you aren't sure either... then I think we should try and see if we can get away with just `attempt_login`.

perma_web/conftest.py
```diff
@@ -662,3 +676,58 @@ def one_two_three_dict():
...
+                success_url=None,
+                success_query=None,
+                form_keys=['form'],  # name of form objects in RequestContext returned with response
+                error_keys=[],  # keys that must appear in form error list
+                *args, **kwargs):
+    """
+        Post to a view.
+        success_url = url form should forward to after success
+        success_query = query that should return one object if form worked
+    """
```
rebeccacremona, line 701: I'm glad you were able to port this utility! I was wondering if it was going to be a pain to redo. It's used extensively in other parts of the test suite, I think, so will come in handy!

perma_web/perma/tests/test_views_auth.py
```diff
@@ -1,159 +1,149 @@
...
+    """
+    Test the login form
+    We should get redirected to the create page
+    """
+    # Login through our form and make sure we get redirected to our create page,
+    # no matter how the email address is capitalized
+    assert LinkUser.objects.filter(email__iexact=link_user.email).count() == 1
+
+    attempt_login(perma_client, link_user.email, TEST_USER_PASSWORD)
+
```
rebeccacremona:
```suggestion
```

perma_web/perma/tests/test_views_auth.py
```diff
@@ -1,159 +1,149 @@
...
+    perma_client.logout()
+    attempt_login(perma_client, link_user.email.upper(), TEST_USER_PASSWORD)
+    perma_client.logout()
+    attempt_login(perma_client, link_user.email.title(), TEST_USER_PASSWORD)
+    perma_client.logout()
+    attempt_login(perma_client, randomize_capitalization(link_user.email), TEST_USER_PASSWORD)
+
+def test_deactived_user_login(perma_client, deactivated_registrar_user):
```
rebeccacremona: Just curious: any particular reason you went with a registrar user instead of a standard user for the deactivated/and not-yet-activated login tests? (Which is fine)

perma_web/perma/tests/test_views_auth.py
```diff
@@ -1,159 +1,149 @@
...
+    Let's make sure we can login and change our password
+    """
+
+    log_in_user(perma_client, user=link_user.email, password=TEST_USER_PASSWORD)
+    assert '_auth_user_id' in perma_client.session
+
+    perma_client.post(reverse('password_change'),
+        {'old_password':TEST_USER_PASSWORD, 'new_password1':'Changed-password1',
+        'new_password2':'Changed-password1'},
+        secure=True)
```
rebeccacremona: Ah, this instance of `secure=True` snuck in. Hooray for the new default.

perma_web/perma/tests/test_views_auth.py
```diff
@@ -1,159 +1,149 @@
...
+    response = attempt_login(perma_client, link_user.email, 'wrongpass', expect_success=False)
+    assert response.status_code == 403
+    assert 'Too Many Attempts' in str(response.content)
+    sleep(2)
+    log_in_user(perma_client, user=link_user.email, password=TEST_USER_PASSWORD)
+    assert '_auth_user_id' in perma_client.session
+
+
+# @override_settings(AXES_FAILURE_LIMIT=2)
+# def test_login_attempts_reset(self):
```
rebeccacremona: Is this one commented out because it's not quite working yet? If so, lemme know if I can help :-)

rebeccacremona (commented):
Thanks for plugging away at this!

Just a few questions :-)

perma_web/conftest.py
kilbergr: Great! Will do.

perma_web/perma/tests/test_views_auth.py
kilbergr: Wow I have no idea. I for some reason thought these were registrar users. I'll change for simplicity's sake.

perma_web/perma/tests/test_views_auth.py
kilbergr: Nope, no reason! I'll put it back!

perma_web/perma/tests/test_views_auth.py
```diff
@@ -8,152 +7,145 @@
...
+    assert '_auth_user_id' not in perma_client.session
+
+def test_logout(perma_client, link_user):
+    """
+    Test our logout link
+    """
+
+    # Login with our client and logout with our view
+    attempt_login(perma_client, link_user.email, TEST_USER_PASSWORD)
+    assert '_auth_user_id' in perma_client.session
```
rebeccacremona: `attempt_login` already checks this assertion, right?

I think you can remove that assertion every time you call `attempt_login`,

perma_web/perma/tests/test_views_auth.py
```diff
@@ -8,152 +7,145 @@
...
+
+    perma_client.post(reverse('password_change'),
+        {'old_password':TEST_USER_PASSWORD, 'new_password1':'Changed-password1',
+        'new_password2':'Changed-password1'})
+
+    perma_client.logout()
+
+    # Try to login with our old password
+    attempt_login(perma_client, link_user.email, TEST_USER_PASSWORD, expect_success=False)
+    assert '_auth_user_id' not in perma_client.session
```
rebeccacremona: Same here: because of `expect_success=False`, I think that assertion has already been made.

rebeccacremona (changes requested):
This is looking great! Made one quick comment about duplicative assertions: I think that's it and then 🚀 !!

rebeccacremona (approved):

PR #3480, opened 2024-03-08
Python and dependency updates

bensteinberg:
This is a work in progress.


bensteinberg:
FWIW, locally I am still seeing one or two test failures, playwright timeouts in functional_tests/test_auth.py::test_log_out[chromium] and functional_tests/test_links.py::test_create_link[chromium]

rebeccacremona (approved):
Looks good to me! Here's what I tested:

- successfully built locally
- created a perma link with standard settings
- set `CELERY_TASK_ALWAYS_EAGER = False` and made a link batch, to exercise the more production-like celery setup
- ran with --debug-toolbar
- checked a few pages of the Django admin
- interacted with Perma Payments
- made changes to the JS and the CSS and observed that compilation still works
- python tests work for me locally, including functional tests

I did not test anything internet archive-related. 

We may want to address the deprecation warnings with the functional tests at some point....

I can't think of anything else to check.

PR #3482, opened 2024-03-18
Install Vue

rebeccacremona:
See ENG-737 and ENG-731.

As a first stage in modernizing Perma's front end, we are beginning a migration to VueJS.

This PR installs Vue, configures webpack to build a new Vue-inclusive bundle for Perma's dashboard page, and arranges for that bundle to be loaded (admins only). 

Since that's another 700K... I decided to go ahead and tack ENG-731 onto this PR as well, so that admin users won't get this addition by default: they'll have to opt in via the query string.

I created a new waffle flag, `vue-dashboard`, via a tiny data migration, and then added that check to the view. Admins can toggle it by adding `?dwf_vue-dashboard=0` (false) or `?dwf_vue-dashboard=1` (true) to the URL as desired. 

Examples with folders: 
- https://perma.test:8000/manage/create?folder=52&dwf_vue-dashboard=0
- https://perma.test:8000/manage/create?folder=52&dwf_vue-dashboard=1


rebeccacremona:
Ah, I meant to say: I was copying the `frontend/pages` directory structure of [h2o](https://github.com/harvard-lil/h2o/tree/develop/web/frontend), imagining there would be `frontend/components` etc. going forward as necessary.

jcushman:
lgtm!

bensteinberg (approved):

tinykite (approved):
Great job!

PR #3484, opened 2024-03-20
Set up webpack loader

rebeccacremona:
This PR installs the webpack loaders for .vue files, and adds that to webpack's config.


tinykite (approved):
Looks reasonable to me! Thank you for doing this

PR #3485, opened 2024-03-21
Add app and test vue components

tinykite:
## What this does 
This adds two components, `App.vue` and `HelloWorld.vue`. It's basically a very light refactoring of what we already had into Vue-specific components: a primary `App.vue` that we'll add all individual components to, and an additional test component with some superfluous text defined in a `ref` to test out that we can scope JavaScript to an individual component. 

This component is intentionally extremely basic for the purpose of iterative development. In the future, another PR will update this component to do something more meaningful than simply render "Hello world"

## Screenshots 
If Vue Dev Tools are installed, and Vue has successfully loaded, there should be a "Vue" tab in dev tools for browsers such as Vue and Arc that displays a tree view of the currently rendered components: 
Screenshot 2024-03-21 at 1 13 17 PM

The rendered "Hello world!" text in Perma
Screenshot 2024-03-21 at 1 13 35 PM
 
 ## How to test
- To toggle the hello world text, you will need to visit: 
[https://perma.test:8000/manage/create?dwft_vue-dashboard=1](https://perma.test:8000/manage/create?dwft_vue-dashboard=1)
- In the process of writing this component, I went through an entire saga of not getting a cookie to be able to be added in response to toggling the feature flag with this url. As best as I can tell, this is happening because of the current login session, and can be fixed by logging out and logging back in again. 
- If the cookie shows up in dev tools but you don't see "Hello world!" on the page, you may need to do a hard refresh or open an incognito window, and verify that: 
  - Vue has loaded, using the Vue dev tools 
  - If the vue bundle has been detected on the page, you should be able to spot the new components in a "vue" tab of your dev tools panel. 
  -  The div that is used to mount the Vue app, ` <div id="vue-app"></div>` is present on the page

 ## Open Questions
- In the process of developing this, webpack created two separate exports in the `static` directory: `dashboard.js` and `dashboard.js.map`. Are these necessary to ship to production, or should we update our `.gitignore` to ignore these and other static exports?
- More generally (please note: these questions don't need to be answered):
  -  Should we be using a code formatter, for example say Prettier in the Django repo? 
  -  Should we be using Typescript?


rebeccacremona (approved):
Yay! So glad you got this working!

My apologies again for my typos re: the feature flag's querystring... that wasted so much of your time.

> In the process of developing this, webpack created two separate exports in the static directory: dashboard.js and dashboard.js.map. Are these necessary to ship to production, or should we update our .gitignore to ignore these and other static exports?

I personally _do_ commit the changed files in `static/bundles` to the repo; if i remember, usually in a separate commit called something like "bump compiled assets"... but I don't always remember. 

These files _do_ need to be in the repo; they are in fact the actual files served in prod. But...... we have Github Actions configured take care of adding an up-to-date version of them if a developer does not. I cannot remember why..... mumble mumble it was more convenient for particular developers' flows, back when? People thought omitting them made PRs easier to review? I honestly don't remember.

Which is to say, feel free to include them or not... And choosing a convention / coming up with a sane strategy for built assets etc. should definitely be part of the larger "new frontend coming soon! woo hoo" plan!

tinykite:
No problem at all!

And thank you for this context — I've never worked on a project that included map files or other forms of production-ready build files, but I've also never worked on a project with an embedded web application within Django, and it's easy enough to commit whatever files are necessary.

PR #3486, opened 2024-03-27
Test with Node 20, updated deps + js-wacz

matteocargnelutti:
How I tested these changes:
- `docker compose up -d --build` to force a rebuild
- In the web container's shell:
  - `npm run build` does generate valid (?) bundles
  - `invoke run`  successfully runs `npm run build`
  - `npx js-wacz --help` prints js-wacz's help menu
  
 Tests are passing and the app behaves normally when tested locally.


rebeccacremona:
This looks great!!!

I tested by:
- Removing my local `node_modules` folder, to make sure it didn't accidentally sync a cache of the old dependencies into the container
- Removing the `perma_node_modules` volume, for the same reason
- Checking out your branch
- Adding a `test-node-dependencies` suffix to the `web` image's tag to force a rebuild without clobbering the `105-28fef2ca2ecd1c74ba3059f0b59a4d34` image
- `docker compose up`, and watching the build
- `docker compose exec web invoke run`, with `CELERY_TASK_ALWAYS_EAGER = False`
- Doing a side-by-side comparison of the CSS on the Perma dashboard with prod
- Running through all of the javascripty parts of the app:
- - making a new folder
- - renaming an existing folder
- - moving a link between folders
- - viewing an old link batch
- - making a new link batch
- - switching folders using the folder tree
- - switching folders using the dropdown
- - expanding and collapsing a link's details in the link list
- - the "upload your own" flow
- - changing a link's options using the controls in the "tray" on the playback page
- - clicking through each tab of the /manage/stats page

I can't think of anything else to check!!

perma_web/Dockerfile
```diff
@@ -36,10 +36,8 @@ RUN apt-get update \
         libxml2-dev \
         libxslt-dev
 
-# pin node version -- see https://github.com/nodesource/distributions/issues/33
-RUN curl -o nodejs.deb https://deb.nodesource.com/node_14.x/pool/main/n/nodejs/nodejs_14.19.3-deb-1nodesource1_$(dpkg --print-architecture).deb \
-    && dpkg -i ./nodejs.deb \
-    && rm nodejs.deb
+# node.js (needs to be pinned?)
```
jcushman: I recommend pinning everything in docker, ideally to the version in prod -- anything you don't pin will eventually waste an hour when it breaks on CI instead of on your local, or breaks on prod instead of CI. If keeping the pin updated is wasting more time than that I could be talked out of it.

bensteinberg: The mechanism for installing node has changed; I believe it's no longer possible to pin to a specific .deb file exactly like we used to; the new mechanism is at https://github.com/nodesource/distributions/issues/33#issuecomment-1698870039

rebeccacremona: > ideally to the version in prod

I note that we do not run nodejs in prod for Perma: it only ever runs inside the docker container

Without pinning here, I think the biggest possible consequences are, a discrepancy between different developers' environments, and an accidental upgrade to a new minor release breaking something that was previously working. Neither of those feel like big risks to me.

bensteinberg: See also https://github.com/nodesource/distributions#new-update-%EF%B8%8F and https://github.com/nodesource/distributions/wiki/How-to-select-the-Node.js-version-to-install

matteocargnelutti: Yep, I think this is fair @jcushman - I personally feel okay using this Node.js 20.x installer unpinned since it Node 20 is already in LTS mode. That of course doesn't mean it's not going to receive an update that is going to break our setup, but makes it less likely 😅 ?

jcushman: I don't feel strongly -- I just feel like I've been punished every time I let versions slip between different dev envs or between dev and CI, including I swear I remember adding node pinning to one of our projects after it wasted some time debugging something or other, so wanted to pass that on.

jcushman:
This is awesome -- I think ship it if Ben is happy? As Becky points out we use pre-built assets in production, so the testing she did is likely to have found any problems that might exist.

perma_web/Dockerfile
matteocargnelutti: @bensteinberg @jcushman I believe this would work, should I push and see if it actually does?
```bash 
# node.js 
RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash - && apt-get install -y nodejs=20.12.0-1nodesource1
```

bensteinberg: Perfect, I was about to make this exact suggestion, and point out that I _just_ installed Debian updates for node on a few machines.

rebeccacremona: @matteocargnelutti is there a build relic to remove after the fact, like the `rm` move we had before? asking just because this image is so big

matteocargnelutti: @rebeccacremona As in: it got much bigger with that change? I am not sure 👀 !

rebeccacremona: @matteocargnelutti no, I just saw that there wasn't an `rm` or `clean` or whatever and thought maybe that was by accident. Ben just checked and it is no big deal, so never mind :-)

bensteinberg (approved):

rebeccacremona (approved):

matteocargnelutti:
Thank you all for having a look and  for the helpful feedback 👋

PR #3487, opened 2024-03-27
Add reactivity experiment

tinykite:
## What this does 
This update adds a global vue store to test out cross-island interactivity between legacy code and new Vue components. Notable updates: 

- Added an additional test component, `HelloWorldB` to make sure that updates to Vue's store are shared across components. 
- Added JavaScript to `dashboard.js` to test out that we can listen to updates to the Vue store outside of Vue — and similarly make calls to update it as well.
- Added a POC for how we might handle multiple types of updates to the Vue store. Right now, we're only handling dispatches to increment the store's count.

## Screenshot 
This small experiment provides three different buttons to increment the store's count (one Vanilla JS, two rendered by Vue)
Screenshot 2024-03-27 at 11 53 43 AM

Clicking any of the buttons should update all three counts:
Screenshot 2024-03-27 at 11 53 56 AM

### DIFF
--- a/perma_web/frontend/pages/dashboard.js
+++ b/perma_web/frontend/pages/dashboard.js
@@ -20,6 +20,7 @@ watch(
 const handleDispatch = (name) => {
     switch (name) {
         case "increment":
+        default:
             globalStore.increment()
         break;
     }
--- a/perma_web/frontend/pages/dashboard.js
+++ b/perma_web/frontend/pages/dashboard.js
@@ -1,3 +1,39 @@
 import { createApp } from 'vue'
 import App from '../components/App.vue'
+import { watch } from 'vue';
+import { globalStore } from '../stores/globalStore';
+
 createApp(App).mount('#vue-app')
+
+const dashboardTestDiv = document.getElementsByClassName("vanilla-div")[0]
+
+// Track updates our Vue app makes to the store
+watch(
+    () => globalStore.count,
+    (count) => {
+        dashboardTestDiv.innerHTML = `Vanilla JavaScript count: ${count}`
+    },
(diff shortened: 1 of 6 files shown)


perma_web/frontend/pages/dashboard.js
```diff
@@ -1,3 +1,39 @@
 import { createApp } from 'vue'
 import App from '../components/App.vue'
+import { watch } from 'vue';
+import { globalStore } from '../stores/globalStore';
+
 createApp(App).mount('#vue-app')
+
+const dashboardTestDiv = document.getElementsByClassName("vanilla-div")[0]
```
tinykite, line 8: I don't know that we necessarily need to keep this in all one file — but including it here for now for the purpose of readability

perma_web/perma/templates/user_management/create-link.html
```diff
@@ -150,6 +150,8 @@ <h3 id="link-list-header" class="body-ah">Your Perma Links</h3>
 
 {% if request.user.is_staff %}
   {% flag "vue-dashboard" %}
+    <div class="vanilla-div">Vanilla JavaScript Count: None</div>
```
tinykite, line 153: Note: The value you see rendered on the page should be 0, not none. I wrote this to verify that I could run the a `watch` function eagerly. 

When the vue app mounts, the watch function should update this to the default starting value of the store count, and then update it again for any future updates to the Vue store.

bensteinberg (approved):
LGTM! But I defer to @rebeccacremona.

rebeccacremona (approved):
This looks great! I love it.

PR #3488, opened 2024-03-29
WARC to WACZ conversion

teovin:
This is the first draft of the WARC to WACZ conversion experiment.

We have a sample of 1000 WARC files that I uploaded to my local minio instance using Ben's [command here](https://linear.app/harvard-lil/issue/ENG-759/provide-a-subset-of-1000-sample-warc-data#comment-d8ce7bfd).

Task will get all the WARC names from the CSV file, get the corresponding WARC from storage, convert the file, and upload the resulting WACZ file to storage. It will also log the conversion duration, status, file size, and error log (if any) in a csv file.

Or optionally, the task can accept a single WARC argument and only process that one file.

 Sample invocation: 

`docker compose exec web invoke dev.benchmark-wacz-conversion --source-csv='perma/wacz_experiment/1000-a-guids.csv' --benchmark-log='perma/wacz_experiment/benchmark.csv'`

or

`docker compose exec web invoke dev.benchmark-wacz-conversion --single-warc='A276-A9A4.warc.gz' --benchmark-log='perma/wacz_experiment/benchmark.csv'`

I also replayed a few of them using Becky's replay [changes here](https://github.com/harvard-lil/perma/compare/develop...rebeccacremona:wacz-experiment?expand=1).


perma_web/perma/celery_tasks.py
```diff
@@ -1224,3 +1228,77 @@ def conditionally_queue_internet_archive_uploads_for_date_range(start_date_strin
...
+    input_file_name = input_path.split('/')[-1]
+    output_path = f"{output_folder}/{input_file_name.split('.')[0]}.wacz"
+    js_wacz_call = subprocess.run(["npx", "js-wacz", "create", "-f", input_path, "-o", output_path],
+                                  capture_output=True, check=True, text=True)
+
+    """
+    Added for A24C-RXK4.warc.gz case where js-wacz logs content-length too small
+    even though the conversion succeeds at the end
+    """
+    if 'WACZ file ready' not in js_wacz_call.stdout:
```
matteocargnelutti: It does not at the moment, but `js-wacz` should return an exit code that is not 0 when something bad happens. 
This would simplify error handling here, as `subprocess.run()` would raise an exception. That way, you wouldn't have to check for `strerr`, or inspect the contents of `stdout`. 

I created a PR for `js-wacz` to help with that: https://github.com/harvard-lil/js-wacz/pull/97

matteocargnelutti:
Really cool, @teovin 🏄 !!

perma_web/perma/celery_tasks.py
teovin: That's very cool @matteocargnelutti! I will definitely adopt it here when it is merged.

perma_web/perma/celery_tasks.py
```diff
@@ -1224,3 +1228,74 @@ def conditionally_queue_internet_archive_uploads_for_date_range(start_date_strin
...
+    input_file_name = input_path.split('/')[-1]
+    output_path = f"{output_folder}/{input_file_name.split('.')[0]}.wacz"
+    js_wacz_call = subprocess.run(["npx", "js-wacz", "create", "-f", input_path, "-o", output_path],
+                                  capture_output=True, check=True, text=True)
+    end_time = time.time()
+    duration = format_time(end_time - start_time)
+    file_size = format_size(os.path.getsize(input_path))
+
+    with open(benchmark_log, 'a') as log_file:
+        row = {
```
jcushman, line 1302: Let's log both the input and output file size.

perma_web/perma/celery_tasks.py
```diff
@@ -1224,3 +1228,74 @@ def conditionally_queue_internet_archive_uploads_for_date_range(start_date_strin
...
+            "file_name": input_file_name,
+            "conversion_status": '',
+            "file_size": file_size,
+            "duration": duration,
+            "error": ''
+        }
+        writer = csv.DictWriter(log_file, fieldnames=row.keys())
+
+        # TODO: Tweak this to check for exit code 1 when js-wacz is released
```
matteocargnelutti: @teovin `js-wacz 0.1.1` with that promised feature is out :) 

https://github.com/harvard-lil/js-wacz/releases/tag/0.1.1

perma_web/perma/celery_tasks.py
```diff
@@ -1224,3 +1228,74 @@ def conditionally_queue_internet_archive_uploads_for_date_range(start_date_strin
...
+    Converts seconds to minutes
+    """
+    if seconds_val >= 60:
+        return f"{round(seconds_val / 60, 1)} minutes"
+    else:
+        return f"{math.ceil(seconds_val)} seconds"
+
+
+@shared_task
+def convert(input_path, output_folder, benchmark_log):
```
matteocargnelutti: Although we are still in the prototyping phase, I think it is probably worth giving more descriptive names to functions related to this task, given that all celery tasks live in the same scope.

perma_web/perma/celery_tasks.py
```diff
@@ -1224,3 +1228,74 @@ def conditionally_queue_internet_archive_uploads_for_date_range(start_date_strin
...
+    end_time = time.time()
+    duration = format_time(end_time - start_time)
+    file_size = format_size(os.path.getsize(input_path))
+
+    with open(benchmark_log, 'a') as log_file:
+        row = {
+            "file_name": input_file_name,
+            "conversion_status": '',
+            "file_size": file_size,
+            "duration": duration,
```
matteocargnelutti, line 1309: A suggestion: It might be worth adding durations and sizes to the CSV in a format that allows us to filter and sort easily. 

It will be for example difficult to sort / plot a column that contains strings such as "322 KB" and "1.2 MB"

perma_web/perma/celery_tasks.py
teovin: done!

perma_web/perma/celery_tasks.py
teovin: Added the raw fields as well as keeping the formatted fields.

perma_web/perma/celery_tasks.py
teovin: done!

perma_web/perma/celery_tasks.py
teovin: @matteocargnelutti I experimented with this and am still getting code 0, will reach out to you to double-check.

teovin:
image

perma_web/perma/celery_tasks.py
teovin: @matteocargnelutti this one is pushed as well.

perma_web/perma/celery_tasks.py
```diff
@@ -1224,3 +1228,88 @@ def conditionally_queue_internet_archive_uploads_for_date_range(start_date_strin
...
+def convert_warc_to_wacz(input_path, output_folder, benchmark_log):
+    """
+    Converts WARC file to WACZ
+    Saves file in a temp folder
+    If successful, saves file in storage
+    """
+    start_time = time.time()
+    input_file_name = input_path.split('/')[-1]
+    output_path = f"{output_folder}/{input_file_name.split('.')[0]}.wacz"
+    exception_occured = False
```
bensteinberg:
```suggestion
    exception_occurred = False
```

perma_web/perma/celery_tasks.py
```diff
@@ -1224,3 +1228,88 @@ def conditionally_queue_internet_archive_uploads_for_date_range(start_date_strin
...
+    error_output = ''
+
+    try:
+        subprocess.run(["npx", "js-wacz", "create", "-f", input_path, "-o", output_path],
+                                      capture_output=True, check=True, text=True)
+    except subprocess.CalledProcessError as e:
+        exception_occured = False
```
bensteinberg:
```suggestion
        exception_occurred = True
```

perma_web/perma/celery_tasks.py
```diff
@@ -1224,3 +1228,88 @@ def conditionally_queue_internet_archive_uploads_for_date_range(start_date_strin
...
+            "raw_warc_size": os.path.getsize(input_path),  # bytes
+            "wacz_size": wacz_size,
+            "raw_wacz_size": os.path.getsize(output_path),  # bytes
+            "duration": duration,
+            "raw_duration": raw_duration,  # seconds
+            "error": ''
+        }
+        writer = csv.DictWriter(log_file, fieldnames=row.keys())
+
+        if exception_occured:
```
bensteinberg:
```suggestion
        if exception_occurred:
```

perma_web/tasks/dev.py
```diff
@@ -1110,3 +1114,78 @@ def email_retained_users(ctx, reports_dir='.'):
...
     if failed_list:
         logger.warning(f"Some users were not emailed: {str(failed_list)}. Check log for fatal SMTP errors.")
+
+
+# WACZ CONVERSION
+
+def save_warc_for_conversion(warc, warcs_dir, file_name):
+    """
+    Gets the file path
+    Get the associates file's contents from storage
```
bensteinberg:
```suggestion
    Get the associated file's contents from storage
```

perma_web/tasks/dev.py
```diff
@@ -1110,3 +1114,78 @@ def email_retained_users(ctx, reports_dir='.'):
...
+    warcs_dir = tempfile.mkdtemp()
+
+    if source_csv:
+        sample_data_guids = os.path.abspath(source_csv)
+        with open(sample_data_guids, mode='r') as file:
+            csv_file = csv.reader(file)
+            for line in csv_file:
+                save_warc_for_conversion(line[1], warcs_dir, f"{line[2]}/{line[1]}")
+    else:
+        # needed to do a bit of reverse engineering here to construct the warc storage file path
```
bensteinberg: If we were operating on Link objects, which we might well do in practice, we could use the Link's `guid_as_path()` method:
https://github.com/harvard-lil/perma/blob/7f328c8e2e57bdab12fc08fd112a6c63ab15e858/perma_web/perma/models.py#L1667-L1671

bensteinberg (changes requested):
Some minor spelling corrections, and one actual fix, I think, at line 1277.

perma_web/perma/celery_tasks.py
```diff
@@ -1224,3 +1228,88 @@ def conditionally_queue_internet_archive_uploads_for_date_range(start_date_strin
 
     else:
         logger.info("Skipped the queuing of file upload tasks: max tasks already in progress.")
+
+
+# WACZ CONVERSION
+
+def format_size(bytes_val):
```
rebeccacremona: There is in fact a nice Django utility that does exactly this! We use it in [the Django admin](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/admin.py#L399C16-L399C30) in a [few places](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/admin.py#L631).

```
from django.template.defaultfilters import filesizeformat
```

How about we reuse that here instead of rolling our own, for consistency?

perma_web/perma/celery_tasks.py
```diff
@@ -1224,3 +1229,73 @@ def conditionally_queue_internet_archive_uploads_for_date_range(start_date_strin
...
+    input_file_name = input_path.split('/')[-1]
+    output_path = f"{output_folder}/{input_file_name.split('.')[0]}.wacz"
+    exception_occurred = False
+    error_output = ''
+
+    try:
+        subprocess.run(["npx", "js-wacz", "create", "-f", input_path, "-o", output_path],
+                                      capture_output=True, check=True, text=True)
+    except subprocess.CalledProcessError as e:
+        exception_occurred = False
```
bensteinberg:
```suggestion
        exception_occurred = True
```
I think this still needs a change.

teovin: Yes, thank you @bensteinberg, missed this!

perma_web/perma/celery_tasks.py
```diff
@@ -23,6 +23,7 @@
 from django.conf import settings
 from django.utils import timezone
 from django.template.defaultfilters import pluralize
+from django.template.defaultfilters import filesizeformat
```
rebeccacremona: The tiniest of tweaks: since the line above it also importing from `django.template.defaultfilters`, can you combine these two lines?

perma_web/perma/celery_tasks.py
```diff
@@ -1224,3 +1229,104 @@ def conditionally_queue_internet_archive_uploads_for_date_range(start_date_strin
 
     else:
         logger.info("Skipped the queuing of file upload tasks: max tasks already in progress.")
+
+
+# WACZ CONVERSION
+
+def format_time(seconds_val):
```
rebeccacremona: The tiniest of suggestions: since this is such a context-specific utility, what would you think of defining it inside the scope of `convert_warc_to_wacz`? That is a pattern I personally enjoy and find tidy when I have a helper with a very narrow use (for example, [here](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/celery_tasks.py#L535), but some people like it less. If you don't prefer that approach, how about naming this function "seconds_to_minutes"?

rebeccacremona (commented):
This looks great! I proposed two trivial tweaks.

I think this is ready to merge, but I'm going to throw one idea out there for your consideration, that I think could simply things. Totally up to you though!

So, Perma's capture jobs use this [decorator](https://github.com/harvard-lil/perma/blob/7f328c8e2e57bdab12fc08fd112a6c63ab15e858/perma_web/perma/celery_tasks.py#L217), which arranges for every single job to run in its own unique temporary directory, one that is cleaned up after the job is completed.

If you were to use that decorator for this job, you could name the warc `{GUID}.warc` and the wacz `{GUID.wacz`, and save them right in the current working directory, and not have to explicitly create special tempfiles or pass around tempfile names. (And, the same thing for the custom pages.jsonl we were chatting about adding next, in a separate PR 😄). You also then wouldn't have to explicitly handle cleanup at the end.

That sounds very neat and tidy to me, but it is only a suggestion!

perma_web/perma/celery_tasks.py
```diff
@@ -1224,3 +1229,104 @@ def conditionally_queue_internet_archive_uploads_for_date_range(start_date_strin
...
+        return f"{math.ceil(seconds_val)} seconds"
+
+
+def save_warc_to_temp(input_guid):
+    """
+    Gets the file path
+    Get the associated file's contents from storage
+    Saves as temp file
+    """
+    warc_storage_path = Link.objects.get(guid=input_guid).warc_storage_file()
```
rebeccacremona: Oh! I also meant to say: since you are using the Link object here, and also at 1298, I recommend making a tweak so that we don't call out to the database twice.

bensteinberg (approved):

PR #3489, opened 2024-04-02
Add form

tinykite:
## What this does 
Duplicates a portion of the form Perma provides to create an archive of a single link, as a portion of ENG-771. Some pertinent notes: 

- This does not attempt to duplicate the behavior of the folder input. This will be handled in the future. For now, all perma archives will be created with the assumption that there is no organization-specific folder chosen for the newly-created capture. 
- There is error handling that will show an error has happened if, for example, you try to upload a capture without supplying a link. But the error will generally be generic. 
- Additionally, the perma browser tools link has not yet been added. 
- For the time being, instead of updating the appearance of the submit button to include a loading spinner and a progress bar, the capture state and upload progress percentage are logged directly to the screen. The capture progress info will not show up until the archive is actively being captured, which should match existing behavior. 
  - This will be updated in a future PR
 - The submit button is not disabled by design, per advice from Adam Silver's blog post ["The Problem with Disabled Buttons and What To Do Instead" ](https://adamsilver.io/blog/the-problem-with-disabled-buttons-and-what-to-do-instead/) and other sources on form design. Submitting the form should surface an error, although right now it will not be a helpful error just yet. 

## Screenshots
Ready to handle single url archive requests
Screenshot 2024-04-03 at 2 06 45 PM

Waiting for a response from the /archives endpoint
Screenshot of a perma dashboard waiting for a response from an API

After the archives API returns a GUID for a link, the state of the archive process will switch to "isQueing"
Screenshot of ready GUID

When a capture is in progress, the page will display the status of the overall archiving process and the progress percent of the capture job specifically
Screenshot of capture in progress

A capture failure
A screenshot of a capture failure

Redirecting to the archive after a successful capture
Screenshot of a Perma archive

## Testing notes 
- I tested this by adding the following to my `settings.py`: 
```
from .deployments.settings_dev import *
CELERY_TASK_ALWAYS_EAGER = False
```
- To make all captures fail, I added poll_data['status'] = 'failure' above [line 314 of the capture job code](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/celery_tasks.py#L314C9-L314C45) so that everything looks like failure.


perma_web/static/js/helpers/general.helpers.js
```diff
@@ -31,22 +31,11 @@ export function setCookie (cname, cvalue, exdays) {
...
-        break;
-      }
-    }
-  }
-  return cookieValue;
+export function getCookie(name) {
+  let matches = document.cookie.match(new RegExp(
+      "(?:^|; )" + name.replace(/([\.$?*|{}\(\)\[\]\\\/\+^])/g, '\\$1') + "=([^;]*)"
+  ));
+  return matches ? decodeURIComponent(matches[1]) : undefined;
```
tinykite, line 41: I find this easier to reason about, while accomplishing the same thing

perma_web/frontend/components/CreateLink.vue
```diff
@@ -0,0 +1,114 @@
...
+import { ref, watch } from 'vue'
+import { globalStore } from '../stores/globalStore'
+import { getCookie } from '../../static/js/helpers/general.helpers'
+// import { Spinner } from 'spin.js'
+// Cannot import spinner like this just yet
+
+const userLink = ref('')
+const userLinkGUID = ref('')
+const userLinkStepCount = ref(0)
+// const spinner = new Spinner({ lines: 15, length: 2, width: 2, radius: 9, corners: 0, color: '#2D76EE', trail: 50, top: '12px' });
```
tinykite: The spinner code here is temporarily commented out — 

`import { Spinner } from 'spin.js` 

Is the recommended way to import spin.js, according to its documentation, but attempting to use `new Spinner` returns a build error that `Spinner` is not a constructor. It may be easier to simply adopt a slightly more modern spinner, although implementing any kind of animation that persists for longer than 5 seconds without a play/pause button is generally understood to be an accessibility failure under the requirements of [WCAG 2.2.2](https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html).

tinykite: I haven't dug further into this, but will!

perma_web/frontend/components/CreateLink.vue
```diff
@@ -0,0 +1,114 @@
...
+                        class="text-input select-on-click form-priority-input" type="text"
+                        placeholder="Paste your URL here." data-placement="bottom"
+                        data-content="To save a link, enter its URL and click the <strong>Create Perma Link</strong> button. To see the links you've saved, click <strong>Library</strong> in the menu to the left."
+                        data-original-title="Start building your library" data-html="true" data-trigger="manual" />
+                    <div class="wrapper">
+                        <button @click.prevent="handleArchiveRequest" class="btn btn-large btn-info _active-when-valid"
+                            id="addlink" type="submit">
+                            Create Perma Link
+                            <!-- <div v-if="globalStore.captureStatus === 'isQueuing'" id="capture-status">Creating your
+                                Perma Link</div> -->
```
tinykite: Commented this out because the implementation isn't quite right just yet — I will be fixing this

perma_web/frontend/components/CreateLink.vue
tinykite: Looks like we can easily fix this by diverging from the docs and just importing spinner as: 

```
import * as spinner from 'spin.js'
```

perma_web/frontend/pages/dashboard.js
tinykite, line 1: We don't need the cross-island interaction logic for now, so I've commented it out

perma_web/frontend/stores/globalStore.js
```diff
@@ -1,6 +1,17 @@
...
+  captureStatus: 'ready',
+
+  updateCapture(state) {
+    this.captureStatus = state
+  },
+
+  captureErrorMessage: '',
+  updateCaptureErrorMessage(message) {
+    this.captureErrorMessage = message
+  },
```
tinykite, line 13: I think if this needs to be an array of errors, that should be easy to update in the future when I implement all possible Perma uploading/capture/archive related errors.

perma_web/perma/templates/user_management/create-link.html
```diff
@@ -11,6 +11,15 @@
...
 
+ <!-- vue app -->
+ {% if request.user.is_staff %}
+  {% flag "vue-dashboard" %}
+  <!-- <div class="vanilla-div">Vanilla JavaScript Count: None</div>
+  <button class="vanilla-button">Increment Counter</button> -->
+    <div id="vue-app"></div>
+  {% endflag %}
+{% endif %}
+
```
tinykite: I moved this up so that I didn't need to keep scrolling to continually test the UI.

perma_web/frontend/components/CreateLink.vue
```diff
@@ -0,0 +1,149 @@
+<script setup>
+import { ref, watch, onBeforeUnmount } from 'vue'
+import { globalStore } from '../stores/globalStore'
+import { getCookie } from '../../static/js/helpers/general.helpers'
+// import * as spinner from 'spin.js'
```
tinykite: If there's a more elegant alias for `../../` please let me know!

perma_web/frontend/components/CreateLink.vue
```diff
@@ -0,0 +1,149 @@
...
+
+const userLink = ref('')
+const userLinkGUID = ref('')
+const userLinkProgressBar = ref(0)
+
+const readyStates = ["ready", "urlError", "captureError"]
+const isReady = readyStates.includes(globalStore.captureStatus)
+
```
tinykite: While this worked initially in testing, I think I should make this return a computed value, for example: 

```
const isReady = computed(() => { readyStates.includes(globalStore.captureStatus) })
```

Otherwise I think `isReady` could easily have a stale value.

rebeccacremona: That sounds believable! Do you want to go ahead and make that tweak?

perma_web/frontend/components/CreateLink.vue
```diff
@@ -0,0 +1,149 @@
...
+                body: JSON.stringify(formData)
+            })
+
+        if (!response?.ok) {
+            throw new Error(response.statusText) // We will handle this more in-depth later
+        }
+
+        const { guid } = await response.json() // Needed to poll Perma about the capture status of a link
+        userLinkGUID.value = guid
+        globalStore.updateCapture('isQueuing')
```
rebeccacremona: Wording suggestion: how about `isQueued` rather than `isQueuing`? I defer to others; I am more used to that usage in US English.

perma_web/frontend/components/CreateLink.vue
```diff
@@ -0,0 +1,149 @@
...
+        globalStore.updateCapture('captureError')
+        globalStore.updateCaptureErrorMessage(error)
+    }
+}
+
+const handleProgressUpdate = async () => {
+    const { step_count, status, error } = await handleCaptureStatus(userLinkGUID.value);
+
+    if (status === 'in_progress') {
+        globalStore.updateCapture('isUploading')
```
rebeccacremona: Wording suggestion: how about "isCapturing" rather than "isUploading"? 

I think this will help us maintain the distinction between "a user POSTed to `/archives`, causing a capture job to be launched, and that capture job is in progress", vs. "a user is using the 'upload your own PNG or PDF' form to upload an asset to Perma in place of a failed capture"... a somewhat hidden feature of Perma which you may not have tried out yet 😄!

perma_web/frontend/pages/dashboard.js
```diff
@@ -5,36 +5,36 @@ import { globalStore } from '../stores/globalStore';
 
 createApp(App).mount('#vue-app')
 
-const dashboardTestDiv = document.getElementsByClassName("vanilla-div")[0]
+// const dashboardTestDiv = document.getElementsByClassName("vanilla-div")[0]
```
rebeccacremona: Would you like to remove this experiment from the codebase at this point, or is it helpful to keep around commented out for preference still? No pressure to remove now if you are finding it helpful!

perma_web/static/js/helpers/general.helpers.js
rebeccacremona: Wow! This is an impressive regular expression! Did you write it yourself?

I confess that reading that regular expression to figure out what it does is beyond me. If I take it on faith that it returns the first match for a cookie of a given cookie name from the cookie jar, then I agree that `matches ? decodeURIComponent(matches[1]) : undefined;` is very easy to reason about! Whereas I can read the original code and see that it iterates through each cookie, looks at that cookie's name, and see if its name starts with the name we passed to the utility function.

I have seen folks use very complex regular expressions that others have written for specific purposes (e.g., credit card number validation, email address validation). I could imagine us doing so here, if there were a clear comment linking to the source of the regular expression and documentation of what it does... and, perhaps, why it is more effective or more efficient or handles edge cases not handled by the string matches, or whatever its advantages may be.

If based on pure readability alone, I think I'd have to argue that the original is more readable.

rebeccacremona (changes requested):
🎉 This looks great!

As per discussion in Slack, I think there is one "state" that wants to be renamed, and then I left a few small questions and less-strongly-suggested tweaks.

Exciting to see this working! Brick by brick!

Apologies again for the slow review; I missed the transition from "draft" to "ready".

perma_web/static/js/helpers/general.helpers.js
tinykite: I did not write this RegExp (although I wish I did), my prowess is definitely not that advanced! I had no idea before that you can write a function that dynamically generates a new RegExp pattern. The RegExp was taken from the [Appendix: cookie functions](https://javascript.info/cookie#getcookie-name) section of the JavaScript.info website. 

I absolutely agree about the need for a comment — if we decided to keep this function, a comment would be super helpful. I could definitely include one that specifically points developers to where the function comes from; explains that `document.cookie` returns a list of cookie values in a semicolon-delimited string of key-value pairs; and that this function dynamically generates new RegExp to match on the name of one specific cookie, using the `; name=<value>` format. 

You're also right that RegExp readability is totally subjective. The reason JavaScript.info provides for using is that using dynamically-generated RegExp is the shortest way to check a cookie's value, but I am totally impartial and okay with reverting to the original code. (I definitely don't think brevity of code or the speed it takes to compute is the most crucial factor here)

Would you prefer the original code?

perma_web/frontend/components/CreateLink.vue
tinykite: For sure! Fixed in 65b58f8

perma_web/frontend/components/CreateLink.vue
tinykite: Yes, agree! Fixed in [revision 3b8a065](https://github.com/harvard-lil/perma/pull/3489/commits/3b8a06583cd388b808f68a6a5802ee3fe614b2ae)

perma_web/frontend/pages/dashboard.js
tinykite: We can totally remove it! It'd be easy enough to add back code like this if / when it's needed. Removed in [revision 5213814](https://github.com/harvard-lil/perma/pull/3489/commits/52138145adc848f1401ca62eaef3d7d3d4ef4376)

Edit: I realized we still had some other remnants related to this code elsewhere, so I removed the experimental html in [revision 5213814](https://github.com/harvard-lil/perma/pull/3489/commits/52138145adc848f1401ca62eaef3d7d3d4ef4376) and the experimental store state and function in  revision  80be457

perma_web/frontend/components/CreateLink.vue
tinykite: You're right — if I understand correctly, the job isn't "waiting" to be queued — it gets added to a queue and then is waiting to be captured. Fixed this in [revision 1622280](https://github.com/harvard-lil/perma/pull/3489/commits/162228017d68824971b4281c190e58ddf11e7f58)

tinykite:
@rebeccacremona can you take another look? Made a handful of fixes, based on feedback. 

Additionally — regarding the cookie utility function — let me know if you want to revert that, that's an easy fix too!

perma_web/static/js/helpers/general.helpers.js
rebeccacremona: Ah! 

If it's the strategy recommended by JavaScript.info, that's good enough for me 😄: while somewhat inscrutable, we know that source is trustworthy!

There is no equivalent of their reformatting of the cookie name (`name.replace(/([\.$?*|{}\(\)\[\]\\\/\+^])/g, '\\$1')`) in our utility (which is just cut and paste from an old version of [Django's docs](https://docs.djangoproject.com/en/3.0/ref/csrf/), so this is "smarter" on at least one axis.

Why don't we go ahead and keep the regex version, but link out to javascript.info and mention that we are using their recommended approach?

rebeccacremona (approved):

perma_web/static/js/helpers/general.helpers.js
tinykite: That sounds super reasonable! Made a fix for this in [revision 35e6028](https://github.com/harvard-lil/perma/pull/3489/commits/35e602851ac647692ed0475ebe843f3fd76e9016)

PR #3497, opened 2024-04-10
Install js-wacz separately

rebeccacremona:
This PR is a little peculiar.

It pulls js-wacz out of Perma's main package.json and creates a separate package.json in the `services` directory, for just that package.

It does this because, presently, nodejs is not installed on any production minions: we only need node packages to compile our frontend assets. (We commit the bundled js and css to the repo!)

To use js-wacz to convert WARCs to WACZs, we'll need to install nodejs on our worker servers, and we'll want to use the same version in prod as in dev. But... we don't necessarily want 222M of unneeded nodejs packages lying around on every worker, when all we are using is js-wacz. And, it might be confusing to see js-wacz listed amongst all our otherwise frontend-only packages since we are not using it on the frontend.

So, this splits them out, and installs js-wacz in a different `node_modules` folder as a separate step in the Dockerfile. The production installation can be informed by this approach, even though it won't be identical.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,3 +6,5 @@
 !perma_web/perma-warcprox-ca.pem
 !perma_web/lil-archive-keyring.gpg
 !playwright/requirements.txt
+!services/js-wacz/package.json
+!services/js-wacz/package-lock.json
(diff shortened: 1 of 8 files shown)


bensteinberg (approved):
Great! I assume it's ok for both setups to have `"name": "perma"`.

bensteinberg:
The js-wacz `node_modules` on a test server takes up 26M.

PR #3499, opened 2024-04-11
Add link counter

tinykite:
## What this does 
- Adds a new component, `LinkCount` plus two new additional store properties, `subscriptionStatus` and `userTypes`. 

## Notes 
- I initially found it hard to reason about the nested if/else statements in the original Django code, so I mapped out a [decision tree related to the link count](https://www.figma.com/file/WATQ1WyeJV7azc2rJory2t/Perma-Link-Count-State?type=whiteboard&node-id=0%3A1&t=K0Bl1moCoAU4IXvF-1), and then created a tentative single, unambiguous `subscriptionStatus` state that could be derived from the conditional logic of the tree. 
  - For example, a user may have metered, pay-as-you-go access to new Perma links. Or users may have unlimited links — on a paid or unpaid/free basis. (Or there may be an issue with their credit card on file)
  - If there are better terms for these categories, I am definitely all ears.   
- If I understand correctly,  there may also be more than one user account type associated with an account. For example, someone may be an individual user and staff. (Or an individual user and a sponsored user, perhaps?) I created `userTypes` to potentially easily reference these values, but right now we only needed to store whether or not someone is an individual. 
- I switched a css class from an id to a class so that it didn't need to be unique, and could be applied to both the current link counter and the modernized one.

## Screenshots 

Displaying the remaining metere links a user has available.
Screenshot 2024-04-10 at 4 35 49 PM

Displaying the remaining metered links a user has available for the rest of the year, if they have an annually renewing Perma subscription. 
Screenshot 2024-04-10 at 4 59 43 PM

Notifying a user they have paid for unlimited perma links. 
Screenshot 2024-04-10 at 4 39 48 PM

Notifying a free Perma user they have been upgraded to unlimited free perma links.
Screenshot 2024-04-10 at 4 38 39 PM

Notifying any individual Perma user that they have an issue with their billing details.
Screenshot 2024-04-11 at 3 15 40 PM

While gathering these screenshots, I noticed that we don't always consistently use periods to end statements. Should we consistently do so? (Or remove the instances where we do use periods?)


perma_web/frontend/lib/store.js
```diff
@@ -0,0 +1,25 @@
+import { globalStore } from '../stores/globalStore'
+
+export const getLinksRemainingStatus = (links_remaining, is_nonpaying) => {
+    if (links_remaining !== 'Infinity') {
```
rebeccacremona: Since `links_remaining = Number({{ links_remaining }})`, I think that means this is not the string `'Infinity'` here, but rather, the JS Infinity object. Do you agree?

image

rebeccacremona (approved):
Hooray! This looks great.

Love the question about some things ending in periods and some not. I guarantee that is by accident. What do you think looks best? Consistently using periods?

I have a thought on the two new JS variables, `is_individual` and `is_nonpaying`. Since they are attributes of `request.user`, and we are already passing `request.user.as_json` to the global `current_user` JS variable, we may want to consider adding that information there, instead of adding these two new variables. But! We can discuss later; possibly that doesn't get us anything!

perma_web/frontend/lib/store.js
tinykite: Yes, you are 100% correct and this totally causes a visual bug if not fixed 😮‍💨 where a user will see a link counter that says that they have "Unlimited" links remaining because they are marked as a metered, not unlimited user

tinykite:
> Love the question about some things ending in periods and some not. I guarantee that is by accident. What do you think looks best? Consistently using periods?

I think using periods is the most consistent. Where I've worked at the past, it was our style to always add periods to statements (regardless of if it was a complete statement) and feels the easiest to reason about. Updated this in [revision 839094c](https://github.com/harvard-lil/perma/pull/3499/commits/839094ca442da814f0520143d22083f05d4f6b43)

Additionally regarding the new variables — yes! I am always in favor of simplifying this if possible, I'll file this away in my brain as a possible future improvement.

PR #3502, opened 2024-04-12
Add wacz-playback feature flag

teovin:
Add a new feature flag called `wacz-playback`. The flag can be enabled/disabled through Django admin.

- If the flag is active, and if we have a WACZ version of the file, we will play the WACZ version, else the WARC version.
- Note: I haven't updated the download archive button logic to download the WACZ yet, I will so if we have consensus that we should also make that change.

**WACZ playback** (see the additional info in the pages section coming from the custom pages.jsonl we started adding to WACZs:

image

**WARC playback:**

image


perma_web/perma/migrations/0039_auto_20240412_1801.py
```diff
@@ -0,0 +1,29 @@
...
+def delete_wacz_playback_feature_flag(apps, schema_editor):
+    Flag = apps.get_model("waffle", "Flag")
+    flags = Flag.objects.filter(name=FLAG_NAME)
+    flags.delete()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('perma', '0038_auto_20240318_2102'),
```
rebeccacremona, line 24: I think this migration needs to mention Waffle too, like this:
https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/migrations/0038_auto_20240318_2102.py#L28

rebeccacremona (changes requested):
This looks great! I think just one tweak to the migration file and good to go!

A note to our future selves: once we have some WACZs in prod, we'll want that info to be stored in the DB so that we don't have to call out to S3 here to see if the WACZ exists.... but for now, this approach is great.

perma_web/perma/migrations/0039_auto_20240412_1801.py
teovin: Done!

rebeccacremona (approved):

PR #3503, opened 2024-04-17
Update sqlparse

bensteinberg:
https://github.com/andialbrecht/sqlparse/blob/master/CHANGELOG

This PR also removes `version` from the docker compose files; after seeing `'version' is obsolete`, I found [this](https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-and-name-top-level-elements).


cmsetzer (approved):
Looks good! Thanks for the context.

PR #3505, opened 2024-04-19
add batch_size arg, time different steps

teovin:
- Add the optional `batch-size` flag to invoke command. If not passed, it will default to 1000. This is so that we can run the job without a guids CSV or a single WARC guid. Now we can run the job with `docker compose exec web invoke dev.benchmark-wacz-conversion --benchmark-log='perma/wacz_experiment/benchmark.csv'` or `docker compose exec web invoke dev.benchmark-wacz-conversion --benchmark-log='perma/wacz_experiment/benchmark.csv' --batch-size=500`
- Time the different steps of the conversion for granularity.

### DIFF
--- a/perma_web/perma/celery_tasks.py
+++ b/perma_web/perma/celery_tasks.py
@@ -1257,12 +1257,19 @@ def convert_warc_to_wacz(input_guid, benchmark_log):
     warc_path = f"{link.guid}.warc.gz"
     wacz_path = f"{link.guid}.wacz"
     pages_path = "pages.jsonl"
+    warc_not_found_error = False
 
     # save a local copy of the warc file
-    with open(warc_path, "wb") as file:
-        copy_file_data(default_storage.open(link.warc_storage_file()), file)
+    warc_save_start_time = time.time()
+    try:
+        with open(warc_path, "wb") as file:
+            copy_file_data(default_storage.open(link.warc_storage_file()), file)
+    except FileNotFoundError:
+        warc_not_found_error = True
+    warc_save_duration = time.time() - warc_save_start_time
 
     # prepare our custom pages.jsonl file
+    jsonl_prep_start_time = time.time()
     jsonl_rows = [
         {"format": "json-pages-1.0", "id": "pages", "title": "All Pages"}
     ]
@@ -1283,9 +1290,11 @@ def convert_warc_to_wacz(input_guid, benchmark_log):
     with open(pages_path, 'w') as file:
         for item in jsonl_rows:
             file.write(json.dumps(item) + '\n')
+    jsonl_write_duration = time.time() - jsonl_prep_start_time
 
     # call js-wacz in a subprocess
-    exception_occurred = False
+    conversion_start_time = time.time()
+    conversion_error = False
     error_output = ''
     subprocess_arguments = [
         "npx",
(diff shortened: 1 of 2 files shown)


perma_web/perma/celery_tasks.py
```diff
@@ -1257,12 +1257,19 @@ def convert_warc_to_wacz(input_guid, benchmark_log):
...
+    warc_not_found_error = False
 
     # save a local copy of the warc file
-    with open(warc_path, "wb") as file:
-        copy_file_data(default_storage.open(link.warc_storage_file()), file)
+    warc_save_start_time = time.time()
+    try:
+        with open(warc_path, "wb") as file:
+            copy_file_data(default_storage.open(link.warc_storage_file()), file)
+    except FileNotFoundError:
```
rebeccacremona: I think this situation should never come up: we should never run this task on Perma Links that failed and therefore lack WARCs. We have a data problem on stage, where the field that should indicate whether a Perma Link can be played back or not is sometimes wrong, but we confirmed that that is _not_ a problem on prod (and shouldn't be a problem in anybody's dev database). So I think we probably do not need to account for FileNotFound errors.

Are you thinking it is just good "defensive coding"? If yes, how about, instead of failing here with FileNotFoundError, instead doing some validation of the Link at the beginning, for instance, making sure that `cached_can_play_back` is True?

perma_web/perma/celery_tasks.py
```diff
@@ -1324,26 +1333,29 @@ def convert_warc_to_wacz(input_guid, benchmark_log):
...
             writer.writerow(row)
         else:
             row["conversion_status"] = "Success"
             writer.writerow(row)
 
-    if not exception_occurred and not wacz_size_error:
-        with open(wacz_path, 'rb') as wacz_file:
-            default_storage.save(link.wacz_storage_file(), wacz_file)
+            with open(wacz_path, 'rb') as wacz_file:
+                default_storage.save(link.wacz_storage_file(), wacz_file)
```
rebeccacremona, line 1353: Nice improvement!

perma_web/perma/celery_tasks.py
teovin: Yep, it was mostly defensive, but I will just remove it since it sounds like we won't run into this in prod.

perma_web/tasks/dev.py
```diff
@@ -1158,5 +1168,15 @@ def benchmark_wacz_conversion(ctx, benchmark_log, source_csv=None, single_warc=N
...
                 convert_warc_to_wacz.delay(line[0], log_file)
-    else:
+    elif single_warc:
         convert_warc_to_wacz.delay(single_warc.split('.')[0], log_file)
+    else:
+        links = Link.objects.filter(
+            is_private=False,
+            is_unlisted=False,
+            cached_can_play_back=True
+        ).order_by('guid')[:batch_size]
```
rebeccacremona: Nice! I have a tiny suggestion here that will make this use less RAM.

When you do a normal `Model.objects.filter(...)` with Django, it retrieves all the fields from the database table (`SELECT * FROM the_model's_table WHERE ......`), and then it makes one python object per retrieved row, with all those fields populated onto the object so that you can access the field's value with the `link.guid` syntax.

That can end up using up a lot of RAM, when the number of retrieved rows/objects is large.

There are two things we can do to lighten the load.

One is, to use Django's `iterator` https://docs.djangoproject.com/en/5.0/ref/models/querysets/#iterator, which is designed specifically for this situation: when you have a big queryset, but all you want to do is iterator over it one by one.

The other is an optimization we can make because in this situation, the only thing we need to know is the guid of the links in question: we just need that one piece of info, so we can pass it to the celery task. In this case, having the whole object around is more than we need. For that, Django offers `values_list` https://docs.djangoproject.com/en/5.0/ref/models/querysets/#values-list.

We use `values_list` in exactly this situation, in a couple other spots, for instance, https://github.com/harvard-lil/perma/blob/a3464914d55426da248aba8767e2d5efed1c6dc3/perma_web/perma/celery_tasks.py#L373

So here, to use less RAM, we could combine the techniques, and do something like:
`Link.objects.filter(...).values_list('guid', flat=True).order_by('guid')[:batch_size].iterator()`

Pretty neat, right????

rebeccacremona (changes requested):
Looks great! I just have one performance-related suggestion 🙂. Really looking forward to running this experiment!!

perma_web/perma/celery_tasks.py
rebeccacremona: Cool beans.

teovin:
> Looks great! I just have one performance-related suggestion 🙂. Really looking forward to running this experiment!!

Thank you for the great explanation and the change is pushed! :)

rebeccacremona (approved):

PR #3507, opened 2024-04-29
Add custom select

tinykite:
## What this does
- Adds back previous dispatch code so that legacy app updates to folder tree data can now also update store
- Determines if a user is org-affiliated and/or sponsored
- Fetches folder info for user based on user affiliations (both org and sponsored folders)
- Adds a `FolderSelect` component. The component uses the term select rather than dropdown for a few reasons. 
    - Commonly, custom components that include listbox or combobox ARIA roles in design systems such as Adobe Spectrum are referred to as simply a custom select. Because they typically attempt to re-implement the behavior of a native select, I think this general name makes the most sense. 
    - A custom select is also the name that the lead accessibility engineer Gerard Cohen felt was most appropriate for his talk, ["Accessible Custom Selects"](https://m.youtube.com/watch?v=ug1elgso_8A) which the implementation of this custom select borrows some advice from. 
- Removes old `HelloWorld` components
- Determines current state of the trigger for the folder select
- Adds incremental accessibility improvements that include custom select interactions as well as additional accessibility roles and aria attributes
  - These improvements are incremental, with the intention that we will revisit them when we have the capacity for a redesign. 
  - I added a hook from the [VueUse](https://v7-7-1.vueuse.org/) library (`onClickOutside`) to handle clicks outside of the custom select as part of this work. In the future, I think VueUse could be super helpful for many other things as well — so it felt like a reasonable addition.  
- Dispatches to the legacy application when a selection changes
- Includes selected folder id with link creation API requests

### Note on custom events
- Perma currently makes use of Jquery-specific custom events, which appear to differ from native custom events. For example, I noticed I wasn't able to listen to them using vanilla JS event listeners, and importing legacy functions that trigger Jquery events will successfully import and call the function itself but not trigger Jquery event it wraps. A super easy workaround for this appears to be to create native custom events and dispatch them to or from legacy portions of the application instead. 

### Interaction notes 
- Hitting spacebar or return on the select trigger/button will toggle the select
- Arrow keys can be used to navigate the select options. The options do not cycle — if you reach the end of the list, hitting the down arrow will not navigate to the beginning of the list. The same is true for hitting up on the first option. (These behaviors match native select behavior)
- Home and end keyboard buttons will navigate to the first and last options, respectively 
- Hitting space or return on an option will select it, and then return focus back to the trigger for the select
- Hitting escape, or clicking outside of the select, will hide the options and also return focus back to the trigger


perma_web/frontend/components/LinkCount.vue
```diff
@@ -14,7 +14,7 @@ const getLinksRemainingText = () => {
 
         case 'metered':
         default:
-            return `You have <span class="links-remaining">${links_remaining}</span> remaining Perma Links${linkRenewalPeriod}</span><br />`
+            return `You have <span class="links-remaining">${globalStore.linksRemaining}</span> remaining Perma Links${linkRenewalPeriod}</span><br />`
```
tinykite, line 17: If we are going to refer to a variable in more than one component, it felt like it might make sense to include it in the store.

perma_web/frontend/lib/store.js
```diff
@@ -12,14 +13,57 @@ export const getLinksRemainingStatus = (linksRemaining, isNonpaying) => {
...
+        globalStore.updateFetchErrorMessage(errorMessage)
+    }
+
+    if (data?.value?.objects.length) {
+        const organizationsById = data.value.objects.reduce((acc, currentValue) => {
+            return { ...acc, [currentValue.id]: currentValue }
+        }, {})
+        globalStore.updateOrganizationFolders(data.value.objects)
+        globalStore.updateOrganizationFoldersById(organizationsById)
+    }
```
tinykite: FYI: I don't believe we currently need `organizationsById` just yet, but this was generated by the legacy code so it felt like an easy addition. I imagine it might be more useful in the future, when more functions are modernized.

perma_web/frontend/components/App.vue
```diff
@@ -1,12 +1,25 @@
...
+    globalStore.updateLinksRemaining(links_remaining)
     getLinksRemainingStatus(links_remaining, is_nonpaying)
-    getUserTypes(is_individual)
+    getUserTypes(is_individual, current_user)
 })
+
+watchEffect(() => {
+    if (globalStore.userTypes.includes('orgAffiliated')) {
+        getOrganizationFolders()
+    }
```
tinykite: In the original implementation, it appears we fetch organization folders for both org and sponsored users — but the fetch doesn't return data for sponsored users. So I if I'm not mistaken, this more conditional fetch might make more sense?

perma_web/frontend/components/FolderSelect.vue
```diff
@@ -0,0 +1,159 @@
...
+const selectContainerRef = ref(null)
+const selectButtonRef = ref(null)
+const selectListRef = ref(null)
+const isSelectExpanded = ref(false)
+const selectLabel = computed(() => globalStore.selectedFolder.path.length ? globalStore.selectedFolder.path.join(" > ") : 'Please select a folder')
+
+const folders = computed(() => globalStore.organizationFolders.concat(globalStore.sponsoredFolders))
+
+// Required for admin users locally, where admins have an "empty root folder" instead of personal links
+const personalFolderNames = ['Empty root folder', 'Personal Links']
```
rebeccacremona: Since this is just a [dumb detail of our test fixtures](https://github.com/harvard-lil/perma/blob/develop/perma_web/fixtures/folders.json#L6), I think it's not worth accounting for in the code. 

I think I see `personalFolderNames` only being used in two spots, both times, to check `personalFolderNames.includes(globalStore.selectedFolder.path[0])`. Since you have `const personalFolderId = current_user.top_level_folders[0].id`, are you able to check the folder's id  in those spots instead of its name, and then eliminate this? I haven't read closely enough to understand if there is a subtlety I'm not thinking of :-)

perma_web/frontend/components/FolderSelect.vue
```diff
@@ -0,0 +1,159 @@
...
+const handleClose = () => {
+    isSelectExpanded.value = false
+    selectButtonRef.value.focus()
+}
+
+const handleSelection = (e) => {
+    const isSpan = e.target.matches('span')
+    const target = isSpan ? e.target.parentElement : e.target
+
+    const { orgid, folderid } = target.dataset
```
rebeccacremona: In this stanza, I'm having a trouble reading through, keeping track of the difference between `orgid` and `orgId`, and similarly with `folderid` and `folderId`. I'm not sure what to suggest, but I'm wondering if there might be a way to rewrite this section such that such similar variable names aren't necessary. Do you think using `target.dataset['orgid']` and `target.dataset['folderid']` rather than nicknaming might be clearer? Or possibly doing the parsing inline, right inside the event definition? 🤷‍♀️ Do you have any thoughts here?

perma_web/frontend/lib/store.js
```diff
@@ -12,14 +13,57 @@ export const getLinksRemainingStatus = (linksRemaining, isNonpaying) => {
...
+export const getUserTypes = (isIndividual, user) => {
     if (isIndividual) {
-        types = types.concat('individual')
+        return globalStore.updateUserTypes('individual')
+    }
+
+    let userTypes = []
+
+    const topLevelFolders = user.top_level_folders
+    const isOrgAffiliated = topLevelFolders.some(folder => !!folder.organization)
```
rebeccacremona: Let's pass this info [straight to the javascript from the Django app](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/templates/user_management/create-link.html#L163-L173), rather than having to calculate it by iterating through all the user's folders. I think that will be much less expensive, especially for admins, who have lots and lots of top level folders :-)

Here's the current method for calculating `is_individual`: you'll see it actually checks all the varieties: https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/models.py#L997... 

We could gather this info efficiently, without too many round trips back and forth with the database, but for now, how about just passing the info straight to the JS without worrying about that? We can clean up afterwards.

perma_web/frontend/lib/store.js
```diff
@@ -12,14 +13,57 @@ export const getLinksRemainingStatus = (linksRemaining, isNonpaying) => {
...
+    const isSponsored = topLevelFolders.some(folder => folder.is_sponsored_root_folder)
```
rebeccacremona: Same, let's pass this info straight to the JS, rather than calculating it from the folders.

perma_web/frontend/lib/store.js
rebeccacremona: Yeah, I bet we can come up with a single data structure that serves both purposes, rather than using up memory holding storing this data twice. I'm less familiar with the idiomatic way to do this in Javascript, but in Python, a person could use an OrderedDict do the job of both: insert objects in order into the dict, and then call `dict.values` when you just want the list of objects, without the keys. Is there something similar that would serve here?

perma_web/frontend/lib/store.js
```diff
@@ -12,14 +13,57 @@ export const getLinksRemainingStatus = (linksRemaining, isNonpaying) => {
...
+    if (userTypes.length) {
+        return globalStore.updateUserTypes(userTypes)
+    }
+}
+
+export const getOrganizationFolders = async () => {
+    const { data, error, errorMessage } = await useFetch('/api/v1/organizations', {
+        limit: 300,
+        order_by: 'registrar, name'
+    })
```
rebeccacremona, line 44: I observe that this method `getOrganizationFolders`, and the name of the variable in the global store `this.organizationFolders`, refers to "organization folders" rather than just "organizations", even though this is a call to the organization API route, but that the same pattern is followed for the naming of `getSponsoredFolders` and `this.sponsoredFolders`. Do you think the similarity of their names suggests that they retrieve and store data of the same type/shape? Would `getUserOrganizations` maybe be a good way to describe what this method is doing?

rebeccacremona (changes requested):
Exciting to see this working! Way cool to see the communication with legacy components working.

I added some comments with a few thoughts; let me know if they don't make sense 😄

perma_web/frontend/lib/store.js
tinykite: For sure, there are definitely ways to iterate on a dictionary like structure in JavaScript — and keep it as both a dictionary and iterate on it like an array. This isn't super common in my experience to do, which might simply be in part due to the fact that until ES2020 the order of an object wasn't guaranteed (and differed in implementation between Webkit and other browsers, I think). But it could definitely make sense!

I deleted this in c017fa75b5f6edf32916c32d332595bcce2e0316 for now, because we don't need it, and think we can easily revisit it in the future. Does that feel reasonable?

perma_web/frontend/components/FolderSelect.vue
tinykite: Great callout — I refactored this in d13e29fd1916a2bddb9482687d84158eae414400.

perma_web/frontend/components/FolderSelect.vue
tinykite: For sure, I definitely wish data-attributes didn't require all lowercase names. I think there are two possible approaches that might make this easier to read: 

1. We could rename the variables inline at the same time they are destructured from `target.dataset` (and entirely avoid the need for the additionally declared variables)  

```js
const { orgid: orgId, folderid: folderId } = target.dataset

    if (!folderId) {
        return handleClose()
    }

    // Call a custom event that triggers triggerOnWindow function
    const updateSelections = new CustomEvent("dropdown.selectionChange", { detail: { data: { folderId: JSON.parse(folderId), orgId: orgId ? parseInt(orgId) : null } } });
    window.dispatchEvent(updateSelections);
```

Renaming variables at the same time they are destructured like this is pretty common, and pretty concise. 

2. But another totally reasonable option is to not use destructuring at all, like you mentioned:  
```js
    if (!target.dataset.folderid) {
        return handleClose()
    }
    
    const updateSelections = new CustomEvent("dropdown.selectionChange", { detail: { data: { folderId: JSON.parse(target.dataset.folderid), orgId: target.dataset.orgid ? parseInt(target.dataset.orgid) : null } } });
    window.dispatchEvent(updateSelections);
```

Which do you find easier to read?

I initially penciled the first approach in, in c017fa75b5f6edf32916c32d332595bcce2e0316, but happy to pursue whatever's easier.

perma_web/frontend/lib/store.js
tinykite: For sure! Great callout — refactored this in 9b238bb1bc24b2404c7950532c6837ce224fd403 to pull this information directly from the user response object. 

I also pulled in `is_staff` in [5b61873](https://github.com/harvard-lil/perma/pull/3507/commits/5b618738af2477d0630640656ac32490a59d77eb) because I noticed that admin users didn't have a folder select that was populating with all organization data.

perma_web/frontend/lib/store.js
tinykite: Also refactored in https://github.com/harvard-lil/perma/commit/9b238bb1bc24b2404c7950532c6837ce224fd403

perma_web/frontend/lib/store.js
tinykite: This is a great question. I renamed the function to `getUserOrganizations` and renamed the global store-related state and update method to match that naming in [c017fa7](https://github.com/harvard-lil/perma/pull/3507/commits/c017fa75b5f6edf32916c32d332595bcce2e0316)

tinykite:
Thank you for the feedback @rebeccacremona! I believe all outstanding questions have been answered — can you take another look?

perma_web/frontend/lib/store.js
rebeccacremona: > and differed in implementation between Webkit and other browsers
lol of course, isn't that always the way?

Cool, thanks yes, this sounds totally reasonable to me :-)

perma_web/frontend/components/FolderSelect.vue
rebeccacremona: Great, this version seems much easier to follow to me! Thanks!

perma_web/frontend/lib/store.js
rebeccacremona: 👍 neato.

question: have you checked to see how this works for registrar users? only asking to make sure, since the attribute in JS is now `isOrganizationUser` rather than `isOrgAffiliated`. Do you need to explicitly account for registrar users now too?

tinykite: Ah, I had checked — but I was testing with a registrar user I'd previously also marked as staff (so I saw the org folders, but not because I was handling registrar users specifically). This should now be handled in [183aff7](https://github.com/harvard-lil/perma/pull/3507/commits/183aff736da6741ded954b0ffa047b79ac133f5b)

rebeccacremona (approved):

PR #3515, opened 2024-05-09
Add dismissable perma tools info

tinykite:
## What this does 
- Adds dismissible information about the Perma Browser Tools. 
- Adds a css utility class, following the spacing tokens scale previously defined in the CAP static project. 

## Screenshots
Paragraph is visible by default, unless a user has previously clicked the close icon.
Screenshot 2024-05-09 at 3 27 12 PM

The position of the message shifts slightly for individual users: 
Screenshot 2024-05-09 at 3 39 21 PM

The visibility of the content depends on the "perma_tools_reminder" key, which follows the naming pattern of the local storage object, "perma_selection"
Screenshot 2024-05-09 at 3 27 30 PM

## Notes 
- We were previously relying on a cookie to determine if this text is visible. Because this is a user preference that is exclusively used client-side in the browser, and doesn't need to be included in API requests, it felt like it might make more sense to include it in local storage, which we were also already using for other client-side UI preferences. 
- There was some overlap between the legacy "Create a New Perma Link" header and the dismissible content that was preventing pointer events from being recognized on the close button. The easiest way to avoid this without introducing unnecessary changes to the legacy application seemed to be to introduce a utility CSS class, that we could easily add and remove if needed.

### DIFF
--- a/perma_web/frontend/components/CreateLink.vue
+++ b/perma_web/frontend/components/CreateLink.vue
@@ -158,7 +158,7 @@ onBeforeUnmount(() => {
                     <FolderSelect v-if="!globalStore.userTypes.includes('individual')" />
                 </fieldset>
                 <p v-if="!isToolsReminderSuppressed" id="browser-tools-message" class="u-pb-150"
-                    :class="globalStore.userTypes === 'individual' ? 'limit-true' : 'limit-false'">
+                    :class="globalStore.userTypes === 'individual' && 'limit-true'">
                     To make Perma links more quickly, try our <a href="/settings/tools">browser tools</a>.
                     <button @click.prevent="handleSuppressToolsReminder" type="button"
                         class="close-browser-tools btn-link">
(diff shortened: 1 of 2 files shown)


perma_web/frontend/components/CreateLink.vue
```diff
@@ -150,6 +157,15 @@ onBeforeUnmount(() => {
                     <LinkCount v-if="globalStore.userTypes.includes('individual')" />
                     <FolderSelect v-if="!globalStore.userTypes.includes('individual')" />
                 </fieldset>
+                <p v-if="!isToolsReminderSuppressed" id="browser-tools-message" class="u-pb-150"
+                    :class="globalStore.userTypes === 'individual' && 'limit-true'">
```
tinykite, line 161: The original legacy code added either `limit-true` or `limit-false` as classes to this paragraph depending on if a user was an individual. I noticed that there is no apparent `limit-false` class defined in the global css. So only `limit-true` is added here.

rebeccacremona (approved):
Cool.

The logic for switching from cookies to local storage sounds reasonable to me! Only thing is that means all users will have this preference unset when we switch over. If we're indeed okay with that: we should probably make sure that the linked "browser tools" page is up to date and is ready for a bunch of fresh views. Something to bring up with the product team, methinks.

tinykite:
> Only thing is that means all users will have this preference unset when we switch over. If we're indeed okay with that: we should probably make sure that the linked "browser tools" page is up to date and is ready for a bunch of fresh views. Something to bring up with the product team, methinks.

That's a reasonable ask — I'll touch base with Clare.

teovin (approved):

PR #3517, opened 2024-05-13
Validate and repair folder trees

rebeccacremona:
Perma's folders are organized into a tree courtesy of [django-mptt](https://github.com/django-mptt/django-mptt). 

Every folder keeps track of its `parent` and has exactly one `tree_id`: a supposed-to-be-unique id associated with the parentless folder at the "top" of each root branch. For efficient lookups, each folder also has three fields, `lft`, `rght`, and `level`, that help make it easier to tell at a glance what the shape of the tree is.

The tree_ids, lfts, rghts, and levels of many of our trees have gotten messed up, because [it is the application's responsibility to ensure that two move operations are never run at the same time](https://django-mptt.readthedocs.io/en/latest/technical_details.html#concurrency)... something that goes beyond running in transactions... and we haven't been doing that, all this time.

This PR includes all the code I have written so far, while diagnosing and attempting to repair the current folder trees. See ENG-743 and its sub-issues for the full adventure :-)

It makes only one actual change to the codebase: everything else is just for reporting. 

It changes the sorting/insertion order of folders from just `name` to `name, id`, so that the order is deterministic, when a particular tree has folders with identical names: https://github.com/harvard-lil/perma/blob/7336c4bb9eb1259b6d0e1e8b03689f1f8354504c/perma_web/perma/models.py#L1335

That situation should be rare... but, because we know that sometimes folder insertion is extremely slow (another problem I'll be working to solve shortly), we know that people sometimes repeat their attempts, resulting in multiple folders.

I do not think that this change will have a negative effect on performance or anything else, but I could be failing to think of something.

We'll plan to run `delete_redundant_personal_links_folders` and `delete_redundant_org_folders` in production as part of a larger process to repair all the folder trees, a process we expect to take around 10 min.

### DIFF
--- a/perma_web/perma/models.py
+++ b/perma_web/perma/models.py
@@ -1332,7 +1332,7 @@ class Folder(MPTTModel):
 
 
     class MPTTMeta:
-        order_insertion_by = ['name']
+        order_insertion_by = ['name', 'id']
 
     def is_empty(self):
         return not self.children.exists() and not self.links.exists()
--- a/perma_web/tasks/dev.py
+++ b/perma_web/tasks/dev.py
@@ -259,6 +259,7 @@ def delete_redundant_personal_links_folders(ctx, dry_run=False):
             assert user.root_folder_id in [folder.id for folder in folders]
             [redundant] = filter(lambda f: f.id != user.root_folder_id, folders)
             assert redundant.is_empty()
+            assert not redundant.contained_links().count()
         except AssertionError:
             print(f"Skipping user {user.id}: their situation isn't accounted for. Please investigate.")
             skipped = skipped + 1
--- a/perma_web/tasks/dev.py
+++ b/perma_web/tasks/dev.py
@@ -141,6 +141,27 @@ def validate_folder_trees(ctx,
         print_errored_ids=False,
         print_exceptions=False,
         print_changed_nodes=False):
+    """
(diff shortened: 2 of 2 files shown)


bensteinberg (approved):
Cool!

(When I see `x = x + 1` I think of using `x += 1` instead, not sure if one is more readable/pythonic than the other.)

rebeccacremona:
> Cool!
> 
> (When I see `x = x + 1` I think of using `x += 1` instead, not sure if one is more readable/pythonic than the other.)

Yeah! I have something in my head warning me about incrementing/decrementing syntactic sugar working differently in different languages, and something about timing (maybe in loops?) that has scared me off ever using it. I should... find out what I'm vaguely remembering, rather than being so broad.

jcushman:
As a shot in the dark, I'll bet you're thinking of pre-increment and post-increment operators like `a = ++b` and `a = b++` from C and C-like languages like javascript -- which have the trickiness of a variable on the right side of `=` being modified, and the trickiness of pre-increment and post-increment doing different things to `a`. This would let you do weird stuff in C-style loops like `for(i=0;i++ < j;){}`.

If it's this, then the good news is none of these vibes reflect on python at all -- operator-assignment doesn't have the same issues.

rebeccacremona:
Ha, that has got to be it! 😁 Thanks Jack!

PR #3520, opened 2024-05-18
Add a secondary storage to support wacz experiments.

rebeccacremona:
We've been working on plans to start capturing and playing back WACZ-formatted archives, either in addition to or instead of WARC-formatted. We've also been experimenting with converting our existing archives to WACZ.

To experiment more safely, we'd like to have another S3 bucket available, one that we can feel comfortable messing around with, deleting artifacts from, etc... entirely separate from where we keep the WARCs for Perma's actual collection, which, appropriately, has very strict rules.

This PR adds config for an additional "secondary" storage, and locally wires it up to minio.

It should be safe to deploy without changing stage or prod config, and without creating any additional cloud storage: we can do that in the future, if we have any code that actually uses the secondary storage that we want to try out in stage or prod.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -92,6 +92,7 @@ services:
       - MINIO_ROOT_PASSWORD=secretkey
       - DATA_DIR=/data
       - BUCKET=perma-storage
+      - SECONDARY_BUCKET=perma-secondary-storage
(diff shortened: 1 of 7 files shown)


matteocargnelutti (approved):
@rebeccacremona Nice!

PR #3521, opened 2024-05-20
Add multiple captures feature

tinykite:
## Summary 
Rebuilds the multiple link batch capture feature within Vue.

## Key updates
- Adds a `CreateLinkBatch` component, which encapsulates much of the new functionality
- Adds a new custom event dispatch to the legacy application when a batch capture is started. This will trigger an update to the batch capture history list. 
- Adds a new, reusable `Dialog` component that can be controlled by an independent trigger element.
   - The Dialog component supports a default `slot`: any content can be passed to it as children.
- Adds a `LinkBatchDetails` component. Combined with the current Dialog, this should make it incrementally easier to rebuild the modals a user can trigger from the batch capture history list.
- Refactors`FolderSelect` component to conditionally render either a `FolderCustomSelect` or a `FolderNativeSelect`, based on need. Both components get passed the same props.
- Adds a new `FolderNativeSelect` component for folder selection with the batch upload dialog.
- Updates `Spinner` component to accept additional props to make it more reusable
- Adds additional batchCapture states to the global store
- Adds component classes, following BEMIT naming practices, for dialog and button elements.

## UI Changes
1. As discussed in Slack, I made a small change to the select element included for folder selection: 
- Previously the select would always display all top-level folders. And if a user selected a nested folder from the folder tree UI, it would display that selected option plus all child sub-folders of the affiliated top-level folder. 
- What has now changed is that the select will still always display all top-level folders. If a user selects a nester folder from the folder tree UI, however, only that additional option will populate within the folder select — which matches the behavior of the custom select. We plan to revisit this in the future during a redesign. 

2. There were two bugs I noticed in the legacy batch capture UI: 
- If a batch capture included no successful captures, it would display that the batch progress was "NaN" complete. 
- For every completed url batch I tested, the total batch progress was never calculated higher than 67%.
- An easy fix for both issues that didn't require much of a refactor is to render "batch complete" instead of "batch [percent] complete" when all links in a batch have either captured successfully or failed. 

## Screenshots 
The batch capture trigger ("create multiple links") looks like a link, but is semantically a button. It will change styling based on if the selected folder is private:
Screenshot 2024-05-21 at 9 44 51 AM

The new dialog element minimally wraps the previous modal styling:
Screenshot 2024-05-21 at 9 45 04 AM

For now, if a capture fails the error will be counted — and communicated inline — but the specific error (for example, an invalid URL) is not intended to be rendered just yet. This will happen in follow-up work:
Screenshot 2024-05-21 at 9 45 20 AM

If a capture succeeds, the dialog will display a link to view the completed archive. Additionally, if all captures are complete, an option to download a CSV describing the capture data for all captures is provided:
Screenshot 2024-05-21 at 9 45 41 AM

After a batch capture is initiated, the new Vue app will dispatch to the existing batch capture list area of the application, which should update with a new entry for each batch capture. Clicking the batch capture link will display batch capture details in the legacy modal: 

Screenshot 2024-05-21 at 11 42 28 AM
Screenshot 2024-05-21 at 11 45 06 AM

## To Do 
- Formatting and rendering specific errors. This is a task that is also outstanding for single link captures, so it felt like it would make sense to create additional linear issues for it and tackle separately. In Linear, I split this out into two additional issues. 
  - One issue for the toast element that can be triggered (that appears and then disappears quickly)
  - One issue for the inline errors that appear about specific links in the dialog itself.
- In the future, during a redesign, there are many opportunities I noticed to improve accessibility.


perma_web/frontend/components/CreateLinkBatch.vue
```diff
@@ -0,0 +1,216 @@
...
+    if (data.value.capture_jobs) {
+        const { allJobs, progressSummary } = handleBatchFormatting(data.value.capture_jobs)
+        batchCaptureJobs.value = allJobs
+        batchCaptureSummary.value = progressSummary
+    }
+}
+
+const handleBatchFormatting = ((captureJobs) => {
+    const steps = 6
+    const allJobs = captureJobs.reduce((accumulatedJobs, currentJob) => {
```
rebeccacremona, line 132: Slick use of `reduce`, very tidy!

perma_web/frontend/components/CreateLinkBatch.vue
```diff
@@ -0,0 +1,216 @@
...
+
+const handleBatchDetailsFetch = async () => {
+    console.log('batch details fetch initiated')
+    const { data, error, errorMessage } = await useFetch(`/api/v1/archives/batches/${batchCaptureId.value.id}`)
+
+    if (error) {
+        return globalStore.updateBatchCaptureErrorMessage(errorMessage)
+    }
+
+    if (data.value.capture_jobs) {
```
rebeccacremona: Am I right that, if `!data.value.capture_jobs`, something is really broken? Is that a condition we need to protect against?

perma_web/frontend/components/Dialog.vue
```diff
@@ -0,0 +1,32 @@
+<script setup>
```
rebeccacremona, line 1: Love this as a reusable component.

perma_web/frontend/components/FolderCustomSelect.vue
```diff
@@ -0,0 +1,152 @@
...
+    handleClose()
+}
+
+
+</script>
+
+
+<template>
+    <div id="organization_select_form">
+        <span class="label-affil">This Perma Link will be affiliated with</span>
```
rebeccacremona: Does this need some way to know whether it's used in the link batch modal or in the regular component, and switch from `<span class="label-affil">This Perma Link will be affiliated with</span>` to `<span class="label-affil">These Perma Links will be affiliated with</span>` accordingly?

perma_web/frontend/components/LinkBatchDetails.vue
```diff
@@ -0,0 +1,62 @@
...
+              <div v-else class="link-desc col col-sm-6 col-md-60">
+                <div v-if="job.user_deleted" class="failed_header">Deleted</div>
+                <div class="item-title">{{ job.title }}</div>
+                <div class="item-subtitle">{{ job.submitted_url }}</div>
+              </div>
+              <div class="link-progress col col-sm-6 col-md-40 align-right item-permalink">
+                <span v-if="job.status === 'pending'">Queued.</span>
+                <ProgressBar v-if="job.status === 'in_progress'" :progress="`${job.progress}%`" />
+                <span v-if="job.status === 'completed'">
+                  <a class="perma no-drag" :href="job.guid ? `/${job.guid}` : null">
```
rebeccacremona: Am I right that if a Capture Job  is `completed`, but has no `guid`, something is very broken? Is that a condition we need to check for?

rebeccacremona (commented):
Huzzah! Looks great. Fun to see this more complex component added to the mix!

I left small questions about a few lines, but nothing major 🙂.

### Folder selection dropdown

Yeah, the only reason the native select currently works the way it does is, it was easiest for us to reuse the select that appears in the Link List, when you click to toggle and reveal Link Details. That select is used for moving links between folders, as an alternative to drag-and-drop... and so it needs to have subfolders. It's built [by the folder tree module](https://github.com/harvard-lil/perma/blob/48666fab86bec2c893ba373d03f53db98674f629/perma_web/static/js/helpers/folder-selector.helper.js#L3), using features that come with that JQuery folder tree module, which is savvy to looping through tree structures. Without recourse to that... yeah, can definitely understand why you'd want to implement something simpler here.

When a subfolder is selected, I personally find it confusing that the name of the subfolder looks just like the top-level folder options, like here 
image

I would expect there to be some indication that it was an outlier / some kind of way-finding, just like in the main dropdown, where the path to the folder is displayed.

What would you think of doing that here as well, E.g. `Test Library > Subfolder > Selected Subfolder` instead of plain `Selected Subfolder`?

### Submitting without selecting a folder

I noticed that you can submit this form without selecting a folder, and if you do, the modal just closes. I know you are planning to add error handling later, separately, but wanted to flag in case that isn't how you wanted this for the time being.

Similarly, I noticed that you can use the main create component to make Perma Links without selecting a folder. I'd consider that one a bug, since the API will not be returning validation error there: only GUI users are required to specify a folder, when making a single link, which is only enforced by client-side code. (Whereas for link batches, folder is mandatory regardless of whether you are using the GUI or are using the API directly.)

Currently we handle both those situations by disabling form submission until you select a folder, which I know you are against as a design pattern, for well-documented UX reasons. 

It could be you want to handle these in different ways, but just wanted to mention here in case you want to handle them in the same/very similar ways!

### Bug while waiting for 200 after POST

Looks like the "batch complete" and folder message are rendering along with the spinner, right when you hit submit, before the API returns a response:
image

### Type warning

I saw this go by in the console, thought I'd share
image

### Finally.....did I see what I saw????

Finally... I can't seem to reproduce this, but I swear that on one occasion, when I had a top-level folder selected, it appeared twice in the new dropdown: once at the top, where selected subfolders are displayed, and then again in the normal part. But... I can't seem to get that to happen again! So maybe I'm wrong, and that's not what happened? Maybe it was some weird interaction between using both the legacy components and the new components, alternating, which will never happen in real life. I honestly don't know. But I just thought I'd mention, in case you ever saw that too, or in case you think there might be a repeatable way to get it to happen again. 

### Merging

These are all just "comments"; since this work is happening incrementally in lots of separate PRs, I don't have much of an opinion about whether any followups need to be included in this PR. Maybe rendering bug, type warning, conditional "this link" vs "these links"? 

But I won't be so bold as to mark this "request changes" 😄. 

Again, fun to see this coming together!

perma_web/frontend/components/CreateLinkBatch.vue
tinykite: This is a fair question. I think this could be handled by removing the if statement here, and revising the one above it: 

```
    if (error || !data.value.capture_jobs) {
        globalStore.updateBatchCapture('batchDetailsError')
        globalStore.updateBatchCaptureErrorMessage(errorMessage)
    }
```

This is handled in [4049cc0](https://github.com/harvard-lil/perma/pull/3521/commits/4049cc013dad9cd1ce8953bdb37123ea81eec6ce)

perma_web/frontend/components/LinkBatchDetails.vue
tinykite: Also a fair question! This is a remnant of legacy code, and after thinking it through, I don't imagine that we need this conditional check in the modernized version.

perma_web/frontend/components/FolderCustomSelect.vue
tinykite: Good question: I think the simplest way to make this flexible enough for a need like that is to add a `selectLabel` prop. Then the component doesn't necessary need to be aware of the context that it's being used in, but will still show the appropriate label.

This is handled [e3ea364](https://github.com/harvard-lil/perma/pull/3521/commits/e3ea36459ae77a56e7fe535c2e72110cfa4299e7)

tinykite:
> What would you think of doing that here as well, E.g. Test Library > Subfolder > Selected Subfolder instead of plain Selected Subfolder?

Sure thing. This is handled in [45c280d](https://github.com/harvard-lil/perma/pull/3521/commits/45c280dbae90e198434dc63dc79ca0593c0223a6)

> Looks like the "batch complete" and folder message are rendering along with the spinner, right when you hit submit, before the API returns a response:

This is fixed in [d0734ed](https://github.com/harvard-lil/perma/pull/3521/commits/d0734ed7f37e5b0775ef80632f3765ea193eb6ce). The batch details won't be visible until the batch is officially queued.

> I saw this go by in the console, thought I'd share
Ope, this type mismatch was an oversight that happened from me generally having the Spinner component entirely commented out. Fixed in [fd54cf0](https://github.com/harvard-lil/perma/pull/3521/commits/fd54cf0086fb2f5d21443d6526dc5c154f0c4f00)

> I noticed that you can submit this form without selecting a folder, and if you do, the modal just closes. I know you are planning to add error handling later, separately, but wanted to flag in case that isn't how you wanted this for the time being.

I penciled in placeholder error handling for this in [d142069](https://github.com/harvard-lil/perma/pull/3521/commits/d1420699feed086748e9a9cce8104a2c0f6bebbe), so that it doesn't get lost or forgotten about while I'm working on error handling for both the single and batch capture. While this placeholder error handiling will still just close the dialog element without any UI indication for now for the batch captures, it will also actually render the correct error UI (along with a more-specific, pencilled-in message) for the single link capture. 

> Finally.....did I see what I saw????

I wasn't able to reproduce this issue across multiple test user accounts, however I did notice that if there was no selected folder set in cookies there was a bug where "Please select a folder" wouldn't be selected and an empty option would be added to the select. This has since been fixed (with [0978ec2](https://github.com/harvard-lil/perma/pull/3521/commits/0978ec2fac2ce0edbcac0a8c96ddd6de43106ebb) and [25fe0ac](https://github.com/harvard-lil/perma/pull/3521/commits/25fe0acd8fe23c803b43fd0b90a953213d465ee1)). 

I will absolutely continue to try to test this, and noodle about if there's anything I can think of that would only happen in one specific use case. One thing I have been thinking about that could potentially make it easier to reproduce bugs is exploring UI development in Storybook. Right now, as I develop specific UI features I try to regularly switch between user accounts, and do my due diligence to test across as many use case as possible. But it's certainly not a perfect process — and there's a lot of room for error when testing for one user and then making a tiny fix while testing for another.

Storybook is a framework-agnostic tool intended to help with this — I don't have a proposal for its use just yet, but might be fun to experiment with during 20% time on a Friday. 

One more thing: I noticed there's a couple tiny merge conflicts. I'll handle them early tomorrow morning and plan to double-check everything is working as expected before merging.

PR #3524, opened 2024-05-23
Add toast component

tinykite:
## What this does 
Adds a component and composable hook for displaying toast notifications. The component currently supports the following: 

- Toast notification message 
- Toast state (For example "success" or "error") 

It's somewhat common in web applications for one or more toast notifications to be visible on the page at one time. For this reason, the toast component is structured to support rendering more than one unique toast notification at a time.

The current Perma CSS styling for a toast will simply stack the most recent toast on top of a previous one. However, if this is redesigned in the future, and we have a need for "stacked" toast notification, we will have the capability for it.

The legacy toast additionally features a subtle animation when it is added to the page. I recreated a subtle fade-in transition for the new component using Vue's built-in `TransitionGroup` wrapper and namespaced css style so Vue can programmatically pick them up. 

## Screenshots 
For the purpose of testing, I added a button to the debug area of the new dashboard: 
image

Clicking it will trigger a toast with dummy text that includes a unique id generated from a timestamp: 
image

## Future notes 
- Like with other UI work, there's definitely more that can be done to improve accessibility with this element during a future redesign.

### DIFF
--- a/perma_web/static/css/style-responsive.scss
+++ b/perma_web/static/css/style-responsive.scss
@@ -5863,3 +5863,14 @@ Example: calc(4rem / 16) is equal to 4px or 0.25rem */
 .u-pb-150 {
   padding-bottom: calc(18rem / 16);
 }
+
+/* CSS transition classes automatically added/removed by Vue */
+.fade-enter-active,
+.fade-leave-active {
+  transition: opacity 0.3s ease;
+}
+
+.fade-enter-from,
+.fade-leave-to {
+  opacity: 0;
+}
--- a/perma_web/frontend/components/App.vue
+++ b/perma_web/frontend/components/App.vue
@@ -1,5 +1,6 @@
 <script setup>
 import CreateLink from './CreateLink.vue';
+import Toast from './Toast.vue';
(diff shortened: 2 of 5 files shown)


perma_web/static/css/style-responsive.scss
```diff
@@ -5862,4 +5862,15 @@ li[data-read_only="true"] {
...
+/* CSS transition classes automatically added/removed by Vue */
+.fade-enter-active,
+.fade-leave-active {
+  transition: opacity 0.3s ease;
+}
+
+.fade-enter-from,
+.fade-leave-to {
+  opacity: 0;
 }
```
tinykite, line 5876: The hyphen-based naming here is required for Vue to be able to find these styles and conditionally add and remove them at the right times. The first part of each class name ("fade") refers to the optional `name` prop passed to the `TransitionGroup` wrapper. (Instead of only having one enter/exit CSS transition, developers can have multiple named transitions)

bensteinberg (commented):
LGTM! But I will add @rebeccacremona as a reviewer.

rebeccacremona (approved):
LGTM too!

PR #3529, opened 2024-05-31
Folder-related transactions, locks, and refactoring

rebeccacremona:
Here's another PR with a difficult-to-read diff. I think the individual commits are pretty clear, though, with the exception of the last one!

In preparing this PR, I read through all the application logic for creating new folders, updating folders' metadata, moving them, and deleting them, and tried to use transactions and locks (via `select_for_update`) to prevent the application from getting into an inconsistent state. I am hopeful this will prevent any further mangling of our folder trees.

I also thoroughly refactored the Folder model's `save` method, so that it is easier to reason about. Previously, we checked to see if a folder was new, if it had a parent or not, and if its parent had changed in a number of places. With this refactor, each case is handled entirely separately, and shared logic is extracted into reusable helper functions. I also renamed a bunch of stuff, and added comments. Since the diff is hard to read, I think the easiest way to review is to read the [new version, uninterrupted](https://github.com/harvard-lil/perma/blob/946ec3176154d54c2854186057ee0194776737d8/perma_web/perma/models.py#L1315-L1471), and if desired, compare to the [current version](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/models.py#L1307-L1452).

I was able to remove a few redundant calls to the database in the process; I doubt we'll notice much of a change in read ops, but you never know!

I don't have a lot of experience with explicit transactions or with locks; if something looks goofy, it's almost certainly wrong 😄. But, the tests are passing! I think so long as I haven't set us up for competing locks, we're good.

See ENG-854.

### DIFF
--- a/perma_web/api/views.py
+++ b/perma_web/api/views.py
@@ -226,7 +226,17 @@ class FolderDetailView(BaseView):
         """
         if not request.parent:
             raise_general_validation_error("PUT is only valid for nested folder endpoints.")
-        return self.folder_update(request, pk, {'parent': request.parent.pk})
+
+        with transaction.atomic():
+            # Lock this folder, so no one makes new subfolders inside it, during the transaction
+            this = Folder.objects.select_for_update().get(pk=pk)
+            # Lock the current subtree, so no one moves anything else around while we are making this move
+            Folder.objects.select_for_update().get(pk=this.tree_root_id)
+            # Lock the destination subtree, if it's different, for the same reason
+            if this.tree_root_id != request.parent.tree_root_id:
+                Folder.objects.select_for_update().get(pk=request.parent.tree_root_id)
+
+            return self.folder_update(request, pk, {'parent': request.parent.pk})
 
     @load_parent
     def delete(self, request, pk, format=None):
--- a/perma_web/api/views.py
+++ b/perma_web/api/views.py
@@ -195,8 +195,14 @@ class FolderListView(BaseView):
         data = request.data.copy()
         if request.parent:
             data.setdefault('parent', request.parent.pk)
-
-        return self.simple_create(data, {'created_by': request.user})
+        with transaction.atomic():
+            if data.get('parent'):
+                # Lock the parent to prevent anyone from deleting it while this operation is validated and saved.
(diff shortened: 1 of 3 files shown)


perma_web/api/views.py
```diff
@@ -195,8 +195,14 @@ def post(self, request, format=None):
...
         if request.parent:
             data.setdefault('parent', request.parent.pk)
-
-        return self.simple_create(data, {'created_by': request.user})
+        with transaction.atomic():
+            if data.get('parent'):
+                # Lock the parent to prevent anyone from deleting it while this operation is validated and saved.
+                parent = Folder.objects.select_for_update().get(pk=data['parent'])
+                # We don't want to insert any new folders while a tree's folders are being moved around.
+                # Since moves lock the whole subtree, we can take out the same lock, to ensure no move is underway.
```
bensteinberg, line 203: I'm not sure what "take out the same lock" means in this context.

bensteinberg (approved):
LGTM! I had one question about a comment. lints failed earlier, but I think you may have just fixed that.

perma_web/api/views.py
rebeccacremona: So the line under this comment is ` Folder.objects.select_for_update().get(pk=parent.tree_root_id)`. That is the identical `select_for_update` done before a move, [here](https://github.com/harvard-lil/perma/pull/3529/files#diff-2f3edeef9470b9bee8fe3a0e9ddd0ee941e96ff170c1662523b48041b8014d86R240). 

So, only one of these things can happen at a time: if the 204 line has the lock, then the 240 line will wait, and vice versa.

Can revise the comment... any suggestions for how to describe that more clearly?

bensteinberg: No, that makes sense. Thanks!

jcushman: if you had a `parent.lock_tree()` method that did `Folder.objects.select_for_update().get(pk=self.tree_root_id)`, you could use it in a few places and maybe some comments would seem superfluous. nbd!

rebeccacremona: Ah, that's a good idea. Might refactor in a future PR!

Thanks for taking a glance.

PR #3530, opened 2024-05-31
Fix Folder admin

rebeccacremona:
This PR adds 'tree_root' to the Folder admin's list of [raw id fields](https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.raw_id_fields), so that, when a folder's detail page renders, Django doesn't offer a `select` with every single Perma folder as an option 😂... which is a big enough request that the page doesn't currently load.


bensteinberg (approved):
👍

PR #3534, opened 2024-06-05
Use regular spaces, rather than non-breaking spaces, in formatted filenames

rebeccacremona:
This tweaks the output of two fields in our report about how things went, when converting a WARC to a WACZ.

The difference:
```
In [1]: from django.template.defaultfilters import filesizeformat

In [2]: filesizeformat(100)
Out[2]: '100\xa0bytes'

In [3]: filesizeformat(100).replace("\xa0", " ")
Out[3]: '100 bytes'
```

See ENG-923.

### DIFF
--- a/perma_web/perma/celery_tasks.py
+++ b/perma_web/perma/celery_tasks.py
@@ -1292,7 +1292,7 @@ def convert_warc_to_wacz(input_guid, benchmark_log):
             return f"{math.ceil(seconds_val)} seconds"
 
     def format_filesize(i):
-        filesizeformat(i).replace("\xa0", " ")
+        return filesizeformat(i).replace("\xa0", " ")
 
     #
     # Launch the conversions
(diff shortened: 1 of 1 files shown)


perma_web/perma/celery_tasks.py
```diff
@@ -1287,6 +1277,27 @@ def convert_warc_to_wacz(input_guid, benchmark_log):
...
+        """
+        Converts seconds to minutes
+        """
+        if seconds_val >= 60:
+            return f"{round(seconds_val / 60, 1)} minutes"
+        else:
+            return f"{math.ceil(seconds_val)} seconds"
+
+    def format_filesize(i):
+        filesizeformat(i).replace("\xa0", " ")
```
teovin: Can we add a `return` here?

teovin (changes requested):

perma_web/perma/celery_tasks.py
rebeccacremona: Oooh, thanks, good catch!

rebeccacremona: Done!

teovin (approved):

PR #3536, opened 2024-06-06
Validate the WACZ produced during the conversion

rebeccacremona:
This is a very basic check to make sure that the WACZ came out okay, during our test/benchmarking WARC -> WACZ conversion experiment.

Check the hash of the original WARC against the hash of the WARC inside the WACZ. This won't tell us whether the archive plays back, but will at least ensure that we didn't lose any valuable data.

To find the remote hash and calculate the new one, I tweaked the logic of Ben's awesome [check_s3_hashes](https://github.com/harvard-lil/perma/blob/develop/perma_web/tasks/dev.py#L401C1-L453C95) task, written back when to help us migrate from local filesystem storage to S3 ❤️. 

See ENG-929.

### DIFF
--- a/perma_web/perma/celery_tasks.py
+++ b/perma_web/perma/celery_tasks.py
@@ -1,3 +1,4 @@
+import hashlib
 import itertools
 
 import os
@@ -13,6 +14,7 @@ from celery.exceptions import SoftTimeLimitExceeded
 from celery.signals import task_failure
 import requests
 from requests.packages.urllib3.exceptions import InsecureRequestWarning
+from zipfile import ZipFile
 
 requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
 
@@ -1350,6 +1352,29 @@ def convert_warc_to_wacz(input_guid, benchmark_log):
         wacz_size = 0
 
     formatted_wacz_size = format_filesize(wacz_size)
+    if wacz_size:
+        # retrieve S3's md5 hash of file it has
+        remote_hash = default_storage.connection.Object(
+            bucket_name=default_storage.bucket_name,
+            key=os.path.join(settings.MEDIA_ROOT, link.warc_storage_file())
+        ).e_tag.strip('"')
+
+        # calculate the same thing for the warc inside the wacz
+        blocksize = 2 ** 20
+        m = hashlib.md5()
+        with ZipFile(wacz_path) as zip_file:
+            with zip_file.open(f"archive/{link.guid}.warc.gz") as embedded_warc:
+                while True:
+                    buf = embedded_warc.read(blocksize)
+                    if not buf:
+                        break
+                    m.update(buf)
+
+        # see if they match
+        warc_checksums_match =  m.hexdigest() == remote_hash
+    else:
+        warc_checksums_match = None
+
     raw_total_duration = time.time() - start_time
     duration = seconds_to_minutes(raw_total_duration)
 
@@ -1366,21 +1391,29 @@ def convert_warc_to_wacz(input_guid, benchmark_log):
             "raw_warc_save_duration": warc_save_duration,
             "raw_pages_write_duration": pages_write_duration,
             "raw_conversion_duration": conversion_duration,
-            "error": ''
+            "error": '',
+            "warc_checksums_match": warc_checksums_match
         }
-        writer = csv.DictWriter(log_file, fieldnames=row.keys())
-
         if conversion_error:
             row["conversion_status"] = "Failure"
             row["error"] = error_output
-            writer.writerow(row)
-        elif wacz_size == 0 or warc_size > wacz_size:
+        elif wacz_size == 0:
             row["conversion_status"] = "Failure"
-            row["error"] = "WACZ is smaller than WARC"
-            writer.writerow(row)
+            row["error"] = "No WACZ produced"
         else:
-            row["conversion_status"] = "Success"
-            writer.writerow(row)
+            if warc_size > wacz_size:
+                row["conversion_status"] = "Failure"
+                row["error"] = "WACZ is smaller than WARC"
+            elif not warc_checksums_match:
+                row["conversion_status"] = "Failure"
+                row["error"] = "The WARC embedded in the WACZ differs from the source WARC"
+            else:
+                row["conversion_status"] = "Success"
 
             with open(wacz_path, 'rb') as wacz_file:
                 default_storage.save(link.wacz_storage_file(), wacz_file)
+
+        writer = csv.DictWriter(log_file, fieldnames=row.keys())
+        writer.writerow(row)
+
+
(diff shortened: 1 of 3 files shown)


perma_web/perma/celery_tasks.py
```diff
@@ -1350,6 +1352,29 @@ def format_filesize(i):
...
     formatted_wacz_size = format_filesize(wacz_size)
+    if wacz_size:
+        # retrieve S3's md5 hash of file it has
+        remote_hash = default_storage.connection.Object(
+            bucket_name=default_storage.bucket_name,
+            key=os.path.join(settings.MEDIA_ROOT, link.warc_storage_file())
+        ).e_tag.strip('"')
+
+        # calculate the same thing for the warc inside the wacz
+        blocksize = 2 ** 20
```
cmsetzer, line 1364: Out of curiosity, is there a reason we want this particular chunk size?

perma_web/perma/celery_tasks.py
```diff
@@ -1350,6 +1352,29 @@ def format_filesize(i):
...
+        m = hashlib.md5()
+        with ZipFile(wacz_path) as zip_file:
+            with zip_file.open(f"archive/{link.guid}.warc.gz") as embedded_warc:
```
cmsetzer: Two thoughts:

* Have we decided to use `{link.guid}.warc.gz` as the WARC filename instead of e.g. `data.warc.gz`? I am not very familiar with how people interact with WACZs aside from reading the spec; is this in line with expected file naming patterns?

* It's a matter of preference, but you could reduce the nesting by combining these into a single `with`:

    ```py
    with (
        ZipFile(wacz_path) as zip_file,
        zip_file.open(f"archive/{link.guid}.warc.gz") as embedded_warc
    ):
    ```

perma_web/perma/celery_tasks.py
```diff
@@ -1366,21 +1391,29 @@ def format_filesize(i):
...
-            row["conversion_status"] = "Success"
-            writer.writerow(row)
+            if warc_size > wacz_size:
+                row["conversion_status"] = "Failure"
+                row["error"] = "WACZ is smaller than WARC"
+            elif not warc_checksums_match:
+                row["conversion_status"] = "Failure"
+                row["error"] = "The WARC embedded in the WACZ differs from the source WARC"
+            else:
+                row["conversion_status"] = "Success"
```
cmsetzer, line 1417: Any reason not to remove the `else` on line 1403 and dedent these cases so they're part of the main `if` block?

cmsetzer (approved):
Looks good to me! I included a few questions/suggestions but nothing that should block merging.

perma_web/perma/celery_tasks.py
rebeccacremona: > Have we decided to use {link.guid}.warc.gz as the WARC filename instead of e.g. data.warc.gz? I am not very familiar with how people interact with WACZs aside from reading the spec; is this in line with expected file naming patterns?

So you can have lots of WARCs inside a single WACZ; they can be named according to whatever convention; data.warc.gz is just what Scoop does, probably because it's filename used as an example in the spec. 

Since our WARCs are currently named `{link.guid}.warc.gz`, if we pass them to js-wacz for conversion, it just keeps that untouched.

I couldn't think of any reason to switch; I asked Matteo if he could think of any reason to switch and he couldn't think of one either. 

I can see it both ways:
- if we introduce heterogeneity into our WACZ collection, then we'll have to account for that in the future, say, if we ever want to programmatically extract WARCs from WACZs... which could be annoying
- on the other hand, this distinction would make it possible to see, from the WACZ itself, without access to our database, whether the WACZ was itself produced by Scoop, or was packaged up after the fact... which could be useful/good to have

I'll add to a list of things to talk about one more time, if we choose to convert the corpus in earnest.

rebeccacremona: > It's a matter of preference,

Oh cool! I... If at any point I knew that licit syntax, I had forgotten. Will definitely tweak.

perma_web/perma/celery_tasks.py
rebeccacremona: I admit to extreme laziness: I simply copied the block size from Ben's old code and hoped that it still worked. I'll go see if there's a record of why, or if I can find S3 docs explaining. I think the story is something like "we need to use the same block size that S3 does, when it is calculating e_tags", but I'll find out if that's true, and will add a comment.

perma_web/perma/celery_tasks.py
cmsetzer: Re: WARC filename, that makes sense. I misunderstood the inclusion of `{link.guid}.warc.gz` here as a new change (I'm used to seeing the output on the Scoop side). I think it's fine to keep doing what we're doing.

rebeccacremona: It was a surprise to me! I wrote code to try and look at `data.warc.gz` and spent longer than I should have mystified by why it wasn't working 🙄 🤣

perma_web/perma/celery_tasks.py
rebeccacremona: Best I can tell testing locally.... it doesn't matter. I can't find any info on why we did it one megabyte at a time, last time.

Maybe I'll add a comment that says, "use a blocksize of 1MB, because it's LIL tradition"

perma_web/perma/celery_tasks.py
rebeccacremona: Yes: if we are in the 1403 `else`, it means there is a WACZ to save to storage (either because it came out well, or because we want to study how it didn't). In the other cases, no WACZ is present.

I can refactor or comment to make this clearer.

rebeccacremona:
Thanks for the review, @cmsetzer! Suggested tweaks forthcoming!

rebeccacremona:
Huh, why does flake8 in CI think that is a syntax error.... https://github.com/harvard-lil/perma/actions/runs/9422297235/job/25958258797?pr=3536

rebeccacremona:
I wonder which version of python the linting step is running. It must be an old one...

rebeccacremona:
Yep.

Ubuntu 20.04, which is running Python 3.8.10 https://github.com/actions/runner-images/blob/ubuntu20/20240603.1/images/ubuntu/Ubuntu2004-Readme.md

This style was added in [3.10](https://stackoverflow.com/a/46481946).

cmsetzer:
My bad! I thought we'd be solid since the app itself is running on our 3.11 image. Curious that the tests seem to run fine on Ubuntu 20.04, but maybe the new task code is just not getting imported.

rebeccacremona:
Yeah I didn't realize that the linting workflow uses the system python, rather than the python in the container! I think no big deal to upgrade! I might try upgrading the `test` workflow to a newer unbuntu some other time, since that one is so much more complex.

PR #3537, opened 2024-06-06
Add multiple error formatting for batch captures

tinykite:
Note: this WIP PR is currently a fork of my single capture error PR because it relies on shared utility functions — when it is merged, I'll rebase on the develop branch and remove the additional commits. 

## What this does 
Adds error handling for the multiple batch captures features of perma.

### 1. When a user attempts to create a batch capture job without submitting any urls: 
image

### 2. When a user submits a url, but it's invalid:
image

### 3. When a user submits multiple urls, and some of them are invalid: 
image

### 4. When a user is out of perma links: 
image

### 5. When a user does not choose a folder destination for their perma link batch: 
image

### 6. When a user does not appear logged in:
image
Note: I tested this by commenting-out the POST logic and manually throwing an HTTP status code of 401

### 7. When a user attempts to create a link batch and Perma responds with an HTTP status code other than 400 or 401: 
image
Note: I tested this by manually throwing an HTTP status code of 500 instead of attempting a POST

### 8. When an unexpected error happens during the batch submission process, and a batch capture job is not created: 
image
Note: I tested this by manually throwing `throw new Error()` instead of attempting a POST

### 9. When a url is valid, but the capture fails:
image

### DIFF
--- a/perma_web/frontend/components/CreateLinkBatch.vue
+++ b/perma_web/frontend/components/CreateLinkBatch.vue
@@ -8,7 +8,7 @@ import { useFetch } from '../lib/data';
 import LinkBatchDetails from './LinkBatchDetails.vue'
 import Dialog from './Dialog.vue';
 import { validStates, transitionalStates } from '../lib/consts.js'
-import { folderError, getErrorFromNestedObject, getErrorFromResponseStatus, missingUrlError } from '../lib/errors';
+import { folderError, getErrorFromNestedObject, getErrorFromResponseStatus, missingUrlError, getErrorResponse } from '../lib/errors';
 import { useToast } from '../lib/notifications';
 import { defaultError } from '../lib/errors';
 
@@ -97,8 +97,8 @@ const handleBatchCaptureRequest = async () => {
             })
 
         if (!response?.ok) {
-            const errorResponse = await response.json()
-            throw { status: response.status, response: errorResponse }
+            const errorResponse = await getErrorResponse(response)
+            throw errorResponse
         }
 
         const data = await response.json()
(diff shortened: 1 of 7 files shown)


perma_web/frontend/components/CreateLinkBatch.vue
```diff
@@ -88,10 +97,12 @@ const handleBatchCaptureRequest = async () => {
             })
 
         if (!response?.ok) {
-            throw new Error(response.statusText) // This is a placeholder for now
+            const errorResponse = await response.json()
```
rebeccacremona: I think this will also want to call the new `getErrorResponse` utility, so this code doesn't break if the error is a non-JSON 502, etc. Am I reading that right?

rebeccacremona (approved):
This also LGTM! 

I like the use of the toast to display error messages when the current UI just... closes the Link Batch modal without explanation.

As with the foundational PR, I left one question about a spot that might want to call `getErrorResponse` too, acknowledging I might just not be reading things correctly.

perma_web/frontend/components/CreateLinkBatch.vue
tinykite: Yup! Was temporarily waiting on #3531 before pulling this in. 

Made some small code consistency changes in [12da131](https://github.com/harvard-lil/perma/pull/3537/commits/12da13129c61b8f3cb127318ddd7f555daf73c7f) and implemented the updated error handling for details fetches in [b1a4c79](https://github.com/harvard-lil/perma/pull/3537/commits/b1a4c797c95a1f7c17cda0f9ef679e12e0cfe31a)

PR #3539, opened 2024-06-13
Update feature flag use

tinykite:
## What this does 
This update readies the release of the modernized Perma dashboard work that has been completed so far to a wider pool of users. 

## Implementation Details + Requirements
- We have a feature flag provided by the django waffle library that allows us to conditionally render the modernized, Vue-driven Perma dashboard. 
  - This is referred to as `vue-dashboard` in the code
- We need to expand the use of the `vue-dashboard` flag to either show the modernized create link UI -or- show the legacy version, not both to the same users. 
- We should also make it possible to open up the Vue dashboard to other special users for testing.
  - Previously, the modernized dashboard would only load for admins who also had the `vue-dashboard` flag set to true, although in staging it is set to true for everyone.

## Additional Improvements
- To avoid showing in-progress work to Perma users who aren't staff, I created a new computed Vue value called `isStaff`. 
  - Combined with the existing django waffle feature flag, this gives us the capacity to test the Vue dashboard with anyone we would like — but still keep continue to gate-off access to certain parts of the dashboard that are needed for staff only 
- I moved the in-progress upload form to beneath the "Create a Link" UI, to try to declutter the page a bit.
- I also removed the import of an unused built-in Vue utility function

### Regular users
Users without the django waffle flag set will only see the legacy create link portion of the dashboard. 
Screenshot 2024-06-13 at 9 56 59 AM
_Note: You can tell this is the legacy version because the submit button is disabled, which has been improved for accessibility in the modernized portion of the dashboard. I tested this by opening an incognito window and logging in as a regular user._

### Special testing group users (non-admin)
Users with the django waffle flag set, who are not admins, will see the updated portion of the dashboard that is ready for early testing.
Screenshot 2024-06-13 at 9 56 17 AM

### Admin users
Admins with the django waffle flag set will see everything
Screenshot 2024-06-13 at 9 55 30 AM

## How to Test
- Test loading the dashboard without the vue-dashboard flag. If you've already set the flag to true previously, an easy way to temporarily ignore it is to open an incognito window and login as a regular user. 
  - You should see the regular legacy Perma dashboard, with a disabled submit button in the "Create a New Perma Link" section of the dashboard
- Test loading the dashboard with the vue-dashboard flag.
  - If it wasn't previously toggled, you can toggle it by visiting https://perma.test:8000/manage/create?dwft_vue-dashboard=1
- Test loading the dashboard as an admin with the vue-dashboard flag
  - You should see both the modernized "Create a New Perma Link" section _plus_ an in-progress form beneath it.


perma_web/frontend/components/CreateLink.vue
```diff
@@ -1,5 +1,5 @@
 <script setup>
-import { ref, watch, watchEffect, computed, onBeforeUnmount } from 'vue'
+import { ref, watch, computed, onBeforeUnmount } from 'vue'
```
tinykite, line 2: watchEffect wasn't being used so I removed it as a clean-up task.

perma_web/perma/templates/user_management/create-link.html
```diff
@@ -12,101 +12,98 @@
 {% block mainContent %}
 
  <!-- vue app -->
- {% if request.user.is_staff %}
   {% flag "vue-dashboard" %}
     <div id="vue-app"></div>
-  {% endflag %}
-{% endif %}
+  {% else %}
```
rebeccacremona: If you were to outdent `{% flag "vue-dashboard" %}` and `{% else %}`, would that let you avoid reindenting most of the below markup? Asking just because I think that would make this diff a lot easier to read. 

(While acknowledging the desirability of letting `prettier` do its thing, we usually try to separate extensive no-op reformatting out into a separate PR.)

If not feasible, 👍

perma_web/frontend/components/CreateLink.vue
```diff
@@ -227,5 +227,7 @@ onBeforeUnmount(() => {
 
     <CaptureError />
 
+    <UploadForm v-if="isStaff" />
```
rebeccacremona: Love the idea of having a toggle to hide still-in-progress work. I wonder if `isStaff` is fine-grained enough: it would be nice to be able to deploy this so that admins like Kristi and Clare could continue to do their work in prod, without seeing still-in-progress components in their UI. 

Maybe let's set up a second Django Waffle flag for whether to show in-progress components? And then we could toggle that as desired, either by adding users to a group, or etc. Does that sound good to you? I could help, if it would be appreciated. Here is the PR where I made the [vue-dashboard](https://github.com/harvard-lil/perma/pull/3482/files) flag.

When installing Django Waffle, I tested accessing the flags from JS, and it worked great: https://waffle.readthedocs.io/en/stable/usage/javascript.html. IIRC, I used the [inline script](https://waffle.readthedocs.io/en/stable/usage/javascript.html#as-an-inline-script) option, and then checked the now-present global `waffle` object for flags. But, I remember that there is also a quick setup that lets you get flag info [as JSON](https://waffle.readthedocs.io/en/stable/usage/json.html), which maybe is more convenient for loading into the Vue store? No opinion from me!!

rebeccacremona (commented):
Nice! Really exciting to see this so close to live 🙌 

I made a couple suggestions; let me know what you think.

perma_web/perma/templates/user_management/create-link.html
tinykite: Sure thing, the indention has been refactored out.

perma_web/frontend/components/CreateLink.vue
tinykite: Sure thing. Following the waffle documentation, I passed the `waffle` object to the global JavaScript scope in [e9ecbdd](https://github.com/harvard-lil/perma/pull/3539/commits/e9ecbdd2c250d37f343bfc2971e6417e1a5f329a). 

I also created a new const, `showDevPlayground` to replace `isStaff` in [0f2c482](https://github.com/harvard-lil/perma/pull/3539/commits/0f2c482180d6031bb042eff3b58a813d331d93a8). Right now, the value of `showDevPlayground` is tied to the value of the `vue-dashboard` flag. But my thinking is that once a new feature flag exists, we could tie it to that. 

Do you have any preference for the name of the new flag? Definitely not at all tied to `devPlayground` if it feels another name is more succinct or more apt.

perma_web/perma/templates/user_management/create-link.html
rebeccacremona: Oh awesome, thanks.

perma_web/frontend/components/CreateLink.vue
rebeccacremona: Love it. `showDevPlayground` sounds like a great name for the flag.

tinykite: Would you prefer to include a migration in this PR, rather than merging this and then creating a separate PR for that?

rebeccacremona: Yeah, might as well keep it all in one place! Then you won't have to merge this with any placeholders.

tinykite: Okay, I think I successfully added a migration to add a new django waffle flag, and swapped it in to show/hide UI either in progress or that is otherwise developer-only. I also rebased on the develop branch, to make sure the migration gets added in the right order.

Then I tested it out in the browser, and verified I could spot the toggled flag in cookies: 
image

rebeccacremona:
LGTM! Let's do this thing!! 🎉

rebeccacremona (approved):

PR #3540, opened 2024-06-13
Produce a persistent logfile for each warc -> wacz conversion, and a task to aggregate

rebeccacremona:
See ENG-919.

### Background

Currently, the benchmarking task [creates a CSV](https://github.com/harvard-lil/perma/blob/8d70dc16678c95f1a9f586e2ab5bce3bf8a0b3c6/perma_web/tasks/dev.py#L1260C1-L1262C29) before queuing jobs, [passes that filename](https://github.com/harvard-lil/perma/blob/8d70dc16678c95f1a9f586e2ab5bce3bf8a0b3c6/perma_web/tasks/dev.py#L1281) to each celery task that converts a warc to a wacz, and then each task [opens the file and appends a line](https://github.com/harvard-lil/perma/blob/8d70dc16678c95f1a9f586e2ab5bce3bf8a0b3c6/perma_web/perma/celery_tasks.py#L1359C1-L1386C33) with info about how the conversion process went.

That limits our ability to experiment in a multi-server setup: if multiple worker minions are involved, they will necessarily write their results to different CSVs that will need to be combined, and, only the machine where the invoke task is run will get the headers.

We are also pretty sure it's fine to have lots of processes appending to the same CSV at the same time… but we're not certain.

We also think that this information would be nice to have as an artifact, during the actual conversion of the corpus, outside the context of a benchmarking experiment.

### This PR

I refactored things so that each conversion produces a JSON file with the info inside, stored next to the WACZ, and then wrote a task for aggregating that info into a CSV.

To make that more convenient: 

- I tweaked the task for enqueuing conversions, so that it optionally logs to a file, in append mode, so, if you are queuing up lots of links to experiment with, you have a record of which links' logs to aggregate afterwards.

- I extracted the logic for gathering querysets of links to look at, so that I could reuse it in the log aggregation task.

### Reviewing

The unified diff is kind of hard to read; it might be easier to just compare the files side-by-side.

### DIFF
--- a/perma_web/perma/celery_tasks.py
+++ b/perma_web/perma/celery_tasks.py
@@ -1,5 +1,9 @@
 import hashlib
+from io import StringIO
 import itertools
+import json
+import math
+import subprocess
 
 import os
 import os.path
@@ -38,9 +42,6 @@ from perma.utils import (
 import logging
 logger = logging.getLogger('celery.django')
 
-import csv
-import subprocess
-import math
 
 ### ERROR REPORTING ###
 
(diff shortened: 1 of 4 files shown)


matteocargnelutti (approved):
Nice work :) LGTM

PR #3541, opened 2024-06-13
Store warc and wacz in separate buckets

rebeccacremona:
See ENG-939.

We've been working on plans to start capturing and playing back WACZ-formatted archives, either in addition to or instead of WARC-formatted. We've also been experimenting with converting our existing archives to WACZ.

To experiment more safely, we'd like to keep WACZs in their own S3 bucket, one that we can feel comfortable messing around with, deleting artifacts from, etc... entirely separate from where we keep the WARCs for Perma's current collection, which, appropriately, has very strict rules.

Support for a secondary storage was added in https://github.com/harvard-lil/perma/pull/3520.

This PR introduces two new settings, `settings.WARC_STORAGE` and `settings.WACZ_STORAGE`.

`settings.WARC_STORAGE` is set to `default`, and every instance of `default_storage.open` or `default_storage.save` in the codebase that is operating on WARCs is changed to `storages["settings.WARC_STORAGE"].open` and `storages["settings.WARC_STORAGE"].save`. This is just an update in syntax: the two are exactly equivalent, and I just made the change for readability/clarity, as we go through this transition.

`settings.WACZ_STORAGE` is set to `secondary`, and every instance of `default_storage.open` or `default_storage.save` in the codebase that is operating on WACZs is changed to `storages["settings.WACZ_STORAGE"].open` and `storages["settings.WACZ_STORAGE"].save`. This is a change. 

This can be deployed safely, since we aren't actually doing anything with WACZ on our staging tier or in prod yet. But, before we do, we'll now have to do some configuring:

`option 1`: we could configure secondary storage to point to exactly the same bucket as primary, for convenience (which might be desirable for stage)

`option 2`: we could make a new S3 bucket, and configure secondary to point there (probably the right choice, right off the bat, for prod).

But, there's no rush on that.

### DIFF
--- a/developer.md
+++ b/developer.md
@@ -219,14 +219,14 @@ In templates, use the `{% static %}` tag and MEDIA_URL:
 Using the `{% static %}` tag instead of `{{ STATIC_URL }}` ensures that cache-busting and
 pre-compressed versions of the files will be served on production.
 
-In code, use Django's default_storage to read and write user-generated files rather than accessing the filesystem directly:
+In code, use Django's `storage` to read and write user-generated files rather than accessing the filesystem directly:
 
-    from django.core.files.storage import default_storage
+    from django.core.files.storage import storages
 
-    with default_storage.open('some/path', 'rb') as image_file:
+    with storages['default'].open('some/path', 'rb') as image_file:
         do_stuff_with_image_file(image_file)
 
-Paths for default_storage are relative to MEDIA_ROOT.
+Paths for default storage are relative to MEDIA_ROOT.
 
 Further reading:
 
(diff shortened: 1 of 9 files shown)


bensteinberg (approved):
LGTM! I gather we're still planning to use `guid_as_path()` in the object names in the WACZ storage.

rebeccacremona:
> I gather we're still planning to use guid_as_path() in the object names in the WACZ storage

For now, anyway. See ENG-760, where there's some discussion of what's at stake. I don't think we're committed to this structure; we probably do want _some_ structure; we may or may not decide that a 6-8 char prefix for every upload is still a good idea even if no longer strictly necessary. I don't myself have strong feelings and am happy follow anybody who does.

PR #3546, opened 2024-06-17
Switch to Celery-ready Scoop API; do not turn on yet.

rebeccacremona:
See ENG-917.

We're in the process of migrating the Scoop API from a homegrown task-management system to Celery. 

This PR arranges for the Celery version to work alongside Perma's dev setup.

Right now by default, Celery will not run, and captures will be processed by the original Scoop API task management code. To switch, set `START_CELERY=true` and `START_CAPTURE_PROCESSES=none` in `docker-compose.yml`.

Fun though: Perma's tests DO pass with it turned on: 
image

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -33,6 +33,13 @@ services:
   #
   # Scoop REST API
   #
+  scoop-redis:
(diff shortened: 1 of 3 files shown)


bensteinberg (approved):
Cool!

PR #3547, opened 2024-06-18
Update urllib3

bensteinberg:
Although this upgrade could have been to 1.26.19, 2.07 is what pip-compile did, suggesting that our dependencies that depend on it are happy with it.

https://github.com/urllib3/urllib3/blob/main/CHANGES.rst#207-2023-10-17

See also https://urllib3.readthedocs.io/en/latest/v2-migration-guide.html

I did not see any urllib3 deprecations when I ran the tests locally.

If the major version update seems iffy, I can redo this with 1.26.19.


rebeccacremona (approved):
Presuming the tests pass, LGTM!

PR #3548, opened 2024-06-18
Enable retrieving WACZ via DEFAULT_TO_WACZ setting

cmsetzer:
This enables WACZ retrieval when the new `DEFAULT_TO_WACZ` setting is enabled. Highlights:

* Added `DEFAULT_TO_WACZ` feature flag to settings (currently I've left it as false)
* Added `wacz_size` field to `Link` model
* When `DEFAULT_TO_WACZ = True`, the Celery task `save_scoop_capture` will retrieve a WACZ file alongside the WARC


rebeccacremona (approved):
Hooray! Love it. 

Only suggestion: returning to this, `DEFAULT_TO_WACZ`, which is what I suggested in Linear "or something", now strikes me as overbroad. What would you think of switching it to "SAVE_WACZS" or "PRESERVE_WACZS" or "STORE_WACZS" or...

(I can imagine a future where we add the equivalent for WARCs, once we are ready to stop saving those!)

cmsetzer:
Agreed about the over-broad setting name. I just pushed up [8bd269e](https://github.com/harvard-lil/perma/pull/3548/commits/8bd269e6bbeb08e65610b7602a8a072faaa8efbc) which does the following:

* Renames `DEFAULT_TO_WACZ` to `WACZ_SAVING_ENABLED`
* Adds the equivalent flag `WARC_SAVING_ENABLED` (currently set to true), which will allow us to toggle WARC storage

Let me know if that looks good...thanks!

rebeccacremona:
> Agreed about the over-broad setting name. I just pushed up [8bd269e](https://github.com/harvard-lil/perma/pull/3548/commits/8bd269e6bbeb08e65610b7602a8a072faaa8efbc) which does the following:
> 
>     * Renames `DEFAULT_TO_WACZ` to `WACZ_SAVING_ENABLED`
> 
>     * Adds the equivalent flag `WARC_SAVING_ENABLED` (currently set to true), which will allow us to toggle WARC storage
> 
> 
> Let me know if that looks good...thanks!

Perfect, I think these are great. Thanks!

PR #3551, opened 2024-06-21
Basic WACZ tests

rebeccacremona:
See ENG-967.

This PR adds a few basic tests of Perma's new WACZ-related features:
- one test for each branch of the "should we launch a WARC or a WACZ playback" code
- one functional test that actually does a WACZ playback, and confirms that it works.

To make the functional test easier to write:
- I set `WACZ_SAVING_ENABLED = True` in testing (seems harmless, right?)
- I did some light refactoring of the functional test fixtures to convert the `logged_in_user` fixture out into a utility function, for clarity and increased flexibility
- I created a new user fixture for whom the Django waffle "wacz-playback" flag is set to True.

### DIFF
--- a/perma_web/functional_tests/test_links.py
+++ b/perma_web/functional_tests/test_links.py
@@ -3,6 +3,14 @@ import re
 two_minutes = 120 * 1000
 
 def create_link(page):
+    """
+    A helper:
+    clicks into the "Create a link" input,
+    submits a link creation request,
+    and waits for the page to redirect playback.
+
+    Expect timeouts on failed attempt.
+    """
     url_field = page.locator('#rawUrl')
     url_field.focus()
     url_field.type("https://example.com/")
@@ -17,7 +25,11 @@ def create_link(page):
 
 
 def test_create_link_warc_playback(page, user, log_in_user) -> None:
-    """It should be possible to successfully create a link from a URL"""
+    """
+    It should be possible to successfully create a link from a URL.
+
+    This user (no feature flag) should see a WARC playback.
+    """
     log_in_user(page, user)
     create_link(page)
 
@@ -27,7 +39,11 @@ def test_create_link_warc_playback(page, user, log_in_user) -> None:
 
 
 def test_create_link_wacz_playback(page, wacz_user, log_in_user) -> None:
-    """It should be possible to successfully create a link from a URL"""
+    """
+    It should be possible to successfully create a link from a URL.
+
+    This user (with feature flag set) should see a WACZ playback.
+    """
     log_in_user(page, wacz_user)
     create_link(page)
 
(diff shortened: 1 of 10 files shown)


perma_web/functional_tests/test_links.py
```diff
@@ -2,34 +2,56 @@
 
 two_minutes = 120 * 1000
 
-
-def test_create_link(logged_in_user) -> None:
-    """It should be possible to successfully create a link from a URL"""
-    url_field = logged_in_user.locator('#rawUrl')
+def create_link(page):
```
cmsetzer, line 5: Is this basically a fixture/helper for the `test_create_link_XXXX_playback` tests? The roles of the test functions here read a little ambiguously to me. Maybe it should be a single parametrized test function?

Not a blocker—I think it's a matter of preference.

perma_web/functional_tests/test_links.py
```diff
@@ -2,34 +2,56 @@
...
     url_field.focus()
     url_field.type("https://example.com/")
-    logged_in_user.locator('#addlink').click()
-    logged_in_user.wait_for_url(re.compile('/[A-Za-z0-9]{4}-[A-Za-z0-9]{4}$'), timeout=two_minutes)
-    assert logged_in_user.title() == 'Perma | Example Domain'
-    assert"Example Domain" in logged_in_user \
+    page.locator('#addlink').click()
+    page.wait_for_url(re.compile('/[A-Za-z0-9]{4}-[A-Za-z0-9]{4}$'), timeout=two_minutes)
+    assert page.title() == 'Perma | Example Domain'
+    assert"Example Domain" in page \
```
cmsetzer: Just fyi, there's a missing space after `assert` (though it pre-dates this PR)

cmsetzer (approved):
Looks good; I left a couple of minor notes for you to act on or ignore as you prefer.

perma_web/functional_tests/test_links.py
rebeccacremona: Indeed there is, nice catch. Will fix.

perma_web/functional_tests/test_links.py
rebeccacremona: > Is this basically a fixture/helper

Yep, just a helper for making a Perma Link.

Hmmm, how can I make this clearer... I'm not sure a single parametrized test function is the way to go, since the assertions are opposites. I think this is awkward because it is simply weird that one user would see a WARC playback and one would see a WACZ playback, just because of who they are (rather than something in the UI), but since that's how this feature flag works for now, it's just awkward.

cmsetzer: Makes sense, thanks!

rebeccacremona: > Hmmm, how can I make this clearer... 

Gee, Cremona, how about you start by not leaving exactly the same cut-paste doc string for both tests 🙄 🤣 

I will at least update these docstrings.

rebeccacremona: Added new docstrings for all three 👍

PR #3555, opened 2024-07-02
add superset service

teovin:
To get it running:

- If docker containers are running, stop and restart them.
- Navigate to http://localhost:8088 and log in using the admin user creds specified in `docker-compose.yml`.
- Once you are in, you should be able to see the exported objects (two dashboards, two charts, two datasets, and one database connection)
- If you stop and restart your containers, and log back in, you should see the data persisting.
- If you remove any objects from your local instance, such as a chart, and restart your containers, the deleted objects should sync back in. 

Additional notes: 

- If you want to experiment with recent prod-like data, [ingest it](https://github.com/harvard-lil/perma/blob/develop/developer.md#import-production-like-data-into-the-database) into your local db before using superset service.
- Swagger docs of available endpoints can be viewed at `http://localhost:8088/swagger/v1`.
- If a dev makes a change such as updates to the dashboard, its objects, etc; the dashboard needs to be added back into the codebase using the export button. (Download the export zip, and place it inside the superset folder replacing the existing zip)

image
- To export multiple dashboards, bulk export should be used. 

image


docker-compose.yml
```diff
@@ -18,8 +18,32 @@ services:
       POSTGRES_USER: perma
     volumes:
       - postgres_data:/var/lib/postgresql/data:delegated
+    ports:
+      - "127.0.0.1:54320:5432"
```
rebeccacremona: Asking just for my own information: the database doesn't need to be exposed to the host for superset to work, right? This is just for developer convenience, if you want to use psql on your host, or similar?

rebeccacremona (approved):
Nice!! It's great it turns out setting up persistent superset dashboards is so tidy 🙂 

A couple of small ideas:

- what would you think of putting the superset config in `docker-compose.override.yml`, so that it is spun up locally, but not in CI?

- what would you think of putting superset on its own docker network and adding the db to that, instead of reusing `default`, so that the only other container superset can talk to is the perma database? we use that strategy with perma payments and the scoop api (https://github.com/harvard-lil/perma/blob/develop/docker-compose.yml#L150), so that it is clearer which containers are part of a standalone service, and where network isolation is feasible.

- what would you think of adding a paragraph to `developer.md` about working with the superset service?

rebeccacremona:
I had asked in Slack, what happens if you run `docker compose up -d` -> `docker compose stop` -> `docker compose up` again, and the entry point re-runs the initialization commands: I pulled this down to check. It looks like no big deal: I did see `perma-superset-1        | Error! User already exists admin` go by, but that doesn't seem to hurt anything. 👍

docker-compose.yml
teovin: When I set it up first, I needed to add this as I was getting errors that superset wasn't able to find the db in port 5432. Having said that, I have changed a bunch of things since initial set up, so I will test it again with that removed and I will also test the network suggestion and will report here.

teovin: It works, I pushed both updates and am looking at your other comments.

teovin:
> * what would you think of putting the superset config in `docker-compose.override.yml`, so that it is spun up locally, but not in CI?

Giving it a shot!

> * what would you think of putting superset on its own docker network and adding the db to that, instead of reusing `default`, so that the only other container superset can talk to is the perma database? we use that strategy with perma payments and the scoop api (https://github.com/harvard-lil/perma/blob/develop/docker-compose.yml#L150), so that it is clearer which containers are part of a standalone service, and where network isolation is feasible.

Completed via last commit.

> * what would you think of adding a paragraph to `developer.md` about working with the superset service?

Great idea, and will add it!

teovin:
Suggestions are complete!

teovin:
Hello, I am not sure if you were planning to look at the changes I made for the suggestions, so I will go ahead and merge this now. Feel free to reach out if you spot anything that needs my attention and I will be happy to revisit.

PR #3556, opened 2024-07-03
Add ability to export organization user list as CSV

cmsetzer:
This adds a new view, `manage_single_organization_export_user_list`, which allows authorized users (admins, registrar users, organization users) to export an organization's user list as a CSV. File and column naming conventions are patterned after Perma's existing "export links to CSV" feature wherever possible. (This view also includes undocumented support for JSON output, which can be helpful for validating/troubleshooting in the browser.)

Try out the new behavior as follows:

1. Run Perma locally from this branch
2. Log in as an admin, registrar user, or organization user
3. Navigate to `/manage/organizations` and click either of the new download buttons pointing to the export endpoint (`/manage/organization-users/2/export?format=csv`):
   image
4. Open the downloaded CSV to view the list of organization users:
   image

A test is also included; run it like so:
```sh
d pytest -k "test_org_export_user_list"
```

To discuss with Perma team: Which of the buttons should we keep, and what additional styling is needed?

### DIFF
--- a/perma_web/perma/views/user_management.py
+++ b/perma_web/perma/views/user_management.py
@@ -29,7 +29,6 @@ from django.utils import timezone
 from django.utils.decorators import method_decorator
 from django.utils.encoding import force_bytes
 from django.utils.http import urlsafe_base64_encode
-from django.utils.text import slugify
 from django.http import (
     HttpRequest,
     HttpResponse,
@@ -708,8 +707,7 @@ def manage_single_organization_export_user_list(
         .annotate(organization_name=F('organizations__name'))
         .values('email', 'first_name', 'last_name', 'organization_name')
     )
-    org_slug = slugify(org_users.first()['organization_name']) if org_users else str(org_id)
-    filename_stem = f'perma_{org_slug[:24]}_{timezone.now():%Y%m%d%HT%H%M%S}'
+    filename_stem = f'perma-organization-{org_id}-users'
 
     # Generate output records from query results and add organization name
     field_names = ['email', 'first_name', 'last_name', 'organization_name']
--- a/perma_web/perma/views/user_management.py
+++ b/perma_web/perma/views/user_management.py
@@ -1,5 +1,6 @@
 import csv
 from datetime import timedelta
+import functools
 import itertools
 import logging
 
@@ -24,6 +25,7 @@ from django.db import transaction
 from django.db.models import Count, Max, Sum
 from django.db.models.expressions import RawSQL
 from django.db.models.functions import Coalesce, Greatest
+from django.forms.models import model_to_dict
 from django.utils import timezone
 from django.utils.decorators import method_decorator
 from django.utils.encoding import force_bytes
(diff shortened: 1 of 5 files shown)


perma_web/perma/views/user_management.py
```diff
@@ -687,6 +695,40 @@ def manage_single_organization_user_reactivate(request, user_id):
...
+        .annotate(organization_name=F('organizations__name'))
+        .values('email', 'first_name', 'last_name', 'organization_name')
+    )
+    filename_stem = f'perma-organization-{org_id}-users'
+
+    # Generate output records from query results and add organization name
+    field_names = ['email', 'first_name', 'last_name', 'organization_name']
+
+    # Export records in appropriate format based on `format` URL parameter
+    export_format: Literal['csv', 'json'] = request.GET.get('format', 'csv').casefold()
```
rebeccacremona, line 716: TIL `casefold`

perma_web/perma/views/user_management.py
```diff
@@ -687,6 +695,40 @@ def manage_single_organization_user_reactivate(request, user_id):
...
+    """Return a file listing all users belonging to an organization."""
+    org_users = (
+        LinkUser.objects.filter(organizations__id=org_id)
+        .annotate(organization_name=F('organizations__name'))
+        .values('email', 'first_name', 'last_name', 'organization_name')
+    )
+    filename_stem = f'perma-organization-{org_id}-users'
+
+    # Generate output records from query results and add organization name
+    field_names = ['email', 'first_name', 'last_name', 'organization_name']
```
rebeccacremona, line 713: I'm curious about the choice to include `organization_name` in every row. Is it for ease of testing? Or, for a future where this code can be reused to export, e.g., all the users associated with a given registrar, across all their orgs? Some other reason I'm not thinking of?

rebeccacremona (approved):
This looks great! I left one question; no notes.

Re: styles/which button:

If you didn't already, I'd recommend trying at different viewport sizes and see if there is room for the floated, right-hand button at all sizes, or if things get wonky.... and if they do, that tips us in favor of the icon button, at least with the current design of this page 😅 . I might also suggest calling it "download user list" rather than "organization info".

perma_web/perma/views/user_management.py
cmsetzer: Good question. That is my attempt to address a request in [ENG-983](https://linear.app/harvard-lil/issue/ENG-983/create-a-csv-download-of-organization-users):

> If it's possible to add [the org name] as metadata or a header in the CSV itself that's ideal — otherwise just have that as an additional populated field.

My feeling is that having a normal-looking, valid CSV (albeit with redundant values) is better than sticking the org name in the header or somewhere else. As you noted, this structure would also lend itself to cross-registrar output or similar. Thoughts?

rebeccacremona: 👍 agreed, that sounds better to me too than some kind of fancier header. and, dealing with formatting names -> filenames sounds No Fun ™️, possibly less useful...

👍 harmlessly redundant data 🙂

PR #3569, opened 2024-07-18
Support more "capture" formats

rebeccacremona:
See ENG-1034.

This PR adds Perma support for additional Scoop special sauce: optional "attachments" included in the web archive https://github.com/harvard-lil/scoop?tab=readme-ov-file#main-features.

So far, we have supported the "main" high-fidelity web capture, screenshot captures, and Scoop's "provenance summary", a report on how the archive was produced and metadata about the capture process.

Going forward, we may want to support its other options. 

This PR tweaks Perma to make a record of which attachments a given archive contains.

It also toggles the dev environment's setup to include all(*) the possible attachments in the WARC/WACZ, for developer convenience.

(*) Okay, not quite all: it turns out that, with the current software, you cannot make PDF snapshots without turning off headful mode, which we do not want to do, for fidelity... TBD.

Temporarily, for admin convenience, I also added a list of which captures/attachments any given archive contains to the tray; we have not yet planned how we want to expose this in the public UI.

image


rebeccacremona:
Oh come on, finicky test 🙄 😆

matteocargnelutti (approved):
Weeeeee 🚀

PR #3570, opened 2024-07-18
CaptureJobs record which archive formats were produced.

rebeccacremona:
While preparing https://github.com/harvard-lil/perma/pull/3569, I managed to get myself thoroughly confused: at one point, I was looking at WACZs produced via our [on-demand conversion](https://github.com/harvard-lil/perma/pull/3566), but thought I was looking at WACZs produced by Scoop, and was utterly mystified by certain subtle differences.

I think you should be able to tell by looking at a Perma Link's CaptureJob, what files we produced/saved.

This PR adds a field to that effect. 

After this is deployed and the database migrations have run, populating the field for existing CaptureJobs, we can make the field non-nullable and remove the default.


cmsetzer (approved):
Looks good!

I wonder if we should deprecate `WARC_SAVING_ENABLED` and `WACZ_SAVING_ENABLED` in favor of e.g.

```py
CAPTURE_JOB_ARCHIVE_FORMATS = ['warc', 'wacz']
```

Just an idle thought; feel free to ignore or adopt as you prefer.

rebeccacremona:
@cmsetzer good idea, that is tidier!! Switching.

cmsetzer (approved):

PR #3578, opened 2024-07-24
Updated upload form handling

tinykite:
## What this does 
Adds form handling to the `UploadForm` component, so that users can either update an existing failed capture with their own static image as an archive of a page, or create a new record entirely (if no GUID ever existed) with an image. 

Satisfies Linear issue ENG-1000.

## Additional notes
- Moves user GUIDs to the global store so that it can more easily be shared within components
- Adds a `watch` function to `UploadForm` so that if a GUID changes the form will update to the appropriate full or abbreviated form fields
- Adds a testing component, `UpdateGUID` so that you can easily test out the upload functionality without a failed previous capture. 
 
## Requirements / Definition of "Done"
Using the provided developer-only testing UI, a reviewer should be able to:
- Create a new perma record that contains the user-provided image
- Update an existing perma record with an image
- The page should redirect to the new or updated Perma record when the upload process is successfully completed

Error handling and loading states will be added in future PRs. For now, testing only needs to happen for use cases where a user fills out all required fields and doesn't upload an image that is too large (over 100mb).

## Screenshots 
GUID testing tools
image

Developer testing-specific error, which surfaces a button for triggering the upload form
image

If no GUID is present in the global store, a user should be shown the full upload form 
image

If a GUID is present, a user should only be shown an abbreviated form
image

Typing in a GUID for a previous capture and hitting the "Update GUID" button should update the store, and allow the abbreviated form to be triggered
image

## How to test
- To test this PR, you'll need to locally toggle both the `vue-dashboard` and `developer-playground` waffle flags. 
  - If they haven't been already toggled, you can toggle them in the django admin or by visiting urls with the following format: `https://perma.test:8000/manage/create?dwft_FLAGNAME=1` 
- Click the "trigger error" button to surface the upload button, and trigger the upload dialog
- Try creating a brand-new perma record by filling out the required fields in the upload dialog. The page should redirect once the POST request is complete
- Try updating an existing perma record. You should be able to do this by copying the GUID from an existing perma record in the Perma link list region into the test GUID input. For example, if a perma url is `perma.test:8000/K9C8-DFGW` the guid is K9C8-DFGW

### DIFF
--- a/perma_web/frontend/components/CaptureError.vue
+++ b/perma_web/frontend/components/CaptureError.vue
@@ -4,6 +4,7 @@ import { globalStore } from '../stores/globalStore';
 import UploadForm from './UploadForm.vue';
 import { showDevPlayground } from '../lib/consts'
 
+
 const showUploadLink = ref()
 const showGeneric = ref()
 const showLoginLink = ref()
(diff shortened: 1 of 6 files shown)


perma_web/frontend/components/CreateLink.vue
```diff
@@ -158,7 +150,6 @@ const handleProgressUpdate = async () => {
     if (status === 'completed') {
         clearInterval(progressInterval)
         globalStore.updateCapture('success')
-        window.location.href = `${window.location.origin}/${userLinkGUID.value}`
```
tinykite, line 161: This has been moved to a watch function, so that it be triggered by both successful captures and successful uploads

perma_web/frontend/components/CaptureError.vue
```diff
@@ -3,6 +3,7 @@ import { ref, watch } from 'vue'
 import { globalStore } from '../stores/globalStore';
 import UploadForm from './UploadForm.vue';
 import { showDevPlayground } from '../lib/consts'
+import UpdateGUD from './UpdateGUID.vue'
```
bensteinberg:
```suggestion
import UpdateGUID from './UpdateGUID.vue'
```

perma_web/frontend/components/CaptureError.vue
```diff
@@ -49,26 +50,23 @@ defineExpose({
 
 <template>
     <div class="container cont-fixed">
+        <UpdateGUD v-if="showDevPlayground" />
```
bensteinberg:
```suggestion
        <UpdateGUID v-if="showDevPlayground" />
```

bensteinberg (changes requested):
I marked a couple of typos; I'm still trying to pull this down and run it, will review again later.

tinykite:
> I marked a couple of typos; I'm still trying to pull this down and run it, will review again later.

Great catch!

bensteinberg (approved):
Looks good, works as advertised.

(I did not know that you could set waffle flags in the URL; I just did so as a regular user in the dev environment -- very handy, but it seems like a possible vulnerability or source of problems.)

tinykite:
> very handy, but it seems like a possible vulnerability or source of problems.

That's a super valid observation — thank you for pointing that out. Yeah, in the past I've primarily worked with feature flags that were set by environment variables, and not accessible via a url. I am not particularly familiar with django waffle, but I think — if ever need be — it appears we could disable the "testing mode" url toggle functionality for a particular flag.

PR #3580, opened 2024-07-27
Fix Folder.DoesNotExist error in local deployment

cmsetzer:
https://github.com/harvard-lil/perma/pull/3563 may have introduced a regression affecting a locally deployed vanilla Perma setup (as far as I can tell, this does not affect prod). After re-deploying Perma from scratch with all Docker images, volumes, and caches cleared, the following error occurs when you submit any URL for capture:

```
  File "/perma/perma_web/api/views.py", line 512, in post
    link.move_to_folder_for_user(folder, request.user)  # also sets link.organization
  File "/perma/perma_web/perma/models.py", line 1794, in move_to_folder_for_user
    Folder.objects.select_for_update().get(pk=folder.tree_root_id)
  File "/usr/local/lib/python3.11/site-packages/django/db/models/query.py", line 637, in get
    raise self.model.DoesNotExist(
perma.models.Folder.DoesNotExist: Folder matching query does not exist.
```

If we change the `get(pk=folder.tree_root_id)` to a `filter(id=folder.tree_root_id).first()`, the query just retrieves 0 rows and then moves on without throwing an exception.

Any concerns with this solution? Or does this suggest a deeper issue?

### DIFF
--- a/perma_web/api/tests/test_link_authorization.py
+++ b/perma_web/api/tests/test_link_authorization.py
@@ -8,7 +8,6 @@ from django.utils import timezone
 from datetime import timedelta
 from django.core.files.uploadedfile import SimpleUploadedFile
 from mock import patch
-from conftest import _fix_json_fixtures
 
 
 class LinkAuthorizationMixin():
@@ -22,7 +21,6 @@ class LinkAuthorizationMixin():
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        _fix_json_fixtures()
 
 
     def setUp(self):
@@ -308,7 +306,6 @@ class LinkAuthorizationTestCase(LinkAuthorizationMixin, ApiResourceTestCase):
 class LinkAuthorizationTransactionTestCase(LinkAuthorizationMixin, ApiResourceTransactionTestCase):
 
     def setUp(self):
-        _fix_json_fixtures()
         super(LinkAuthorizationTransactionTestCase, self).setUp()
         self.post_data = {'url': self.server_url + "/test.html",
                           'title': 'This is a test page'}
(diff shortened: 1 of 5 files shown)


rebeccacremona:
Hmmmmm. Yes, I think that does suggest a deeper issue. Nice catch!!

All folders need to have `tree_root_id` populated, and if there are any folders that _don't_, those folders are in a bad state.

Here's what I'm going to bet is going on:
- since locally I've been using the same dev database for a long while, this [data migration](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/migrations/0041_folder_tree_root.py) took care of it for me
- during test set up, we use [this utility](https://github.com/harvard-lil/perma/blob/develop/perma_web/conftest.py#L79) to fix up the installation of our JSON fixtures
- during local dev setup.... we do neither: we just [install the fixtures](https://github.com/harvard-lil/perma/blob/develop/perma_web/tasks/dev.py#L91)

These old JSON fixtures are a total pain in the butt, and we've migrated... maybe half?... of the suite from using them. I can't wait until we are done with that project and can lose them.

In the meantime, I think probably the quickest fix is to run `_fix_json_fixtures` inside `invoke dev.init_db`? 

I think we don't want to tweak the code so this isn't an error; if a production folder finds itself in this state, I think it ought to be noisy about it... though, now that we have migrated the production data to use this new field, we might be able to improve the _way_ in which it is noisy about it (e.g., make the field non-nullable).

cmsetzer:
Thanks @rebeccacremona! I had a hunch it might be something along those lines.

I agree about fixing the root issue in `init_db` rather than suppressing the exception. I'll try it and push up the change once everything checks out.

rebeccacremona:
@cmsetzer wonderful, thanks!

if it isn't working for any reason... we can figure out how to update the json files so that they just have all the correct field values. i initially thought that would be too annoying, but could be that was the wrong call 🙂

cmsetzer:
I took a swing at using `_fix_json_fixtures` in the DB init task but it did not like importing from conftest.py. Rather than deal with that, I just went ahead and corrected the source data in `folders.json` so no fixer function is needed.

This takes care of the local issue on my end and tests are passing. Want to take a look?

rebeccacremona (approved):
Ughh, sorry for that hassle!! Thanks for doing this 🙏!

PR #3581, opened 2024-07-29
Fix missing step_count property

tinykite:
## What this fixes
After a user requests a new capture, there's periodic polling that happens to update the Perma UI with the status of the in-progress capture. This process can fail — for example, due to a 503 bad gateway error. 

We do not currently have error-handling in place for this situation because if one status update request fails due to an issue like connectivity reasons, the web application should currently just wait another interval to attempt another status request. 

However, a small bug that needs to be fixed is that while error handling doesn't need to happen for an error like a 503, we currently do always need to return a `step_count`, even if it's unused — because of the way props are destructured. 

For consistency, this adds fallback values to a successful and failed status request so that there should never be an undefined `step_count`. 

## Sentry reporting
Addresses PERMA-38M

## Linear issue
Satisfies LIL-2250

## How to test 
To test this, I edited `handleCaptureStatus` to intentionally throw an error on every request, and verify that if this function throws any error (no matter how generic) it shouldn't cause an issue for a user.

```
const handleCaptureStatus = async (guid) => {
    try {
        const response = await fetch(`/api/v1/user/capture_jobs/${guid}`)

        if (!response?.ok) {
            throw new Error(response.status)
        }

        // const job = await response.json()

        // return {
        //     step_count: job.step_count ?? 1,
        //     status: job.status,
        //     error: job.status === 'failed' ? job.message : ''
        // }

        throw new Error(response.status)

    } catch (error) {
        console.log(error)
        return {
            step_count: 0,
            status: 'indeterminate',
            error
        }
    }
}
```

### DIFF
--- a/perma_web/frontend/components/CreateLink.vue
+++ b/perma_web/frontend/components/CreateLink.vue
@@ -127,7 +127,7 @@ const handleCaptureStatus = async (guid) => {
         const job = await response.json()
 
         return {
-            step_count: job.step_count ?? 0,
+            step_count: job.step_count ?? 1,
             status: job.status,
             error: job.status === 'failed' ? job.message : ''
         }
@@ -137,7 +137,7 @@ const handleCaptureStatus = async (guid) => {
         console.log(error)
         return {
             step_count: 0,
-            status: 'failed',
+            status: 'indeterminate',
             error
         }
     }
--- a/perma_web/frontend/components/CreateLink.vue
+++ b/perma_web/frontend/components/CreateLink.vue
@@ -127,7 +127,7 @@ const handleCaptureStatus = async (guid) => {
         const job = await response.json()
 
         return {
-            step_count: job.step_count ?? 1,
+            step_count: job.step_count,
             status: job.status,
             error: job.status === 'failed' ? job.message : ''
         }
--- a/perma_web/frontend/components/CreateLink.vue
+++ b/perma_web/frontend/components/CreateLink.vue
@@ -127,7 +127,7 @@ const handleCaptureStatus = async (guid) => {
         const job = await response.json()
 
         return {
-            step_count: job.step_count,
+            step_count: job.step_count ?? 0,
             status: job.status,
             error: job.status === 'failed' ? job.message : ''
         }
@@ -135,7 +135,11 @@ const handleCaptureStatus = async (guid) => {
     } catch (error) {
         // TODO: Implement maxFailedAttempts logic here
         console.log(error)
-        return
+        return {
+            step_count: 0,
+            status: 'failed',
+            error
+        }
     }
 }
 
--- a/perma_web/frontend/components/CreateLink.vue
+++ b/perma_web/frontend/components/CreateLink.vue
@@ -136,7 +136,6 @@ const handleCaptureStatus = async (guid) => {
         // TODO: Implement maxFailedAttempts logic here
         console.log(error)
         return {
-            step_count: 0,
             status: 'indeterminate',
             error
         }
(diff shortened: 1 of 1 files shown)


perma_web/frontend/components/CreateLink.vue
```diff
@@ -127,15 +127,19 @@ const handleCaptureStatus = async (guid) => {
         const job = await response.json()
 
         return {
-            step_count: job.step_count,
+            step_count: job.step_count ?? 1,
```
tinykite: This might be overkill — happy to remove

rebeccacremona:
Thanks for jumping on this!

Question about how this PR works: if I understand how `step_count` is used in the UI, with this change, it would mean that, if there is a 503 when polling, then the progress bar will get reset to the beginning, to 0, no matter where it was previously. 

Is that true? If that's true, I think that might not be the desired behavior. I think it would be most natural if, should the polling fail, `step_count` remains unaltered from its previous value. How does that strike you?

perma_web/frontend/components/CreateLink.vue
rebeccacremona: Yeah, get the instinct to add a fallback, but think yeah, it might be better without the it, since having a fallback implies that sometimes the API might return a capture job without a step_count attribute, which I think is never the case.

tinykite: For sure — removed this in [43fad93](https://github.com/harvard-lil/perma/pull/3581/commits/43fad932e741df891780693ee0e2221c34235834)

tinykite:
> Question about how this PR works: if I understand how step_count is used in the UI, with this change, it would mean that, if there is a 503 when polling, then the progress bar will get reset to the beginning, to 0, no matter where it was previously.

Great question! A 503 would not update or otherwise reset the progress bar. The UI only updates when the returned status of a capture is `in_progress`, rather than the `indeterminate` status a 503 would return:

```
    if (status === 'in_progress') {
        globalStore.updateCapture('isCapturing')
        userLinkProgressBar.value = `${step_count / 5 * 100}%`
    }
```

But based on what the `handleCaptureStatus` function returns, I think this is easy to miss — because if a function returns a `step_count` value, I think it's totally reasonable to expect that it would be used. I can update this code so that it avoids the prop destructuring syntax — which would require a `step_count` to be returned even if it isn't used.

rebeccacremona:
> The UI only updates when the returned status of a capture is in_progress, rather than the indeterminate status a 503 would return

Aha, I see!! Thanks!

tinykite:
@rebeccacremona I made a tiny refactor of the code in [13f75c3](https://github.com/harvard-lil/perma/pull/3581/commits/13f75c3a5b38d95eab6d794031a12c435ef14615). 

With this change, only the `status` property is destructured (for readability within the if blocks), and there are no fallbacks for `step_count` — which should both hopefully help make the polling logic easier to intuit.

I tested that this works for the following cases: 

- A successful capture
- A failed capture (by adding `poll_data['status'] = 'failure'` to line 344 of  `celery_tasks.py`)
- A capture for which the status can't be polled, by manually throwing a 500 within the `try` block of the `handleCaptureStatus` function, (eg, `throw new Error("500")`)

rebeccacremona:
@tinykite love it!

I see that @teovin is the reviewer for this PR, but will also say that it looks good to me!

teovin:
> @tinykite love it!
> 
> I see that @teovin is the reviewer for this PR, but will also say that it looks good to me!

I trust your judgment more than I trust mine on this one, so feel free to proceed without me :)

rebeccacremona (approved):

PR #3582, opened 2024-07-30
Fix custom select keyboard navigation issue

tinykite:
## What this does 
If a user navigates to the custom select and hits the down arrow, a console error is thrown because a DOM node cannot be found and focused. 

This behavior should be fixed, and match the previous behavior of the legacy custom dropdown:

- If a user hits spacebar, clicks, or presses the down arrow key on the menu trigger button, the select options menu should become visible. 
- A user should be able to hit the down arrow to navigate to the first item in the options menu.

## Sentry error 
Addresses PERMA-38P

## Linear issue
Further detailed in LIL-2251

## Screenshot 
image

## How to test
- Make sure you've toggled the feature flag for the vue dashboard: 
```https://perma.test:8000/manage/create?dwft_vue-dashboard=1```
- Use the tab key to navigate to the custom select trigger button.
- Hit spacebar while the custom select is focused: this should trigger the options menu
- Hit the down arrow: this should navigate to the first option in the menu

Optionally:

- You can test if hitting the `end` button and the `home` button navigate to the last and first options in the menu, respectively. (They should focus an option, but not select it)
- Hitting escape should close the options menu and focus the select trigger button again. 
- You can also retest this process using the return key, rather than the spacebar, to trigger the options menu.

### DIFF
--- a/perma_web/frontend/components/FolderCustomSelect.vue
+++ b/perma_web/frontend/components/FolderCustomSelect.vue
@@ -1,5 +1,5 @@
 <script setup>
-import { computed, ref } from 'vue'
+import { computed, ref, nextTick } from 'vue'
 import { globalStore } from '../stores/globalStore'
 import { onClickOutside } from '@vueuse/core'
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: perma_web/frontend/components/FolderCustomSelect.vue [hunk 1]
```diff
@@ -50,6 +50,21 @@ const handleSelectToggle = () => {
...
+    if (!isButton) {
+        return
+    }
+
+    if (isSelectExpanded.value) {
+        return handleArrowDown(e)
+    }
+
+    handleSelectToggle()
+    await nextTick()
```
tinykite, line 65: Note: `nextTick` is a built-in Vue function that is intended to make sure the app waits until all DOM updates occur. 

In practice, this means the application should always wait until the select menu is present in the DOM tree before attempting to focus it.

bensteinberg (approved):

PR #3584, opened 2024-07-31
Update Docker Compose references to V2

cmsetzer:
Per [Docker docs](https://docs.docker.com/compose/migrate/), `docker-compose` has been deprecated since July 2023 in favor of `docker compose`. This PR updates `docker-compose` references throughout the Perma codebase, most notably in the init shell script.


bensteinberg (approved):
Thanks!

PR #3586, opened 2024-08-05
Add upload error handling

tinykite:
## What this does 
This update primarily  passes errors from Perma's backend related to archive uploading to the new `UploadForm` component. 

It also: 

- Moves where the `UploadForm` component is being rendered. Previously, it was being rendered as a child of a div with the id `error-container`. There was some CSS that applied styling at a very high specificity, using the `error-container` id, that applied a handful of undesired styles to the dialog and its children. Moving the component should fix this without requiring CSS to be written at an even higher specificity. 
- Adds a function `getGlobalErrorValues` for splitting out errors — not tied to the key for any particular form field — from the rest of the error group.
- Adds a function `getErrorFromStatus` for generating errors for failed uploads that do not return a proper response body. 
- Renames a previous function to `getErrorFromResponseOrStatus` to better differentiate that it will return an error pulled from a response body or a response status. 

## Screenshots

If a user doesn't enter a url or upload a file:
image

If the file size of an archive is larger than 100 MB: 
image

If the format of a url is invalid: 
image

_Note: Perma's backend doesn't check if a url correctly resolves — only if the format of the url itself is valid_

If there is no API response body (or it can't be properly parsed), only a status:
image

If there is no status nor valid response body: 
image  

## How to test 
- Make sure you've toggled both the `vue-dashboard` and `developer-playground` django waffle flags, either in the django dashboard or using a url query parameter. 
- You can either click "Trigger error" from the developer playground-provided UI, attempt to request a capture without providing a url, or manually make all captures fail. 
  - To make all captures fail, add this error to approximately line 344 of `celery_tasks.py`: `poll_data['status'] = 'failure'`, just after: 
  ```
        # Poll until done
        poll_network_errors = 0
        while True:
  ```
- Click the "Upload your own archive" button
- Try to submit the form without providing any information
- Try to submit the form with a file that is too large — over 100MB. There are some on the [Wikimedia page that lists media categorized as "large images"](https://commons.wikimedia.org/wiki/Category:Large_images)
- Try to submit the form with a file and an invalid format url, for example "shfhffhsagsgf" 
- Try to submit the form while manually throwing an error in `handleUploadRequest`: 
   - You can do this you comment out everything inside the `try` block and include this instead: 
   - ```        
       const error = {
            status: 500
        }
        const errorResponse = await getErrorResponse(error)

        throw errorResponse
        ```
- Try to correct your mistakes and resubmit the form. After the archive uploads, you should be redirected to the new perma capture.

### DIFF
--- a/perma_web/frontend/components/CaptureError.vue
+++ b/perma_web/frontend/components/CaptureError.vue
@@ -49,6 +49,7 @@ defineExpose({
 </script>
 
 <template>
+
     <div class="container cont-fixed">
         <UpdateGUID v-if="showDevPlayground" />
         <div v-if="globalStore.captureErrorMessage" id="error-container">
@@ -61,7 +62,6 @@ defineExpose({
                     <p>You can <button @click.prevent="handleOpen">upload your own
                             archive</button> or <a href="{{contact_url}}">contact
                             us about this error.</a></p>
-                    <UploadForm ref="uploadDialogRef" />
                 </template>
                 <p v-else>You can <button id="upload-form-button">upload your own archive</button> or <a
                         href="/contact">contact us
@@ -69,4 +69,8 @@ defineExpose({
             </template>
         </div>
     </div>
+
+    <template v-if="showDevPlayground && globalStore.captureErrorMessage">
+        <UploadForm ref="uploadDialogRef" />
+    </template>
 </template>
--- a/perma_web/frontend/components/CreateLink.vue
+++ b/perma_web/frontend/components/CreateLink.vue
@@ -9,7 +9,7 @@ import LinkCount from './LinkCount.vue';
 import FolderSelect from './FolderSelect.vue';
 import { useStorage } from '@vueuse/core'
 import CreateLinkBatch from './CreateLinkBatch.vue';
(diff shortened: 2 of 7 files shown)


teovin:
Looks good! 

One thing that maybe we can work on through a new ticket is when we use a big file in the form and click the "Create a Perma Link" button, the form stays on the screen for a few seconds before redirecting us to the new capture without any indication that we are processing the form. Maybe we can add a progress icon or display some type of text about the request being in progress as soon as the link is clicked and the form validated. Without that, as a user, my intuition is to keep clicking the button or think something is broken :D Just an idea!

teovin (approved):

tinykite:
> One thing that maybe we can work on through a new ticket is when we use a big file in the form and click the "Create a Perma Link" button, the form stays on the screen for a few seconds before redirecting us to the new capture without any indication that we are processing the form. Maybe we can add a progress icon or display some type of text about the request being in progress as soon as the link is clicked and the form validated. Without that, as a user, my intuition is to keep clicking the button or think something is broken :D Just an idea!

Great ask — yes, this will be addressed soon in another PR!

PR #3588, opened 2024-08-07
Add upload loading states

tinykite:
## What this does 
Updates form handling logic and adds in a loading indicator to the `UploadForm` component. As usual, this isn't — for now — intended to meet the demands of the accessibility needs outlined in [WCAG SC 2.2.2: Pause, Stop, Hide](https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html)

## Notable changes
- Moves computed variables used within the `CreateLink` component to `store.js` so that they can be reused across components.
- Adds utility class for min-height

## Screenshot
image

## How to test
- Make sure you've toggled the `developer-playground` and `vue-dashboard` django flags, either with a url or via the django admin 
- Trigger the upload dialog — you can do this easily by clicking "Trigger error" from the developer-only testing UI and then the clicking "upload your own archive"
- Clicking upload should momentarily flash a loading spinner. If you want to see the loading indicator for more than a single flash of a second, I'd recommend trying to upload a large file. The ["large images" category of Wikimedia commons](https://commons.wikimedia.org/wiki/Category:Large_images) provides some good options. 
  - If you've set your OS settings to reduce motion, you should seen plain, unstyled "loading" text. 
- If a url in a valid format and an image of a valid size is added, you should be redirected to the preview page for your new or updated perma link when the page is ready.


perma_web/frontend/components/UploadForm.vue
```diff
@@ -137,8 +141,13 @@ defineExpose({
             "This will create a new Perma Link." }}
             </p>
             <div class="modal-body">
+
+                <div class="u-min-h-48" v-if="isLoading">
+                    <Spinner length="2" color="#2D76EE" top="48px" />
+                </div>
```
tinykite, line 148: The loading spinner (which still relies on legacy spinner.js code) doesn't typically take up space where it is used elsewhere on the dashboard. Adding a div, with a utility class here with a min-height, felt like the easiest way to add it for this specific situation.

perma_web/frontend/components/UploadForm.vue
```diff
@@ -137,8 +142,13 @@ defineExpose({
             "This will create a new Perma Link." }}
```
teovin, line 142: Looks good! One suggestion is to lowercase the l in Link. However, I see both usages in the code base, uppercase and lowercase; so up to you!

teovin (approved):

perma_web/frontend/components/UploadForm.vue
bensteinberg: I always think "Perma Link" looks weird, but it is a term of art in Perma docs, I believe. @clare-stanton has spoken firmly to me about this before.

clare-stanton: Perma Link, please  :)

tinykite: I think "Perma Link" as a two-word proper noun has always confused my brain a bit. So I understand both of these perspectives — and definitely support the decision to keep it capitalized as a matter of consistent style. Thank you for providing these perspectives, y'all!

PR #3598, opened 2024-09-05
Dependency updates

bensteinberg:
In order to update urllib3, I had to update boto3 and botocore, as well as s3transfer:

https://urllib3.readthedocs.io/en/latest/changelog.html

https://github.com/boto/boto3/blob/develop/CHANGELOG.rst

https://github.com/boto/botocore/blob/develop/CHANGELOG.rst

Although the s3transfer README says

> If you are planning to use this code in production, make sure to lock to a minor version as interfaces may break from minor version to minor version.

the changelog looks ok:

https://github.com/boto/s3transfer/blob/develop/CHANGELOG.rst


rebeccacremona (approved):
LGTM!

PR #3602, opened 2024-09-18
add the org user expiry feature

teovin:
This PR adds the ability to associate users with organizations with an affiliation expiration date.

**Registrars and admins can create organization users with affiliation expiration dates. By default, the form will open up with an indefinite affiliation selection. (manage/organization-users/add-user?email={email})**
_Deleted organizations are filtered out from the select options._ 

image

**If the `Indefinite affiliation` checkbox is unchecked, the date selection field will display.**

image

**If the checkbox is checked again, the date selection field will hide. Form can be submitted in each state.**

**If the user already has an account, the form will only come with the organization and expiration fields, with the select options not displaying the organizations that the user is already affiliated with.**

image

**Registrars and admins can view the date on the user details page. (manage/organization-users/{id})** Registrars can only see the affiliations with organizations that are associated with them. Admins can see all of the affiliations.

image

**Users can view their affiliations on the user settings along with the expiry dates. (settings/affiliations)**

image

**Admins can view and modify the date via the admin dashboard. (admin/perma/linkuser/{id}/change/)**

image

Two scheduled tasks will run once a day checking the organization users' expiration dates: 

- The removal task will remove users whose affiliation has expired. 
- The warning task will send users an email telling them how many days they have left in their affiliation period.

**Sample notification email sent to the user:**

image

Note: This PR changes the initial behavior of the form where we were able to add multiple affiliations to the user. Since now we are adding an expiry date to each, I changed the behavior to choose one.


teovin:
I just realized I added a new section to the profile page to display the affiliations, and there is already a page for it accessible through the left-hand menu Your Affiliations item. I will make the change to update that instead.

*This is complete and I updated the pr description accordingly.

perma_web/perma/admin.py
```diff
@@ -542,10 +543,6 @@ def __init__(self, *args, **kwargs):
 
         super(LinkUserChangeForm, self).__init__(*args, **kwargs)
 
-        # make sure that user's current organizations show even if they have been deleted
-        self.initial['organizations'] = Organization.objects.all_with_deleted().filter(users__id=kwargs['instance'].pk)
-        self.fields['organizations'].queryset = Organization.objects.all_with_deleted()\
-            .filter(Q(user_deleted=False)|Q(pk__in=self.initial['organizations']))
```
rebeccacremona, line 548: I wonder if we need to re-implement something like this, for the UserOrganizationAffiliation inline. Do you know?

The way to test would be, create a new organization, add a user to that organization, then soft-delete the organization using this button on `/manage/organizations`, and see if the affiliation still shows up in the admin's inline.

image

perma_web/perma/admin.py
```diff
@@ -572,7 +574,7 @@ class LinkUserAdmin(UserAdmin):
             'fields': ('email', 'password1', 'password2'),
         }),
     )
-    raw_id_fields = ['registrar', 'organizations']
```
rebeccacremona, line 575: I wonder if we need the equivalent of this, for the `UserOrganizationAffiliationInline`. Have you run into `raw_id_fields` before? It is a performance optimization. Normally, when Django renders these on admin edit pages, it tries to give you a dropdown, from which you can select all the available options. So, in this case, it would have given you a dropdown with every single possible organization. For large tables, that turns out to be a performance problem... and can sometimes even stop the page from loading. Using this option tells Django just to print the related object's ID, instead, and gives you a number field you can change if you want.

It's probably worth checking with a production-like database, to make sure these pages performance are okay without a similar tweak to `UserOrganizationAffiliationInline`.

perma_web/perma/views/user_management.py
```diff
@@ -999,32 +999,36 @@ def edit_user_in_group(request, user_id, group_name):
         # registrar users can edit their sponsored users,
         # and users who belong to any of their registrar's organizations
         sponsorships = target_user.sponsorships.filter(registrar=request.user.registrar)
-        orgs = target_user.organizations.all() & Organization.objects.filter(registrar=request.user.registrar)
+        user_org_affiliations = UserOrganizationAffiliation.objects.filter(user=target_user).all()
+        registrar_orgs = Organization.objects.filter(registrar=request.user.registrar)
+        affiliations = user_org_affiliations.filter(organization__in=registrar_orgs)
```
rebeccacremona, line 1004: Hmmmm. If I'm reading correctly, these are not quite the same. 
`orgs = target_user.organizations.all() & Organization.objects.filter(registrar=request.user.registrar)`
 is a sum: you get all the orgs that the user is a personally a member of, and then you add to that list all the orgs associated with request.user.registrar.

I think the revised code does something different:  `user_org_affiliations = UserOrganizationAffiliation.objects.filter(user=target_user).all()` helps find all the orgs that the user is a member of, and `Organization.objects.filter(registrar=request.user.registrar)` helps find all the orgs associated with `request.user.registrar`, but `affiliations = user_org_affiliations.filter(organization__in=registrar_orgs)`, instead of producing a big list of orgs, combining the two, instead makes the queryset of `user_org_affiliations` smaller, restricting it to only include affiliations associated with that one single registrar.

I am not sure if the original logic is entirely correct: I think we _do not allow_ registrar users to also be members of any organizations, which in practice, I think means `target_user.organizations.all()` should always be an empty queryset...

But, if that is the case, I think that also means that in practice, `affiliations = user_org_affiliations.filter(organization__in=registrar_orgs)` will always be an empty queryset.

I think this is worth another look.

perma_web/perma/admin.py
teovin: Yep, I tested that flow in the admin user change form. And I did this through the init method in the UserOrganizationAffiliationAdminForm class. Here they are:
image
image

perma_web/perma/admin.py
teovin: Yep, I tested everything with a copy of prod db from yesterday and didn't run into any performance issues. I will read about raw_id_fields and see if I can incorporate it back just in case.

(Re: our convo on a related topic yesterday: Tested the rendering of manage/organization-users with prod db as well, and didn't run into any performance issues)

perma_web/perma/views/user_management.py
teovin: `orgs = target_user.organizations.all() & Organization.objects.filter(registrar=request.user.registrar)`

is making an intersection of all the organizations of the user and the organizations tied to the user (the one viewing request - registrar user in this case). So it is filtering out any organizations that this registrar isn't associated with which was what I tried to implement.

And it sounds correct to me that if a registrar is viewing a user, we want them to only see the affiliations that are under their supervision since we don't want them to have the ability to remove the affiliations that are tied to different registrars (through that registrar's organizations). But as always, correct me if my assumptions are not correct :)

perma_web/perma/admin.py
rebeccacremona: sweet!

perma_web/perma/views/user_management.py
rebeccacremona: Gah! So sorry, my brain was mxing up `target_user` and `request.user`.

The things I said make no sense. 

Carry on 😁

perma_web/perma/celery_tasks.py
```diff
@@ -1577,8 +1593,41 @@ def warn_expiring_sponsored_users(warning_days=None):
...
+                    "organization_name": affiliation.organization.name,
+                    "registrar_name": affiliation.organization.registrar.name,
+                    "registrar_email": affiliation.organization.registrar.email,
+                    "expiration_days": day,
+                    "expiration_date": expiration_date + timedelta(days=1)
+                }
+                try:
+                    email_expiring_user(affiliation.user, "email/organization_user_expiry_notification.txt", context)
+                except Exception as e:
+                    logger.error(f"Error emailing user {affiliation.user_id}: {e}")
```
cmsetzer: We might consider using `logger.exception` so the full traceback gets included:
```suggestion
                except Exception:
                    logger.exception(f"Error emailing user {affiliation.user_id}")
```
The error might be obvious in most cases, though — I'll leave this up to you.

perma_web/perma/celery_tasks.py
```diff
@@ -1546,6 +1546,22 @@ def deactivate_expired_sponsored_users(expiration_date=None):
...
+def remove_expired_organization_user_affiliations(expiration_date=None):
+    """
+    Removes affiliation of organization users whose affiliation expired
+    """
+    if not expiration_date:
+        expiration_date = timezone.now().date()
+
+    deleted_affiliations = UserOrganizationAffiliation.objects.filter(expires_at__lt=expiration_date).delete()
+
+    if deleted_affiliations[0] > 0:
```
cmsetzer, line 1559: Should we add another condition here to avoid erroring out when `deleted_affiliations` is empty? (Or maybe I'm misunderstanding the object returned by `delete()`?)
```suggestion
    if deleted_affiliations and deleted_affiliations[0] > 0:
```

cmsetzer (approved):
Nice work on this! (And thanks for your patience with this review.)

Everything looks good for the most part. One point of UX feedback: the checkbox label "Indefinite affiliation" feels a bit engineer-ish to me. If I were an organization/registrar admin, "Permanent affiliation" or something like that might feel more intuitive. (Clare may have already weighed in on this, in which case feel free to ignore me.)

perma_web/perma/celery_tasks.py
teovin: Great suggestion on the label naming. I made the change for this and also for a similar label I had created for the registrar sponsorship form. I also made the change to both to use logger.exception().

When it comes to deleting affiliations, if no db objects are impacted, meaning nothing is deleted, then `deleted_affiliations[0]` will be `0`. delete() returns a tuple with its first item being the number of objects deleted.

PR #3604, opened 2024-09-26
Fix bookmarklet

jcushman:
We should include ?url= in the integration tests, but I didn't yet. :) LMK if you want to do that now or hold off since tests are being refactored.


jcushman:
Added a test.

rebeccacremona (approved):
Nice!

PR #3605, opened 2024-10-01
Refactor test_views_common.py

rebeccacremona:
I set out to convert `test_views_common.py` to pytest format, and to remove its dependency on our legacy JSON test fixtures, swapping in our pytest test fixtures instead.

I did that, but in the process, I decided to split both `views/common.py` and `test_views_common.py` into a few separate modules: one for memento-related views, one for playback, one for views related to our contact and reporting forms... and then whatever was left, still in "common". 

I did this because "common" has never felt descriptive to me, the file was huge, and it was hard to find things in it. I think this is a lot more manageable, and I have wanted to do it for ages.

I made no non-trivial changes to the tests themselves, even if I thought the tests could use improvement, to reduce the scope of this PR. I made _zero_ changes to any of the view code. There is no way you will be able to tell that from the diff 🤣.

As far as reviewing goes... I feel like the only thing worth looking at is, whether you see anything that looks obviously bad/weird in `conftest.py`. I think other than that, it is not worth anybody's time to review: let it be on my head if I messed these tests up during the migration 😂.


perma_web/conftest.py
```diff
@@ -402,26 +410,46 @@ class InProgressCaptureJobFactory(PendingCaptureJobFactory):
...
+    status = 'success'
+    record_type = 'response'
+    content_type = 'text/html'
+
+
+@register_factory
+class ScreenshotCaptureFactory(CaptureFactory):
+    role = 'screenshot'
+    status = 'success'
+    record_type = 'respsonse'
```
bensteinberg:
```suggestion
    record_type = 'response'
```

perma_web/conftest.py
```diff
@@ -662,7 +702,119 @@ def f(customer):
...
+            cached_can_play_back=True,
+        ) for time in [today, a_few_days_ago, within_the_last_year, over_a_year_ago, three_years_ago]
+    ]
+
+    return {
+        "domain": domain,
+        "url": url,
+        "links": [
+            {
+                "guid":link.guid,
```
bensteinberg:
```suggestion
                "guid": link.guid,
```

perma_web/conftest.py
```diff
@@ -662,7 +702,119 @@ def f(customer):
...
+                #
+                "timestamp":  link.creation_timestamp
```
bensteinberg:
```suggestion
                "timestamp": link.creation_timestamp
```

perma_web/conftest.py
```diff
@@ -662,7 +702,119 @@ def f(customer):
...
+
+### For testing email ###
+
+@pytest.fixture
+def email_details():
+    return {
+        "from_email": 'example@example.com',
+        "custom_subject": 'Just some subject here',
+        "message_text": 'Just some message here.',
+        "refering_page": 'http://elsewhere.com',
```
bensteinberg: I wish the key here were `referring_page`, but that would require changes elsewhere; maybe I'll make a PR.

bensteinberg (changes requested):
I've made a few suggestions.

perma_web/conftest.py
rebeccacremona: Thanks for these!! 

I can make that change, hold one...

rebeccacremona: Done! I considered intentionally making a typo in the commit message, like `refferring` 😁

bensteinberg (approved):
❤️

PR #3608, opened 2024-10-02
Convert bottom half of dashboard to Vue

jcushman:
Here's a first cut at converting the rest of the Perma dashboard to Vue:

* The `vue-dashboard` flag now enables the new code -- the Vue stuff already in production moves out of the flag. So, to test this, set vue-dashboard = true. I put a gray box behind the new stuff so it's easier to visually distinguish for testing.
* There are a few changes to the existing Vue code, like moving the state object to Pinia. So, if we wanted to deploy this to prod without enabling the new stuff, we would use vue-dashboard = false and have to kick the tires on that. For now let's not.
* At this point I could use help from people kicking the tires and finding a punchlist of things that don't work. Just a quick "when I do x, it does y on prod but z in this branch".

Known limitations (let's list more):
* When dragging links, the little "[link]" thing is super far from the mouse. I messed with this and didn't find a fix.
* The "Loading links ..." thing flashes in and out really fast.


rebeccacremona:
I will QA with `vue-dashboard = false` locally.

PR #3609, opened 2024-10-03
fix org user expiry bug

teovin:
To reproduce, navigate to `/manage/organization-users/<user_id>`, click `remove`. The page errors with 404. 

image

The fix is to grab the affiliation id coming from the post request of the view and delete the affiliation. Previously the affiliation id was being used to query the organizations. I also renamed the `org` field in the template to represent the value it carries more accurately.


rebeccacremona (approved):
Hooray for a successful bug hunt!!

PR #3610, opened 2024-10-04
Separate sign-up views into their own module

cmsetzer:
Following on @rebeccacremona's work in https://github.com/harvard-lil/perma/pull/3607, this breaks the sign-up views off of `user_management.py` and puts them into their own module, `user_sign_up.py`.

There are some functions such as `approve_pending_registrar` that fall somewhere in between "user management" and "sign-up" in terms of functionality; in such cases, I did my best to minimize import dependencies between modules and keep functions alongside their associated views. Open to feedback though.


rebeccacremona (approved):
Yes! Into it!

PR #3611, opened 2024-10-04
Vue fixes

jcushman:
Fixing stuff found in QA (thanks, Becky!). I think this makes the `developer-playground = false` stuff ready to QA and ship, before I work on the `developer-playground = on` stuff, and also fixes all the already-on-prod issues.

>In your accessibility settings, turn on "reduce motion". Make a Perma Link in production. Observe that you see the word "loading", instead of a spinner. (TBD, if that is the right word. I think it probably isn't.) With this branch, the spinner is displayed regardless.

This turned out not to reproduce. But, I didn't like the "Loading" text, so I switched `<Spinner>` to show a paused spinner instead with reduced motion, so it's more consistent to use in a design.

> In prod, click "make multiple links" to open the link batch modal. Observe that the Vue version of the modal opens. With this branch, observe that the old modal opens.

Regrettably this is because I messed up the feature flags and the top half was rendering the old code rather than new code; apologies for wasting QA time! Feature flags are now fixed 🤞 so that:
* with both flags off, top half of dashboard is Vue, bottom half is old code
* `vue-dashboard` does nothing
* `developer-playground` switches the bottom half to Vue

> Submit an invalid URL. Observe that you are offered the opportunity to upload your own, in the error message. You should not be, I think? (Luckily, if you DO attempt to upload your own, with that same invalid URL, it fails validation.)

Removed the upload option for "Not a valid URL" error.

> Make a Perma Link. Observe that the spinner appears while your capture job is pending. Observe that the progress bar appears when capture begins. Observe that the spinner remains. That is a bug: the spinner should be replaced by the progress bar.

I tried this but my first attempt looked weird. Let's keep both for now. I think it'll help that I switched the reduced-motion to be a paused spinner.

> Make a Perma Link. After you are redirected to the new Link, use your back button to return to the dashboard. The UI will report that your link is still in progress.

We now call resetForm on success, which resets the button, error message, and URL contents. Note the URL box doesn't have focus on back button, not sure if it should.

> Make a link batch. Observe that the styling is broken for a second, right when you submit the request. There should be a spinner for a second. Instead, regardless of whether you have "reduce motion" on or off, you see "loading", with incorrect padding.

I moved the spinner to the content area of the modal and added a bunch of padding; seems better.

> Use the link batch history to open the modal for a completed batch, or make a new batch. Observe that the Vue version of the modal is weirdly skinny, and that text wraps, causing irregular spacing.

Added the same `modal-lg` class as the old modal.

Also in this commit:
* fix a typo in an error handler (`getErrorFromResponseStatus` -> `getErrorFromResponseOrStatus`)
* move `captureStatus`, `isLoading`, and `isReady` out of globalStore since they're only used in CreateLink; other variables may need this treatment too.


perma_web/frontend/components/CaptureError.vue
```diff
@@ -58,7 +62,7 @@ defineExpose({
                 </span></p>
             <p v-if="showGeneric" class="message">We’re unable to create your Perma Link.</p>
             <template v-if="showUploadLink">
-                <template v-if="showDevPlayground">
+                <template v-if="false && showDevPlayground">
```
rebeccacremona: Is this `false` here on purpose?

perma_web/frontend/components/CaptureError.vue
```diff
@@ -70,7 +74,7 @@ defineExpose({
         </div>
     </div>
 
-    <template v-if="showDevPlayground && globalStore.captureErrorMessage">
+    <template v-if="false && showDevPlayground && globalStore.captureErrorMessage">
```
rebeccacremona: Same here.

perma_web/frontend/lib/helpers.js
```diff
@@ -6,4 +6,8 @@ export function getCookie(name) {
       "(?:^|; )" + name.replace(/([\.$?*|{}\(\)\[\]\\\/\+^])/g, '\\$1') + "=([^;]*)"
   ));
   return matches ? decodeURIComponent(matches[1]) : undefined;
-}
\ No newline at end of file
+}
+
+export function prefersReducedMotion() {
+  return window.matchMedia('(prefers-reduced-motion: reduce)').matches;
+}
```
rebeccacremona, line 13: Can you say a word about how this differs from the `import { usePreferredReducedMotion } from '@vueuse/core'` strategy (other than being more readable 😜). Is it the same thing, under the hood?

rebeccacremona (approved):
Thanks for all these fixes! I did not pull this down and try it locally, but LGTM! One question about some possibly stray `false`s; one question for curiosity's sake.

perma_web/frontend/lib/helpers.js
jcushman: It's the same media query -- I just found `usePreferredReducedMotion().value === 'reduced'` clunky enough that I wanted a helper for it, and the helper simple enough not to require a library.

usePreferredReducedMotion() does [a bunch of work](https://github.com/vueuse/vueuse/blob/main/packages/core/usePreferredReducedMotion/index.ts#L13) to be reactive, so if we wanted to watch and respond to changes in reduced motion it would be helpful.

perma_web/frontend/components/CaptureError.vue
jcushman: On purpose but I think the wrong decision -- my thinking was since the upload form isn't done I should disable it, but if you're QAing with developer-playground = true you'll want to see the new one, done or not. Pushed a fix.

PR #3613, opened 2024-10-04
Convert settings views' tests to pytest syntax

rebeccacremona:
This PR converts all the tests for the settings views, extracted in https://github.com/harvard-lil/perma/pull/3607, to pytest syntax, and switches them to use pytest fixtures instead of JSON fixtures + mocks.

I don't love some of these tests, but I resisted the urge to change them too much: not the goal of this exercise 🤣.

As before, I think only a lightweight review is in order, and let it be on my head if I messed things up in the conversion: not worth anybody's time to study this. 

But, a quick buzz through for code style would be appreciated. Do you see anything that offends sensibilities?


perma_web/conftest.py
```diff
@@ -270,11 +270,7 @@ class PayingRegistrarFactory(RegistrarFactory):
     cached_paid_through = GENESIS
     base_rate = Decimal(100.00)
     in_trial = False
-
-
-@register_factory
-class PayingLimitedRegistrarFactory(PayingRegistrarFactory):
-    unlimited = False
+    unlimited=True
```
cmsetzer: Totally minor linting point but maybe 
```suggestion
    unlimited = True
```
for consistency?

perma_web/perma/tests/test_views_user_settings.py
```diff
@@ -0,0 +1,648 @@
...
+    orgs = soup.select('.settings-block p')
+    assert len(orgs) == 2
+    for org in orgs:
+        assert org.text.strip() in org_names
+
+
+def test_affiliations_registrar_user(client, registrar_user):
+    _, soup = get_afffilation_markup_for_user(client, registrar_user)
+
+    [registrar] = soup.select('h4')
```
cmsetzer: TIL about this list unpacking syntax. Cool! I do feel like it would read more clearly as
```suggestion
    registrar = soup.select_one('h4')
```
but it's a matter of preference — will leave it up to you.

perma_web/perma/tests/test_views_user_settings.py
```diff
@@ -0,0 +1,648 @@
...
+        assert setting.text.strip() in ["Website", "Email"]
+
+
+def test_affiliations_pending_registrar_user(client, pending_registrar_user):
+    response, soup = get_afffilation_markup_for_user(client, pending_registrar_user)
+
+    assert b'Pending Registrar'in response.content
+    assert b'Thank you for requesting an account for your library. Perma.cc will review your request as soon as possible.' in response.content
+
+    [registrar] = soup.select('.sponsor-name')
```
cmsetzer: See previous comment. Optional suggestion:
```suggestion
    registrar = soup.select_one('.sponsor-name')
```

cmsetzer (approved):
Looks great! Thanks so much for this, I'm sure it was a slog but this will set the table for lots of further test improvements.

I included a few optional stylistic suggestions; feel free to adopt or ignore them at your discretion.

perma_web/perma/tests/test_views_user_settings.py
rebeccacremona: Ah, good to know about `select_one`!

I agree that it is clearer to read. It is also slightly different: `[registrar] = soup.select('h4')` will cause the test to fail if more than one `h4` is present; `registrar = soup.select_one('h4')` will happily ignore everything but the first instance. I think in these tests... that difference is fine. But I will see if I still think so on Monday morning 😀 . And if yes, will commit both suggestions.

Thanks for the review!!

PR #3615, opened 2024-10-07
Prevent sass deprecation warnings

jcushman:
As a bit of housekeeping, this prevents deprecation notices during sass compile from [using slash as a division operator](https://sass-lang.com/documentation/breaking-changes/slash-div/) and using [minus-sign-variable without parens](https://sass-lang.com/documentation/breaking-changes/strict-unary/).

As suggested by sass, I did

> npm install -g sass-migrator
> sass-migrator division static/css/\*.scss
> sass-migrator strict-unary static/css/\*.scss

and then went and cleaned up the changes.

Fun fact for division: these are mostly in an early section of the css that must have an interesting story behind it. There's a bunch of stuff like `.col-sm-10by3 { width: 100% * math.div(5, 18); // 10/3 width }` that looks like it was copied from bootstrap, modified, and in some cases not currently used. I didn't go down the rabbit hole of figuring out what should be imported from bootstrap or deleted.

Fun fact for unary negation: I believe the rules `strict-unary` updated were in fact _not_ operating as intended, which of course is why they started printing warnings about them. There's stuff in here like `margin: 0 -$grid $double -$grid;` which apparently would have been parsed as `margin: (0-$grid) ($double-$grid)`, which I'm sure can't be what the author meant. I arbitrarily chose to have these do what they read like rather than how browsers have been interpreting them, by updating to `margin: 0 (-$grid) $double (-$grid);`, so at least the authorial intent will be visible if anyone ever goes back to revise. I looked at the affected elements (the single-link menu dropdown, and the link browser item-containers) and they look fine to me.


rebeccacremona (approved):
😁 This is such a good one.... we'll never know.  Thanks for doing this!

PR #3616, opened 2024-10-09
Vue css

jcushman:
Fixes to most recent QA 🙏

> Load a folder with lots of links. Look at the bottom. How it should look: same height as the folder tree, then a subtle drop shadow, links list scrolls
> The drop shadow should not be present when the list of links is too short to scroll.

I applied a modern flexbox layout to the folder tree + link list area, so the whole thing is 90vh and the folder tree, batch history and link list all scroll. I removed the slick custom scrolling drop shadow effect 😢 in favor of a constant drop shadow at the bottom of the box; I think it's more clear that way.

NOTE we decided the single-column layout for narrow windows is out of scope for now; it's not great.

> Expand a link in the link list. The "Default View" label is much larger now.

Fixed.

> The  vertical spacing of the "select a folder" dropdown looks a little wonky . Before and after.

The odd look is caused by using a unicode└ to indicate subfolders. We decided to keep it.

> I am pretty sure that "or create multiple links" used to disappear, when you started making a link.

Fixed.

> I checked private links and private orgs too, now. I think it is mostly correct! Only thing that is weird: the "Create Private Perma Link" button turns blue on hover/focus. I don't think that's correct.

Button is now darker orange and turns lighter orange on hover/focus.

Also in this PR:

* In working on the button, I discovered that we hadn't brought over the old folder-select code to handle private folders flags. Added.
* In adding flexbox I discovered that the batch history was still using a bootstrap "dropdown" to hide/show. Switched to using a reactive Vue boolean.


perma_web/frontend/components/FolderTree.vue
```diff
@@ -23,31 +23,37 @@ const onNodeSelect = (node) => {
...
+    isReadOnly,
+    isOutOfLinks,
     path: jstreeRef.value.getFolderTree().get_path(node) || [],
-    folderId: node.data.folder_id || '',
-    orgId: node.data.organization_id || '',
-    sponsorId: node.data.sponsor_id || '',
-    isPrivate: node.data.is_private || false,
-    isReadOnly: node.data.read_only || false,
-    isOutOfLinks: node.data.out_of_links || false,
+    isPrivate: org?.default_to_private || false,
```
jcushman, line 41: This change adapts these two functions that I hadn't brought over properly:

```
function sendSelectionChangeEvent(node) {
  var data = {};
  if (node.data) {
    data.folderId = node.data.folder_id;
    data.orgId = node.data.organization_id;
    data.sponsorId = node.data.sponsor_id;
    data.readOnly = node.data.read_only;
    data.path = folderTree.get_path(node);
  }
 
function handleSelectionChange(data) {
  var currentOrg = data.orgId;
  var currentSponsor = data.sponsorId;
  var readOnly = data.readOnly;
  var path = data.path;
  var outOfLinks = !readOnly && !currentSponsor && !currentOrg && !link_creation_allowed;

  // update top-level variables
  currentFolder = data.folderId;
  currentFolderPrivate = organizations[currentOrg] && organizations[currentOrg]['default_to_private'];
  var updateFolderSelection = new CustomEvent("vueDispatch", {
    bubbles: true,
    detail: {
      name: 'updateFolderSelection',
      data: {
        path: path,
        orgId: data.orgId,
        folderId: currentFolder,
        sponsorId: currentSponsor,
        isPrivate: currentFolderPrivate,
        isReadOnly: readOnly,
        isOutOfLinks: outOfLinks
      }
    }
  });
```

rebeccacremona (approved):
Huzzah!

PR #3618, opened 2024-10-11
Capture error handling

jcushman:
Some improvements to error handling when capturing URLs, both single link and batch.

Single link errors are shown closer to the form, smaller, read by screen readers, and tweaked handling for a couple of messages:

image

I also looked at error handling for the batch capture modal and found that its error handling was pretty clunky -- if you don't put in urls, it closes the modal and shows a toast, instead of showing an inline error to let you fix the problem. Pulling this thread, I ended up:

* refactoring FolderSelect to use a consistent dropdown inside and outside the batch capture modal
* adding inline error handling to capturebatch

image

Accessibility notes:

This is forward progress on accessibility, as the single-link errors now get read by VoiceOver. Neither the toasts or inline form errors are being read by VoiceOver for me, however. I think we should address that as a separate bug.


bensteinberg (approved):
LGTM! I'll get this on stage, where other people can have a look, as discussed.

PR #3619, opened 2024-10-11
Post-Vue-dashboard file moves

jcushman:
This PR makes a bunch of cleanup moves now that we're moving the Vue dashboard out of a developer flag. These tend to step on file history, so I'm doing them in one big batch in the hopes of getting that out of the way:

* Move everything out of the developer-playground flag. We're keeping the flag itself for potential use later.
* Re-indent the new Vue files in frontend/ with two spaces, which is what we generally use for HTML/CSS/JS. (It would make sense at some point to adopt ESLint and Prettier, but I just did a quick effort with Pycharm for now since the files are moving anyway, in the hopes of reducing the diff size for that effort later.)
* Delete all of the now-unused javascript from the old create page.
* Move the frontend/ folder to static/frontend/. In working with Vite I found that it was hard to get both developer and production builds working with the right paths when we had Vue and CSS in different static folders.


bensteinberg (approved):
LGTM! I'll merge and deploy to stage. I think this will now require pretty thorough QA on stage before deploying to prod; if we need to do a bugfix before that QA happens, we might have to revert this PR and maybe the previous one, #3618, before applying the fix.

PR #3625, opened 2024-10-18
Tweak error messages and upload your own behavior.

rebeccacremona:
In this PR, I'm doing three uncontroversial things, and one potentially controversial one.

### Uncontroversial

1) While doing QA of the new Vue dashboard, I noticed that some of our error messages are unnecessarily terse or are otherwise a little clumsy. I tweaked them to suit my personal sensibilities; please feel free to suggest alternatives!

2) I noticed that the upload-your-own form wasn't handling titles and descriptions correctly: they should always be optional, and you should get the chance to supply them both when uploading fresh, or when replacing a failed capture.

image

3) I realized that I was wrong last week, when I said that the option to upload your own should not appear when you enter an invalid URL. That is in fact not correct: many valid URLs, like `https://afdc.energy.gov/stations#/analyze?country=US&region=US-VA&fuel=ELEC`, currently fail validation, and our users rely on this workaround. See LIL-836.

### Potentially controversial

[This PR](https://github.com/harvard-lil/perma/pull/3618) significantly improved error handling.

But, in the process, it switched to unstyled messages... which I personally am finding difficult to parse, because the messages are long.

Here, I tried reintroducing a little styling.

I can easily revert it, if we prefer the unstyled, one-line version.

Examples:

image
image
image
image


rebeccacremona:
(I also think it is a goofy we have two versions, a bunch that start with "We're unable to create your Perma Link" and a bunch that start with "Perma can't create this link." I am tempted to fix that up too... but am not sure.)

jcushman:
Looks great! Let's do it.

bensteinberg (approved):
LGTM. I notice that we vary in using `’` and `'` for apostrophes, but maybe we can fix that systematically another time.

rebeccacremona:
> LGTM. I notice that we vary in using `’` and `'` for apostrophes, but maybe we can fix that systematically another time.

Yeah I am the worst about that 😅 . I use straight quotes everywhere.

PR #3629, opened 2024-10-22
Don't post a message when a link is deleted.

rebeccacremona:
See LIL-2768.

You are allowed to (soft) delete a Perma Link for 24hrs after its creation. When you do, we return you to the dashboard. We used to use [Django's message framework](https://docs.djangoproject.com/en/5.1/ref/contrib/messages/) to post a little notification, communicating that your link had been successfully deleted. But, now that this page is rendered by Vue, that message isn't getting displayed. Instead, Django holds on to the message and displays it... whenever it next gets a chance to display a message. For instance, on the login page 😂 .

We could change things so that you get a javascript-mediated message... but I honestly don't think it's necessary.

This PR just removes the code for posting a message.


rebeccacremona:
I am looking forward to seeing if this breaks a test. I hope it does :-)

bensteinberg (approved):

PR #3630, opened 2024-10-22
Infinite scroll fix

jcushman:
In the link list, limit and offset weren't properly reset when calling fetchLinks with append=false.

* Call reset_pagination() when the folder reloads
* Avoid jumping to top during scrolling by changing how the spinner renders
* Also a couple of tweaks to prevent hypothetical race conditions if you loaded a bunch of folders and did a bunch of scrolling before earlier fetches return. This wouldn't fix every possible out-of-order response, but would fix the obvious ones where you change folders during a load.


perma_web/webpack-stats.json
```diff
@@ -217,5 +217,7 @@
       "single-link-permissions.js",
       "single-link-permissions.js.map"
     ]
-  }
+  },
+  "error": "ModuleBuildError",
```
rebeccacremona: Ah! I think if you recompile fresh, this error message will go away.

rebeccacremona (changes requested):
Thanks! Looks like just webpack-stats.json needs to be recompiled.

Issue #1749, opened 2016-10-13
court, journal, faculty signup forms slightly weird for logged in users


rebeccacremona:
The library signup form behaves very differently than the court, journal, and faculty signup forms, when visited by a logged in user or a logged out but already-signed-up user. 

The court, journal, and faculty forms are exactly the same, regardless of login status. You may enter an email address even while you are logged in. With the court form, you may indicate you want a new individual account created or not; with the journal and faculty forms, it is assumed you want a new account. The court form uses special custom logic to see if the email address you supply is already associated with an account, and behaves differently depending; the journal and faculty forms simply reject submissions from existing users ('A user with that email address already exists.')

The library form only asks people for an email address if they are logged out. Otherwise, it ties the registrar request to the logged in user.

Maybe these are all different for a reason. Probably they should be made more consistent.

cmsetzer:
It looks like this got closed by accident; we noticed that the inconsistency and unexpected UX described above still exists.

PR #3631, opened 2024-10-22
Fix sign-up form inconsistency for logged-in users

cmsetzer:
Per https://github.com/harvard-lil/perma/issues/1749 (from back in 2016), Perma's sign-up forms do not behave in a predictable or consistent way for logged-in users. This attempts to fix the form weirdness, which will in turn unblock the Other Org form workflow improvements I've got on another branch.

Before:
* [Libraries](https://perma.cc/libraries) form does not display user fields (first name, last name, email) when user is logged-in
* [Courts](https://perma.cc/sign-up/courts) and [Other Org](https://perma.cc/sign-up/firms) forms do display user fields even when user is logged-in, leading to confusing UX

After:
* For all three forms, the behavior is the same: display user fields for logged-out user, do not display them for logged-in user
* I also tweaked a few view names/paths for better consistency, namely:
  + `libraries` view is now `sign_up_libraries`, to match the other views
  + `/libraries` URL now redirects to `/sign-up/libraries`, to match the other URLs
  + `sign_up_firm` view is now `sign_up_firms`, to match the other views

### DIFF
--- a/perma_web/perma/forms.py
+++ b/perma_web/perma/forms.py
@@ -195,8 +195,8 @@ class UserForm(forms.ModelForm):
         """
         return (
             hasattr(self, 'request')
-            and hasattr(self.request, 'user')
-            and isinstance(self.request.user, LinkUser)
+            and hasattr(self.request, 'user')  # noqa: W503
+            and isinstance(self.request.user, LinkUser)  # noqa: W503
         )
 
 
--- a/perma_web/perma/forms.py
+++ b/perma_web/perma/forms.py
@@ -303,10 +303,18 @@ class CreateUserFormWithFirm(UserForm):
 
     def __init__(self, *args, **kwargs):
         super(CreateUserFormWithFirm, self).__init__(*args, **kwargs)
-        self.fields['first_name'].label = "Your first name"
-        self.fields['last_name'].label = "Your last name"
-        self.fields['email'].label = "Your email"
+
+        self.fields['first_name'].label = 'Your first name'
+        self.fields['last_name'].label = 'Your last name'
+        self.fields['email'].label = 'Your email'
         self.fields['would_be_org_admin'].label = 'Would you be an administrator on this account?'
+
+        # Populate from logged-in user
+        if hasattr(self, 'request') and hasattr(self.request, 'user'):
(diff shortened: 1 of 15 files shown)


perma_web/perma/forms.py
```diff
@@ -187,6 +187,19 @@ def make_random_password(len=20):
...
+        """Determine whether there is a currently logged-in user.
+
+        This may be useful for determining, e.g., whether or not to
+        display certain user registration fields.
+        """
+        return (
+            hasattr(self, 'request')
+            and hasattr(self.request, 'user')
+            and isinstance(self.request.user, LinkUser)
+        )
```
rebeccacremona: Nice! 

Elsewhere in the application where we do something similar, we check `request.user.is_anonymous` or `request.user.is_authenticated` rather than `isinstance(self.request.user, LinkUser)`. Do you think there is any reason to use one of those here instead of the type check?

perma_web/perma/views/user_sign_up.py
```diff
@@ -120,8 +120,10 @@ def sign_up_courts(request):
     Register a new court user
     """
     if request.method == 'POST':
-
-        if something_took_the_bait := check_honeypot(request, 'register_email_instructions', check_js=True):
```
rebeccacremona, line 124: LOL. Yes, perhaps this use of `:=` was a bit much 😉

rebeccacremona (approved):
Hooray!! Unbelievable that this is finally going to make sense, after all this time 🙌 

Just one stylistic question.

LG(reat)TM!

perma_web/perma/forms.py
cmsetzer: Good catch! This was an oversight on my part. I just pushed up [48ef889](https://github.com/harvard-lil/perma/pull/3631/commits/48ef889e8682461898db5e6d80dcd8b92642476d), which tweaks `UserForm.user_is_logged_in` to use `is_authenticated`.

(It also now requires the `request` to be passed in explicitly. Otherwise, there might someday be a confusing condition where `user_is_logged_in()` returns false not because the user is not logged in, but because we forgot to set `request` as an attribute of the form.)

PR #3634, opened 2024-10-24
Remove count from pagination

bensteinberg:
As discussed in Slack, this speeds up slow folder loads in the API when a folder has many links, at the cost of changing the conventional behavior of the API. Previously, a client consuming paginated results could stop when the `next` link was `null`; now, such a client should keep consuming pages and following `next` links until `objects` is an empty list.

### DIFF
--- a/perma_web/api/utils.py
+++ b/perma_web/api/utils.py
@@ -25,7 +25,7 @@ logger = logging.getLogger(__name__)
 
 class TastypiePagination(LimitOffsetPagination):
     """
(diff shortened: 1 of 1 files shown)


rebeccacremona (approved):

bensteinberg:
I'm taking a look at https://github.com/harvard-lil/perma-extension to see whether this change will require a change to the extension.

bensteinberg:
> I'm taking a look at https://github.com/harvard-lil/perma-extension to see whether this change will require a change to the extension.

(I don't see any use there of total count or the next link, but I'd love a second opinion.)

PR #3636, opened 2024-10-25
Show better errors in bulk create and upload forms

jcushman:
I was going to do a quick fix for this sentry error, but got more pulled into it than I meant to ...

image

Dealing with a failed fetch is tricky, because you may or may not have a response object, if the connection is broken, and may or may not have a message in the json payload, and if you do have one there's different ways they get formatted. In the hopes of not having different error if-thens in different components, I put in a new function that is pretty permissive about trying to process the response and show whatever info is available, so if a form fails someone can at least send us a useful error message. To avoid going too far down the rabbit hole, I used the new function just for bulk and upload but not for the main capture page, so I didn't end up fully replacing the other error handling logic.

Here's some examples of it working, created by modifying the fetch to use an invalid url or to not include all the form data. The idea is these may not be the prettiest, but at least the JS error handling code doesn't throw an error, and shows something that is likely to be interpretable even for error states we haven't anticipated.

image
image
image
image


bensteinberg (approved):
LGTM!

PR #3642, opened 2024-10-30
Faster top_level_folders serialization

jcushman:
I did some late-afternoon performance profiling on the Perma dashboard, and found that the primary hotspot for admins is serializing their 3000 top-level folders to json.

This matches something I'd seen with CAP -- if you're returning thousands of items from a DRF endpoint, most of the time goes to turning dictionaries into Django model instances and using a serializer to turn them back into dictionaries. If you go dictionary->dictionary without the objects in the middle that time disappears. This is where python's usability-speed tradeoff wraps around on itself -- faster languages can afford the extra layers of abstraction, but python can't.

So, this PR rewrites user.top_level_folders() to return a Folder queryset instead of a list of folder instances, which takes a little trickiness to get the sorting right. Then the user serializer can fetch the dictionaries it wants directly with values(). I also added a test of the results for various users.

Note I didn't optimize the /folders endpoint yet to go dict->dict, just the /user endpoint, but I think we don't use the top-level folders endpoint for anything in the site itself.

In my local, this speeds up time to render the dashboard from several seconds to 150ms. The dashboard on prod takes about 6 seconds to render, which I'm hoping is also mostly due to this one thing.

### DIFF
--- a/perma_web/api/serializers.py
+++ b/perma_web/api/serializers.py
@@ -1,6 +1,7 @@
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.validators import URLValidator
+from django.db.models import F
 import requests
 from rest_framework import serializers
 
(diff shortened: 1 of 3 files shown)


bensteinberg (approved):
This is awesome

PR #3643, opened 2024-10-30
Save and play back WACZs

rebeccacremona:
...and configure the local dev environment's Scoop API to capture All The Things (TM), just like the prod/staging tiers of the Scoop API will do shortly.


bensteinberg (approved):

cmsetzer (approved):
Looks good!

Issue #3345, opened 2023-06-01
Users affiliated with more than 300 orgs can't select them all from the dropdown


rebeccacremona:
Similar to https://github.com/harvard-lil/perma/issues/3178.

Perma users who are affiliated with organizations can use this dropdown menu to select which of their organizations a link should be affiliated with:
![image](https://github.com/harvard-lil/perma/assets/11020492/7ecea204-84d4-4c78-b105-0073196d0d8a)

But, the dropdown only [shows 300 options](https://github.com/harvard-lil/perma/blob/develop/perma_web/static/js/create-link.module.js#L286)... and some registrars evidently have more than that!!! Amazing!

A dropdown with more than 300 options is.... absolutely not the UI we want to be offering people like that 😱 . And, indeed, they can successfully use the folder tree to accomplish the same task (though I should triple check that that is true). We should think through what design might be best for folks like these. 

But in the meantime... we should at least show them all their options.

rebeccacremona:
Hmm, though interesting, it looks like the largest registrar has only 198 orgs.
```
SELECT registrar_id, COUNT(*) as orgs FROM perma_organization GROUP BY registrar_id ORDER BY orgs DESC;

 registrar_id | orgs 
--------------+------
          648 |  198
          938 |  150
...
```
Maybe this is more complicated than I thought.

rebeccacremona:
Update: this is not more complicated than I thought 😂 .

The registrar user in question can indeed see all their orgs in the dropdown at Perma.cc, because there are fewer than 300.

Admins, however, see Perma's first 300 orgs, alphabetically.

So, this issue for the web app is currently only admin-facing.

The registrar user in question is instead hitting a limit in the browser extension; I made a separate ticket there.

jp-tosca:
Hi Rebecca! 

Probably a text field/dropdown where people can type and filter the elements could be helpful for longer lists and also a good addition to make it more usable. 

I haven't looked at this more deeply because I am just getting familiar with perma.cc but I will look more at it later, but I did some quick testing changing on `perma_web/perma/templates/user_management/create-link.html` around line 66: 

https://jsfiddle.net/jnbsdkv4/


![Animation](https://github.com/harvard-lil/perma/assets/142103991/f89b0d76-8e63-4cf8-bb0c-8f40029be107)

I will look more at this later but wanted to share this.

PR #3645, opened 2024-10-31
Improve folder dropdown

jcushman:
I started pulling a thread here, and I'm not sure if I pulled it the right amount before I stopped. So here's the story ...

After speeding up the initial page rendering on the dashboard, the next slow steps are loading the organizations and, if it's a big folder, the links:

image

which probably just means in general that the API is slow. But, I was curious about why the organizations were slow in particular, and realized that that query is almost entirely redundant of the top_level_folders data we provide in the initial page load and we may as well not do it. We first render `/api/v1/user` into the template, so send down everything as `current_user.top_level_folders`, and then we fetch `/api/v1/organizations?limit=300` to get the first 300 folders over again to render the folder dropdown and get the `default_to_private` setting for top level folders, and that second fetch takes a full second to run. Shouldn't we just not?

So, I added `registrar`, `registrar_name`, and `default_to_private` to the `/api/v1/user` `top_level_folders` response:

image

And with that extra data the page can now entirely skip that `/api/v1/organizations` fetch. This isn't much of a performance improvement, because page load wasn't waiting on that API call to return. But it's a correctness improvement for anyone with more than 300 orgs (no one?) -- before we would have shown orgs after the first 300 as `default_to_private=false` regardless of their real setting, and we wouldn't have put more than 300 into the dropdown.

We don't skip the sponsored folders fetch, because those aren't top-level folders, so that one sticks around.

So, mostly-useless performance improvement for me. But in the course of adapting the folder dropdown to use top_level_folders, I found and fixed some other issues:

* A couple of places were using the bare `current_user` global JS variable still, instead of `globalState` like they should. There are still more in JSTree which I didn't mess with cause that's its own refactor.

 * Folder dropdown wasn't showing the registrar dividers between orgs in the dropdown, now returned:
image

* Keyboard nav wasn't working on folder select. I returned the keyboard nav interactions [described here](https://github.com/harvard-lil/perma/pull/3507). This required a bunch of changes so I'm not sure when it broke.

* Folder dropdown had a weird gap below the button, which I removed. Before/after:
image
image

I think my design question is, are we happy jamming `registrar`, `registrar_name`, and `default_to_private` into `/api/v1/user` and calling it a day? The most correct thing would probably be to rethink the API a bit -- speed up the organizations endpoint, add a sponsorships endpoint, stop using the user.top_level_folders key. But I'd like to stop pulling the string for a minute so maybe this is fine.

Edit: this means a practical thing is, if you wish I hadn't messed with the API, I can pull out the optimization side of this and just keep the bug fixing side. That's a bit of a hassle but would be a reasonable ask.


rebeccacremona:
> But it's a correctness improvement for anyone with more than 300 orgs (no one?) -- before we would have shown orgs after the first 300 as default_to_private=false regardless of their real setting, and we wouldn't have put more than 300 into the dropdown.

See https://github.com/harvard-lil/perma/issues/3345

jcushman:
An interesting side effect of the correctness improvement is the dropdown rendering is now noticeably slow (maybe a second) for admins -- I thought I had broken something, but it's just that it has 10x as many entries.

perma_web/static/frontend/components/FolderSelect.vue
```diff
@@ -139,38 +159,47 @@ const handleSelection = (e) => {
...
           {{ getFolderHeader(folder) }}
         </li>
-        <li tabindex="-1" class="dropdown-item" role="option"
-            :aria-selected="selectedFolder.folderId === folder.id"
-            :data-folder="JSON.stringify(folder)"
+        <li 
+          tabindex="-1"
+          class="dropdown-item"
+          :data-index="index"
+          :data-id="folder.id"
```
rebeccacremona, line 190: I am not that familiar with the `listbox` role and its associated aria patterns; I can certainly look it up. Checking: did you drop the `role="option"` and `:aria-selected="selectedFolder.folderId === folder.id"` attributes here on purpose or by accident?

perma_web/static/frontend/components/FolderSelect.vue
```diff
@@ -60,12 +67,29 @@ const handleKeyboardSelectToggle = async (e) => {
     return
   }
```
rebeccacremona, line 69: Yikes, thanks for all these keyboard nav fixes. Happy to double check in stage; not going to pull down and check in advance.

perma_web/api/serializers.py
```diff
@@ -49,13 +49,30 @@ def get_field_names(self, *args, **kwargs):
...
+            'parent',
+            'has_children',
+            'path',
+            'organization',
+            'registrar',
+            'registrar_name',
+            'sponsored_by',
+            'is_sponsored_root_folder',
+            'read_only',
+            'default_to_private',
```
rebeccacremona, line 75: > I think my design question is, are we happy jamming registrar, registrar_name, and default_to_private into /api/v1/user and calling it a day?

Yes, I am totally onboard with adding these extra fields here.

> The most correct thing would probably be to rethink the API a bit -- speed up the organizations endpoint

Yeah I am running into more and more speed related things these days: things that worked the last time I reached for them are now too slow to use. I think we probably want to devote some time to this topic in general, some time soon.

rebeccacremona (approved):
Big fan of this refactor, even in the abstract, independently of the fixes. Thanks for tackling.

One question about accessibility/aria, but other than that, LGTM!

> I thought I had broken something, but it's just that it has 10x as many entries

Yeah, we clearly need some other option for admins. Although TBH... I don't think admins generally make links for orgs _at all_, and probably, we could have a better/more useful admin-only dashboard in general.

perma_web/static/frontend/components/FolderSelect.vue
jcushman: Oops, good catch, I intentionally deleted them but shouldn't have -- re-adding.

PR #3648, opened 2024-11-01
Bonus links and transactions sketch

jcushman:
This is a quick sketch to try to reduce opportunities for [deadlocks](https://library-innovation-lab-harvard.sentry.io/issues/6034662098/?referrer=slack&notification_uuid=1ccabc49-ad02-4145-9b77-ba73f48a5b6f&alert_rule_id=14453866&alert_type=issue) in saving or moving links.

* Don't use a transaction in user.save() unless needed. I'm not sure if that's relevant, but why not.
* Add an update_bonus_links() helper that does bonus_links updating in the database instead of on the instance, and then marks the instance's bonus_links value as outdated. This means in principle you can do lots of these in parallel without needing a lock on the user -- with the exception that you could give someone an extra link or two if they came in at just the right time.
* Be a notch more selective in when to update bonus_links in Folder.save() -- I'm not sure how often this matters.
* For clarity, don't let bonus_links be null.

Sketch:

* update_bonus_links is mostly AI implemented, I didn't test, but I think it's about right.
* I didn't remove many saves or select_for_updates here -- not sure yet if this approach itself helps, or just lets us potentially remove some locks elsewhere, or misses the problem.


jcushman:
> Be a notch more selective in when to update bonus_links in Folder.save() -- I'm not sure how often this matters.

Oh, I suspect this would have made the difference here! The deadlock is on this line:

image

And the deadlocked call was "UPDATE "perma_linkuser" SET "bonus_links" = **NULL** WHERE "perma_linkuser"."id" = %s", which suggests bonus_links weren't in play. So moving the save() up into the if block above would have prevented it entirely.

---

I think the other important question is, is this select_for_update still doing anything?

image

If that was just to protect the bonus_links count I think we can lose it, but maybe it's for something else too, in which case we could make it more specific.

PR #3651, opened 2024-11-07
Bulk organization user creation/addition feature

teovin:
#### This feature adds a new button to the organization user management view (/manage/organization-users): 

image

#### Upon clicking, the bulk upload form comes up:

image

#### It is also invalid if it doesn't include all of the column headers, if it doesn't have any users at all, if it doesn't include the email data for a given user, if it has duplicate users, and if an email address is invalid. 

image

image

image

image

image

image

image

#### The affiliation field can be toggled just like in the single user flow.

image

#### Once submitted, users that don't exist will be created; those that do will have their organization affiliation updated. New users will receive new user email with account activation email. Existing users will receive the user added to organization email.

image

image

#### A generic success message will be displayed at the end if there are no validation errors:

image

#### If some of the users get processed, but some can't due to those being an admin or registrar, below will be displayed:

image

#### If all users in the CSV are admins or registrars, the below error message will be displayed:

image

#### This PR also fixes a bug where the existing user email wasn't displaying the organization name.

image

**Disclaimer:** Any behavior and the wording here are open to feedback. This was my initial take on it :)

**EDIT:**

As of Nov 20th, I put this feature behind a feature flag so we can do some more testing on it in stage before a possible production deployment.


christiansmith:
Reviewing this with @bensteinberg  (running locally), we noticed a few minor details with the CSV handling form.

1. The CSV upload fails without a header row or with incorrect column names. 
	1. We could provide a template for users to download and edit 
	2. We could also handle the case of a missing header row and validate the structure of the provided data rows.
2. When there is a validation error, instructions should remain on the screen along with the error message.
3. If a malformed CSV _is_ uploaded, it would be good to have better feedback for the user than the current stack trace.

bensteinberg (commented):
As discussed, @christiansmith and I just looked at this together. I see this is added to the test for permissions, but otherwise has no tests.

Also, we were just discussing this with @rebeccacremona, who pointed out that we might prefer to assemble all the work and run it in a single database query, rather than going line-by-line. It might also make sense to fail if there are any validation errors, rather than create some users and not others.

teovin:
@christiansmith @bensteinberg Pushed a couple of commits for the desired behavior and also added a test. I am checking with Clare about below to see what her preference would be. My initial approach was that it is good to process the good data and ignore the rest, as in something is better than nothing if it makes sense. Will check with Becky on the point of running a single db query.

> It might also make sense to fail if there are any validation errors, rather than create some users and not others.

teovin:
@christiansmith @bensteinberg I made some changes to the PR since you last looked, I'd appreciate it if you could look again. Clare preferred to process the valid users and notify the requestor for those that were invalid, and I updated the PR description and UI messaging to reflect that.

perma_web/perma/forms.py
```diff
@@ -437,6 +440,161 @@ def save(self, commit=True):
...
+
+        for row in reader:
+            row_count += 1
+            email = row.get('email')
+            email = email.strip() if email else None
+
+            if not email:
+                raise forms.ValidationError("Each row in the CSV file must contain email.")
+            if not re.match(r'^[^\s@]+@[^\s@]+\.[^\s@]+$', email):
+                raise forms.ValidationError("CSV file must contain valid email addresses.")
```
rebeccacremona: Thanks for adding this check! I am not sure how that regex may differ from Django's own email validation (https://docs.djangoproject.com/en/5.1/ref/validators/#emailvalidator); I am not sure whether we need to concern ourselves with any discrepancies between the two. It may be safest to switch to use the Django validator, but it probably doesn't matter 🤷‍♀️ . Do you have any thoughts about that?

perma_web/perma/forms.py
```diff
@@ -437,6 +440,161 @@ def save(self, commit=True):
...
+        reader = csv.DictReader(file)
+
+        # validate the headers
+        headers = reader.fieldnames
+        if not all(item in headers for item in ['first_name', 'last_name', 'email']):
+            raise forms.ValidationError("CSV file must contain a header row with first_name, last_name and email columns.")
+
+        # validate the rows
+        seen = set()
+        row_count = 0
```
rebeccacremona, line 503: I noticed something fun. If you want, you actually don't need the `seen` set or the `row_count` integer! `if email in self.user_data` is the same as `if email in seen`, and `row_count` is the same as `len(self.user_data)` 🙂 But if you prefer the aesthetics of `seen` and `row_count`, no objections here!

perma_web/perma/forms.py
rebeccacremona: I am also wondering what you think about mentioning the invalid email address in the error message, so that the user knows which one is bad.

perma_web/perma/forms.py
```diff
@@ -437,6 +440,161 @@ def save(self, commit=True):
...
+            email = row.get('email')
+            email = email.strip() if email else None
+
+            if not email:
+                raise forms.ValidationError("Each row in the CSV file must contain email.")
+            if not re.match(r'^[^\s@]+@[^\s@]+\.[^\s@]+$', email):
+                raise forms.ValidationError("CSV file must contain valid email addresses.")
+
+            if email in seen:
+                raise forms.ValidationError("CSV file cannot contain duplicate users.")
```
rebeccacremona: What would you think of including the duplicate email address in the error message, so the user knows which one is repeated?

perma_web/perma/forms.py
```diff
@@ -437,6 +440,161 @@ def save(self, commit=True):
...
+        file.seek(0)
+        self.cleaned_data['csv_file'] = file
+        return file
+
+    def save(self, commit=True):
+        expires_at = self.cleaned_data['expires_at']
+        organization = self.cleaned_data['organizations']
+
+        emails = set(self.user_data.keys())
+        existing_users = LinkUser.objects.filter(email__in=emails)
```
rebeccacremona: Ah, we still have that small bug that I mentioned: in Perma, the `email` field is always an all-lowercase version of the user's `raw_email`. The csv could have email addresses with capital letters. So, you will need to cast the email addresses to lowercase, before trying to run this query, otherwise we might miss people.

perma_web/perma/forms.py
```diff
@@ -437,6 +440,161 @@ def save(self, commit=True):
...
+                    self.ineligible_users[user.email] = user
+                else:
+                    updated_user_affiliations.append(user)
+                    self.updated_users[user.email] = user
+        
+        new_user_emails = emails - set(self.ineligible_users.keys()) - set(self.updated_users.keys())
+        created_user_affiliations = []
+
+        if new_user_emails and commit:
+            for user in new_user_emails:
```
rebeccacremona: I think it is a little confusing this variable is named "user", when it is an email. what would you think of `for email in new_user_emails`?

perma_web/perma/forms.py
```diff
@@ -437,6 +440,161 @@ def save(self, commit=True):
...
+                self.created_users[user] = new_user
+
+                created_user_affiliations.append(
+                    UserOrganizationAffiliation(
+                        user=new_user,
+                        organization=organization,
+                        expires_at=expires_at
+                    )
+                )
+
```
rebeccacremona, line 575: I regret am having a little trouble reading this section. I think it is partly because of variable names.

`created_user_affiliations` is a list of affiliation objects (which makes sense to me) but, the similarly named `updated_user_affiliations` is a list of user objects. 

similarly with `preexisting_affiliations_set` and `all_user_affiliations`; both are sets, and both are sets of user objects. 

I am confused whether `updated_user_affiliations` and `all_user_affiliations` are in fact needed. Is this true? `updated_user_affiliations` is only used to make `all_user_affiliations`, and `all_user_affiliations`  is the same as `set(self.updated_users.values())`? If that's true, I think this would be easier to read if you used that directly.

perma_web/perma/forms.py
```diff
@@ -437,6 +440,161 @@ def save(self, commit=True):
...
+        if commit:
+            # create the affiliations for new users
+            UserOrganizationAffiliation.objects.bulk_create(created_user_affiliations)
+
+            # create or update the affiliations of existing users
+            # affiliations that already exist
+            preexisting_affiliations = (UserOrganizationAffiliation.objects.filter(user__in=updated_user_affiliations,
```
rebeccacremona, line 582: This is just a thought about readability. Do you think it would be easier to read if, right here after you get the existing affiliation objects, you did the update? Then, you could do the calculations to figure out if any new affiliation objects need to be created, and then create them, in its own section with its own comment. I think that might be clearer than doing them together, but you may disagree 🙂

perma_web/perma/forms.py
```diff
@@ -437,6 +440,161 @@ def save(self, commit=True):
...
+                                                                                   organization=organization))
+
+            preexisting_affiliations_set = set(affiliation.user for affiliation in preexisting_affiliations)
```
rebeccacremona, line 585: Ah, here's a good Django + database thing to know about! Since you are iterating through the `UserOrganizationAffiliation` queryset and fetching each object's related `user` object, this will make one database query per object. Django includes a utility for situations like this, `select_related`: https://docs.djangoproject.com/en/5.1/ref/models/querysets/#select-related. If you call `UserOrganizationAffiliation.objects.filter(...).select_related('user')`, then Django will get all those user objects in one single query!

perma_web/perma/views/user_management.py
```diff
@@ -747,26 +752,69 @@ def get_context_data(self, **kwargs):
...
+            messages.add_message(self.request, level, f'<h4>{title}</h4>{body}', extra_tags='safe')
+
+        def send_emails(users, email_function, email_template, extra_context, *args):
+            for obj in users.values():
+                try:
+                    if email_function == 'email_new_user':
+                        email_new_user(*args, obj, email_template, extra_context)
+                    else:
+                        send_user_email(obj.raw_email, email_template, extra_context)
+                except Exception as e:
```
rebeccacremona, line 766: Thanks for adding this error handling!

As mentioned, I am a little worried that, when we are actually calling out to the mailgun API and actually sending emails, that it will be too slow to do this inline, resulting in users who upload large CSVs seeing a 502 despite the request succeeding (like happens with large link batches). I think we might need to make this async. 

Possibly to be discussed!! Even if we do, that definitely doesn't have to be part of this PR!!

rebeccacremona (changes requested):
This is looking great! This feature has been a lot of hard work, and it shows! 

I noticed one tiny bug: if the CSV includes email addresses of existing users and uses capital letters, we will not find them and will try to create a new user with the capitalized email address (which will fail).

I think that bug is worth fixing before merging. But that is the only one I saw!

I also left some thoughts on readability, a few tweaks a person might consider making, a few things we might want to think about in the future, etc. 

Please take or leave that feedback 🙂 I don't think any of it, even if it's something we want to do, should stand in the way of this being merged 🙂.

perma_web/perma/views/user_management.py
teovin: I will test in stage to see at what point we start seeing failures. Depending on that, I can add a limit to the rows.

perma_web/perma/forms.py
```diff
@@ -505,7 +505,7 @@ def clean_csv_file(self):
         for row in reader:
             row_count += 1
             email = row.get('email')
-            email = email.strip() if email else None
+            email = email.strip().lower() if email else None
```
rebeccacremona: Note that this will result in us actually saving the lowercase version of the email in `raw_email`, rather than preserving the submitted capitalization. Not that big a deal; a bug similar to the long-standing LIL-2921. If you can think of a way to just cast to lowercase before filtering, that might be best, but if you cannot, we can fix that some other time (maybe along with LIL-2921).

teovin: Ahh, very good point! Let me know if what I pushed is too hacky..

rebeccacremona: Great, thanks for doing this!

rebeccacremona (approved):

PR #3654, opened 2024-11-13
Save only WACZ, not WARC

rebeccacremona:
We recently began saving the WACZs produced by Scoop, in addition to continuing to save the WARCs, the archive format that Perma has long supported. 

This is entirely redundant: the WACZ is just the WARC and other useful files bundled together in a ZIP and given the `.wacz` file extension.

This PR stops Perma from saving the WARC separately. And then:

- It tweaks any part of Perma's code that wants a WARC to use a utility function, so that we continue retrieving the WARC out of storage if we have it (that is, all the WARCs we already have saved), but if we don't (everything we'll have going forward), extracts it from the WACZ instead.

- It removes any features of Perma that presume we have a WARC in storage:
- + the application-level feature-flag (currently toggled off) that forces us to play back only from stored WARCs, even if WACZs are available
- + the admin-only playback control, that lets an admin opt in to a WARC playback even when a WACZ is available, for comparison purposes (which is a feature we'd LOVE to have going forward, but will be tricky to implement... to be discussed at a later date 🙂) 

It updates the tests accordingly.

It is pretty simple, despite the size of the diff.

I would love thoughts on whether the new method, `link.get_warc()`, is implemented elegantly and with the parameters you as a dev would like to have, or whether it could be better.


bensteinberg (approved):
LGTM! Nice!

Issue #3008, opened 2022-01-05
Decrease size of HistoricalLink table


rebeccacremona:
We use Django Simple History to track changes on a number of our models. The historical link table is huge:
```
+---------------------------------------+--------+-------+-------+------------+---------+
| CONCAT(table_schema, '.', table_name) | rows   | DATA  | idx   | total_size | idxfrac |
+---------------------------------------+--------+-------+-------+------------+---------+
| perma.perma_historicallink            | 18.76M | 5.66G | 2.30G | 7.96G      |    0.41 |
| perma.perma_link                      | 2.16M  | 0.81G | 0.92G | 1.73G      |    1.14 |
| perma.django_session                  | 2.83M  | 1.02G | 0.20G | 1.21G      |    0.20 |
| perma.perma_uncaughterror             | 0.81M  | 1.01G | 0.02G | 1.03G      |    0.02 |
| perma.perma_historicallinkuser        | 2.64M  | 0.54G | 0.45G | 1.00G      |    0.83 |
| perma.perma_capture                   | 5.34M  | 0.73G | 0.21G | 0.94G      |    0.28 |
| perma.perma_historicalregistrar       | 1.82M  | 0.44G | 0.08G | 0.53G      |    0.19 |
| perma.perma_capturejob                | 1.77M  | 0.31G | 0.21G | 0.52G      |    0.69 |
| perma.perma_historicalorganization    | 1.73M  | 0.15G | 0.15G | 0.30G      |    1.00 |
| perma.perma_link_folders              | 2.33M  | 0.14G | 0.14G | 0.28G      |    1.06 |
...
```
The routine creation of a link makes 3 rows, one of which is a duplicate which I think [`clean_duplicate_history`](https://django-simple-history.readthedocs.io/en/latest/utils.html#clean-duplicate-history) will remove. 

Maybe let's try running that, and see how it goes? The table is so big, I'm not sure it will run painlessly, but we can hope. We probably want to run it regularly, maybe as an hourly/daily celerybeat task.

Also, Jack says, "seems totally viable to `clean_old_history` with some long time window, too."

rebeccacremona:
We're trying a customized version of the duplicate cleanup command on a clone of the prod db, to see how much it reduces the final size and how long it takes, on a db where nothing else is happening. For this experiment, we're running it as a fab task in 4 tmux sessions on one minion; in real life we probably would run on the IA minions, maybe mediated by celery, maybe not? TBD.

```
@task
[...]

PR #3660, opened 2024-11-19
Remove HistoricalLink

bensteinberg:
This PR removes the HistoricalLink model. In production, this table takes up about a third of the database, and as far as I know, has never been used for anything. Removing this will make backups smaller and backing up quicker; it will also likely make an upcoming database upgrade quicker.


rebeccacremona (approved):
Yes! Let's finally do this!!

bensteinberg:
@rebeccacremona points out issue #3008, "Decrease size of HistoricalLink table". My instinct is to drop the table entirely, rather than reducing it by a third (?), with a fair amount of fussing. Also, note that the size of the table in production has nearly doubled in size since that issue was created at the beginning of 2022.

cmsetzer (approved):

PR #3659, opened 2024-11-18
Add approval workflow for paid registrar requests

cmsetzer:
This builds upon our work in https://github.com/harvard-lil/perma/pull/3595 and https://github.com/harvard-lil/perma/pull/3631 to add an approval workflow for sign-up requests by paid registrars (a.k.a. "firms"). An approval workflow already exists for library sign-up requests; this branch generalizes the library approval form so it supports firms as well.

## Before

* User submitted a paid registrar request at https://perma.cc/sign-up/firms
* Perma admins then emailed them, going back and forth a few times to get registrar info and set a base rate
* Perma admins then used the Perma DB admin interface to manually create a registrar, populate fields, and assign the base rate

## After

* User submits paid registrar request at https://perma.cc/sign-up/firms
* A pending registrar is automatically created in the Perma DB with registrar name, email, etc. populated
* An email is sent out to user and Perma admins with all the field data supplied by the user
* Using the approval link in the email (or by clicking a button on the Manage Registrars screen), Perma admins can approve/reject the registrar request and change the base rate in one place
  + As we've discussed, if the pending registrar does not already have a registrar user associated, Perma admins must assign a registrar user before they can approve/reject the registrar. A user search interface is provided on the approval page

PR #3668, opened 2024-11-26
Annotate query results with QuerySet instead of BaseManager

cmsetzer:
In https://github.com/harvard-lil/perma/pull/3659, we discussed finding a better type annotation for Django ORM querysets than `BaseManager`.

I had the impression we'd need to add an external dependency for this, but it turns out we can in fact do `QuerySet[LinkUser]` and so on. Moreover, [django-stubs](https://github.com/typeddjango/django-stubs) seems to have its own nuances (and only supports Mypy?); seems not worth bringing in at present.

This branch swaps `QuerySet` type hints in for all our existing `BaseManager`s. There should be zero impact on Perma functionality, as this is purely a dev-side change.


rebeccacremona (approved):
Oh great!!

PR #3670, opened 2024-12-03
Tools for working with WACZs

rebeccacremona:
See LIL-2863.

We recently started saving WACZs from Scoop instead of WARCs. We have long had some facility for users to download Perma Links' WARC files, both using the API and via the GUI. This PR adds identical functionality for WACZ files:

- `wacz_size` and `wacz_download_url` fields have been added to the `Link` and `AuthenticatedLink` serializers, so that any API route that includes information about `Link` objects will now return that info

- the `warc_download_url` field was updated: it will now be populated if a WARC _or_ a WACZ is available for a given Perma Link, since the WARC can be extracted from the WACZ and served.

- a "Download WACZ" button has been added to the tray next to the "Download WARC" button, when a WACZ is available:

image
 
For compatibility, that "button" is a link pointing to a URL of the form `https://perma.test:8000/B925-DU9S?type=wacz_download`, following the pattern of the WARC "button", which is a URL of the form `https://perma.test:8000/B925-DU9S?type=warc_download`. (This pattern predates the existence of the `/download` API route.)

The only change in existing functionality: now, if a Perma Link redirects to another Perma Link (as with `https://perma.cc/AAAA-AAAA`), the file format will always be included in the query string of the redirected API request to download it (e.g. `https://api.perma.cc/v1/archives/69AE-PWJB/download?file_format=warc`).


rebeccacremona:
Edge case: I should double check (and add to the test suite) what happens when you manually request to download the WACZ of a Perma Link for which there is only a WARC. The button won't appear in the tray, and the API will report `wacz_download_url: null`, but the request shouldn't result in a RuntimeError...

rebeccacremona:
> the request shouldn't result in a RuntimeError...

Ah, yep, if you manually construct URLs like `https://perma.test:8000/Y3YJ-4B9G?type=wacz_download` or `https://perma.test:8000/api/v1/archives/Y3YJ-4B9G/download?file_format=wacz` for links without WACZs, say, old ones, or user loads, that's currently a 500 w/RuntimeError. 

~Fixing!~ Fixed.

bensteinberg (approved):

christiansmith (approved):
Looks good to me! Nice work @rebeccacremona!

PR #3673, opened 2024-12-04
Tiny tweaks to form processing

rebeccacremona:
This PR makes just a few tweaks to the how the new form for adding many users to an organization at a time saves data. 

- it adjusts when `if commit` is checked, so that all the logic except for the actual write operations is run if `commit` is `False`

- it creates new `OrganizationAffiliation` objects all together in one batch, instead of two batches (one for brand new users and one for existing users)

- when retrieving the collection of existing affiliation objects, it uses `select_related('user')`, so that later, when running `set(affiliation.user for affiliation in preexisting_affiliations)`, each call to `affiliation.user` does not trigger a separate call to the database

- it adjusts how we handle potentially mixed/uppercase and cast-to-lowercase email addresses so that a) duplicate emails are detected in the CSV even if they differ in casing, and b) we can retrieve the raw email address from `self.user_data`

I took the liberty of renaming a few variables, adding a few comments, and slightly changing the order of operations, so that existing users are handled first, and then we move on to new users.

### DIFF
--- a/perma_web/perma/forms.py
+++ b/perma_web/perma/forms.py
@@ -499,11 +499,7 @@ class MultipleUsersFormWithOrganization(ModelForm):
             raise forms.ValidationError("CSV file must contain a header row with first_name, last_name and email columns.")
 
         # validate the rows
-        seen = set()
-        row_count = 0
-
         for row in reader:
-            row_count += 1
             email = row.get('email')
             email = email.strip() if email else None
 
@@ -516,7 +512,7 @@ class MultipleUsersFormWithOrganization(ModelForm):
             except ValidationError:
                 raise forms.ValidationError(f"CSV file contains invalid email address: {email}")
 
-            if email in seen:
+            if email in self.user_data:
                 raise forms.ValidationError(f"CSV file cannot contain duplicate users: {email}")
             else:
                 seen.add(email)
@@ -525,7 +521,7 @@ class MultipleUsersFormWithOrganization(ModelForm):
                     'last_name': row.get('last_name', '').strip()
                 }
 
-        if row_count == 0:
+        if not len(self.user_data):
             raise forms.ValidationError("CSV file must contain at least one user.")
 
         file.seek(0)
--- a/perma_web/perma/forms.py
+++ b/perma_web/perma/forms.py
@@ -531,18 +531,19 @@ class MultipleUsersFormWithOrganization(ModelForm):
     def save(self, commit=True):
         expires_at = self.cleaned_data['expires_at']
         organization = self.cleaned_data['organizations']
-        affiliations_to_create = []
 
         all_emails = set(self.user_data.keys())
-        existing_users = LinkUser.objects.filter(email__in=all_emails)
+        affiliations_to_create = []
 
+        # find any existing users, and exclude any that are ineligible to become org users
+        existing_users = LinkUser.objects.filter(email__in=all_emails)
         for user in existing_users:
             if user.is_staff or user.is_registrar_user():
                 self.ineligible_users[user.email] = user
             else:
                 self.updated_users[user.email] = user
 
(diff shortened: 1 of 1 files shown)


perma_web/perma/forms.py
```diff
@@ -516,16 +512,16 @@ def clean_csv_file(self):
...
-                seen.add(email)
-                self.user_data[email] = {
+                self.user_data[email.lower()] = {
+                    'raw_email': email,
                     'first_name': row.get('first_name', '').strip(),
                     'last_name': row.get('last_name', '').strip()
                 }
 
-        if row_count == 0:
+        if not len(self.user_data):
```
cmsetzer: This is trivial, but you could simplify the condition even further to:
```suggestion
        if not self.user_data:
```

perma_web/perma/forms.py
```diff
@@ -536,69 +532,58 @@ def save(self, commit=True):
...
+        preexisting_affiliations = UserOrganizationAffiliation.objects.filter(
+            user__in=self.updated_users.values(),
+            organization=organization
+        ).select_related('user')
+        if preexisting_affiliations and commit:
+            preexisting_affiliations.update(expires_at=expires_at)
+
+        # build affiliation objects for existing users that need them
+        users_with_existing_affiliations = set(affiliation.user for affiliation in preexisting_affiliations)
+        users_without_existing_affiliations = set(self.updated_users.values())- users_with_existing_affiliations
```
cmsetzer: I think a space got dropped here:
```suggestion
        users_without_existing_affiliations = set(self.updated_users.values()) - users_with_existing_affiliations
```

perma_web/perma/forms.py
```diff
@@ -536,69 +532,58 @@ def save(self, commit=True):
...
+        for user in users_without_existing_affiliations:
+            affiliations_to_create.append(UserOrganizationAffiliation(
+                user=user,
+                organization=organization,
+                expires_at=expires_at
+            ))
+
+        # create new users and their affiliation objects
+        new_user_emails = all_emails - set(self.ineligible_users.keys()) - set(self.updated_users.keys())
```
cmsetzer: You can drop the calls to `dict.keys()` if you want, although maybe it is clearer as is?

```suggestion
        new_user_emails = all_emails - set(self.ineligible_users) - set(self.updated_users)
```

cmsetzer (approved):
Looks good to me. I left a few comments about minor points of coding style; feel free to adopt or ignore as you prefer.

perma_web/perma/forms.py
rebeccacremona: Ha! Good point, not sure what happened mentally there :-)

perma_web/perma/forms.py
rebeccacremona: Thanks!

perma_web/perma/forms.py
```diff
@@ -536,69 +532,58 @@ def save(self, commit=True):
         expires_at = self.cleaned_data['expires_at']
         organization = self.cleaned_data['organizations']
 
-        raw_emails = set(self.user_data.keys())
-        # lower casing the emails to feed into the filter query in order to prevent duplicate user creation
-        lower_case_emails = {email.lower() for email in self.user_data.keys()}
-        existing_users = LinkUser.objects.filter(email__in=lower_case_emails)
-        updated_user_affiliations = []
+        all_emails = set(self.user_data.keys())
```
rebeccacremona:
```suggestion
        all_emails = set(self.user_data)
```

perma_web/perma/forms.py
rebeccacremona: Ah, I didn't realize that would happen implicitly. Good to know!

teovin (approved):

PR #3679, opened 2024-12-16
Save user uploads as WACZs

bensteinberg:
This is a first cut at preserving user uploads as WACZ files rather than WARCs. I'm not sure which Linear ticket is the best one to link here. I'm making this a draft for now.

This works, in the sense that it produces a valid WACZ that replayweb.page can play back, but it does not yet play back in Perma; the error message, from wabac.js. is e.g.

```
Archived Page Not Found

Sorry, this page was not found in this archive:

file:///8MXD-LZ6V/upload.jpg?version=040666551871161394
...
```
(That page is in fact in the archive.)

I'm guessing I need a slightly different set of options to py-wacz.

### DIFF
--- a/perma_web/api/tests/test_link_authorization.py
+++ b/perma_web/api/tests/test_link_authorization.py
@@ -147,13 +147,13 @@ class LinkAuthorizationTestCase(LinkAuthorizationMixin, ApiResourceTestCase):
         #   capture were properly associated with actual web archive files, which is always
         #   the case outside of tests
         self.link.archive_timestamp = timezone.now() + timedelta(1)
-        self.link.warc_size = 1
+        self.link.warc_size = 0
         self.link.wacz_size = 1
         self.link.save()
 
-        # This link has a warc and a wacz
+        # This link has a wacz and no warc
         self.link.refresh_from_db()
-        self.assertTrue(self.link.warc_size)
+        self.assertFalse(self.link.warc_size)
         self.assertTrue(self.link.wacz_size)
 
         old_primary_capture = self.link.primary_capture
@@ -168,10 +168,10 @@ class LinkAuthorizationTestCase(LinkAuthorizationMixin, ApiResourceTestCase):
 
         self.assertTrue(Capture.objects.filter(link_id=self.link.pk, role='primary').exclude(pk=old_primary_capture.pk).exists())
 
-        # This link now only has a warc, but not a wacz
+        # This link still only has a wacz
         self.link.refresh_from_db()
-        self.assertTrue(self.link.warc_size)
-        self.assertFalse(self.link.wacz_size)
+        self.assertFalse(self.link.warc_size)
+        self.assertTrue(self.link.wacz_size)
 
 
     def test_should_reject_patch_with_file_for_out_of_window_link(self):
--- a/perma_web/api/tests/test_link_authorization.py
+++ b/perma_web/api/tests/test_link_authorization.py
@@ -147,14 +147,10 @@ class LinkAuthorizationTestCase(LinkAuthorizationMixin, ApiResourceTestCase):
         #   capture were properly associated with actual web archive files, which is always
         #   the case outside of tests
         self.link.archive_timestamp = timezone.now() + timedelta(1)
-        self.link.warc_size = 0
         self.link.wacz_size = 1
         self.link.save()
 
-        # This link has a wacz and no warc
         self.link.refresh_from_db()
-        self.assertFalse(self.link.warc_size)
-        self.assertTrue(self.link.wacz_size)
 
         old_primary_capture = self.link.primary_capture
 
@@ -167,11 +163,12 @@ class LinkAuthorizationTestCase(LinkAuthorizationMixin, ApiResourceTestCase):
                                   data={'file':file_content})
 
         self.assertTrue(Capture.objects.filter(link_id=self.link.pk, role='primary').exclude(pk=old_primary_capture.pk).exists())
+        self.assertTrue(Capture.objects.filter(link_id=self.link.pk, role='provenance_summary').exclude(pk=old_primary_capture.pk).exists())
 
-        # This link still only has a wacz
         self.link.refresh_from_db()
-        self.assertFalse(self.link.warc_size)
+
         self.assertTrue(self.link.wacz_size)
+        self.assertTrue(self.link.wacz_size != 1)
 
 
     def test_should_reject_patch_with_file_for_out_of_window_link(self):
--- a/perma_web/api/tests/test_link_authorization.py
+++ b/perma_web/api/tests/test_link_authorization.py
@@ -163,7 +163,7 @@ class LinkAuthorizationTestCase(LinkAuthorizationMixin, ApiResourceTestCase):
                                   data={'file':file_content})
 
         self.assertTrue(Capture.objects.filter(link_id=self.link.pk, role='primary').exclude(pk=old_primary_capture.pk).exists())
-        self.assertTrue(Capture.objects.filter(link_id=self.link.pk, role='provenance_summary').exclude(pk=old_primary_capture.pk).exists())
+        self.assertTrue(Capture.objects.filter(link_id=self.link.pk, role='provenance_summary').exists())
 
         self.link.refresh_from_db()
 
--- a/perma_web/api/tests/test_link_resource.py
+++ b/perma_web/api/tests/test_link_resource.py
@@ -94,7 +94,7 @@ class LinkResourceTestMixin():
             'private_reason',
         ]
 
-    def assertRecordsInArchive(self, link, upload=False, expected_records=None, check_screenshot=False, check_provenance_summary=False, filetype='wacz'):
+    def assertRecordsInArchive(self, link, upload=False, expected_records=None, check_screenshot=False, check_provenance_summary=False):
 
         def find_recording_in_warc(index, capture_url, content_type):
             warc_content_type = "application/http; msgtype=response"
@@ -129,8 +129,7 @@ class LinkResourceTestMixin():
         self.assertTrue(link.primary_capture.content_type, "Capture is missing a content type.")
 
         # create an index of the warc
-        extract = filetype == 'wacz'
-        with link.get_warc(extract) as warc_file:
+        with link.get_warc() as warc_file:
             index = index_warc_file(warc_file)
 
         # see if the index reports the content is in the warc
@@ -655,7 +654,7 @@ class LinkResourceTransactionTestCase(LinkResourceTestMixin, ApiResourceTransact
                                        user=self.org_user)
 
             link = Link.objects.get(guid=obj['guid'])
-            self.assertRecordsInArchive(link, upload=True, filetype='wacz')
+            self.assertRecordsInArchive(link, upload=True)
             self.assertEqual(link.primary_capture.user_upload, True)
 
     def test_should_create_archive_from_jpg_file(self):
@@ -666,7 +665,7 @@ class LinkResourceTransactionTestCase(LinkResourceTestMixin, ApiResourceTransact
                                        user=self.org_user)
 
             link = Link.objects.get(guid=obj['guid'])
-            self.assertRecordsInArchive(link, upload=True, filetype='wacz')
+            self.assertRecordsInArchive(link, upload=True)
             self.assertEqual(link.primary_capture.user_upload, True)
 
     def test_should_reject_jpg_file_with_invalid_url(self):
@@ -687,7 +686,7 @@ class LinkResourceTransactionTestCase(LinkResourceTestMixin, ApiResourceTransact
 
             link = Link.objects.get(guid=obj['guid'])
             self.assertEqual(link.submitted_url, 'http://asdf.asdf')
-            self.assertRecordsInArchive(link, upload=True, filetype='wacz')
+            self.assertRecordsInArchive(link, upload=True)
             self.assertEqual(link.primary_capture.user_upload, True)
 
     def test_should_reject_invalid_file(self):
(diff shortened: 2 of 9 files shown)


bensteinberg:
Apart from getting this to work, the main question is what the metadata should look like.

bensteinberg:
I think the problem here is that the CDX index is broken, because the entries for `file:///` URIs are getting rewritten to reduce consecutive slashes to one; I think this is down to [cdxj_indexer just using `surt` on the URL](https://github.com/webrecorder/cdxj-indexer/blob/9ad2b9e1c54d2d20c391050fdb831ca1ee981504/cdxj_indexer/main.py#L301-L305), rather than doing [what warcio does with `getSurt()`](https://github.com/webrecorder/warcio.js/blob/71e4fdd03b4fefb4849f3d8a7449118a33730579/src/lib/utils.ts#L33-L35), in the Scoop context:
```
    if (!url.startsWith("https:") && !url.startsWith("http:")) {
      return url;
    }
```
I'm going to see if I can demonstrate this. If this is the problem, I'm not sure where to make a fix.

rebeccacremona:
(A possible thread to pull when working further with this: does using a `response` type record instead of a `resource` record affect the indexing? It shouldn't, in theory... but, since I recall that in Scoop-produced WARCs, everything is a response record, including attachments like the screenshot, provenance summary, etc., and I do not know why, I am a little suspicious this could be related.)

bensteinberg:
Thanks. I've tried with both response and resource records; not sure it matters here. (Philosophically, these might better be resource records.) I've removed the dependency on the indexer and am now creating the WACZ inline. I'm still not producing a working file, but I've pushed the changes for now.

bensteinberg:
The thread I'm pulling now is that replayweb claims "2 hashes verified, 1 invalid" -- but I haven't been able to find an invalid hash. I am not perfectly confident in the way I'm calculating offset in the WARC, which I suppose could be the problem.

rebeccacremona:
I can try and find the code we used to use for making CDXlines, back when that was a function of Perma. Those were not CDXJ, but it could be interesting just to see.

bensteinberg:
One of my thoughts here is that since what we are trying to do is so minimal, there's no need for a general solution to indexing or any other part of WACZ production.

bensteinberg:
Thanks @rebeccacremona and @matteocargnelutti for the help!

> replayweb claims "2 hashes verified, 1 invalid"

(This is true for working WACZs.)

Is it worth adding the WARC header REFERS_TO_TARGET_URI?

perma_web/api/tests/test_link_authorization.py
```diff
@@ -168,10 +168,10 @@ def test_should_allow_user_to_patch_with_file(self):
 
         self.assertTrue(Capture.objects.filter(link_id=self.link.pk, role='primary').exclude(pk=old_primary_capture.pk).exists())
 
-        # This link now only has a warc, but not a wacz
+        # This link still only has a wacz
         self.link.refresh_from_db()
-        self.assertTrue(self.link.warc_size)
-        self.assertFalse(self.link.wacz_size)
+        self.assertFalse(self.link.warc_size)
+        self.assertTrue(self.link.wacz_size)
```
rebeccacremona, line 170: I think this test needs further changes, with this new setup, to prove that the patch worked. Before, demonstrating that once there was a WACZ and now there is a WARC demonstrated that the process succeeded. Now, just asserting that warc_size is still zero and wacz_size is still non-zero doesn't, I think, assert much.

Instead, I think it would be more meaningful to a) check for the presence of a new `Capture` object with `role='provenance_summary'` (more on that below) and b) assert that the wacz_size is different than what it was before. That still doesn't assert anything about the contents of the WACZ (easier to test an upload-your-own playback in our functional tests, if we ever wanted to), but I think that might be more complete?

perma_web/api/tests/test_link_resource.py
```diff
@@ -655,7 +655,7 @@ def test_should_create_archive_from_pdf_file(self):
                                        user=self.org_user)
 
             link = Link.objects.get(guid=obj['guid'])
-            self.assertRecordsInArchive(link, upload=True, filetype='warc')
+            self.assertRecordsInArchive(link, upload=True, filetype='wacz')
```
rebeccacremona: With this change, I suspect that having `filetype` as a parameter of `assertRecordsInArchive` is now unnecessary, and that the filetype is now always wacz, and that whole branch can be removed. Do you agree?

perma_web/perma/templates/provenance-summary.html
```diff
@@ -0,0 +1,103 @@
...
+  </head>
+
+  <body>
+    
+    <main>
+
+      <header>
+        <h1>Provenance Summary</h1>
+        <p>The data present in this capture were uploaded by a Perma user to replace a failed or unsatisfactory capture of {{ url }} on {{ now }}.</p>
+      </header>
```
rebeccacremona, line 96: Hooray for a provenance summary!

TBD! I think probably let's add some more details here before this goes live?

perma_web/perma/utils.py
```diff
@@ -498,26 +504,208 @@ def decrypt_from_perma_payments(ciphertext, encoder=encoding.Base64Encoder):
...
+
+        # ...datapackage...
+        datapackage = {
+            "profile": "data-package",
+            "wacz_version": "1.1.1",
+            "title": title,
+            "description": f"User upload for {url}",
+            "mainPageURL": warc_url,
+            "created": ts_string,
+            "software": "Perma.cc",  # version?
```
rebeccacremona: `# version?`

Hmmm... 

So if we wanted to consider the production "tag" the version, then we'd have to somehow expose that to the application and interpolate here... Which sounds like a bit of a pain in the butt? (Some salt state that runs a command that gets the current tag and puts it into an ENV var, and then an application setting that reads that ENV var?)

Some new version that we just increment if/when we change this code (we'll be prone to forget, I imagine).

Or skip entirely?

Hmmmmmmmmm.

perma_web/perma/models.py
```diff
@@ -1997,14 +1996,23 @@ def write_uploaded_file(self, uploaded_file, cache_break=False):
                           user_upload='True',
                           content_type=mime_type,
                           url=warc_url)
```
rebeccacremona: We'll want to make another `Capture` object here, with `role='provenance_summary'`, to go along with the primary.

See also the discussion about whether, at this time, we should remove the random string appended to the end. If we are going to do that, I think we should do it now rather than later, make the corpus more consistent.

rebeccacremona (changes requested):
Thanks so much for taking this on!! This is cool!

I read through your code once so far; I'd love to take a more detailed look later, including looking at a couple of the finished WACZs. But in the meantime, shared a few thoughts.

> replayweb claims "2 hashes verified, 1 invalid"
> (This is true for working WACZs.)

Hmmm, interesting. I wonder if this is true just for these new WACZs or for other WACZs we produce... Maybe let's dig.

> Is it worth adding the WARC header REFERS_TO_TARGET_URI?

Interesting. TBD!

perma_web/api/tests/test_link_authorization.py
bensteinberg: Yeah, sounds good -- let me see about adding and checking a new `Capture` object.

perma_web/perma/templates/provenance-summary.html
bensteinberg: What details do you have in mind? This is minimal partly because it's a placeholder and opportunity for discussion, but also because I'm inclined to keep the file as short as possible. I don't think it's necessary to provide the level of provenance detail Scoop provides.

perma_web/perma/utils.py
bensteinberg: Yeah, the automated mechanism for accomplishing this would be slightly baroque and the manual way would be failure-prone. I think I still want to add it? Let me think about the best approach -- I think it will be something like what you describe, maybe writing settings directly rather than using an env var.

perma_web/perma/models.py
bensteinberg: > the discussion about whether, at this time, we should remove the random string appended to the end

I haven't seen this discussion yet -- is it a comment on this PR, or elsewhere?

perma_web/perma/utils.py
bensteinberg: If we do set this up, should the setting be visible in templates? Would we want to expose it elsewhere in the application?

bensteinberg:
> > replayweb claims "2 hashes verified, 1 invalid"
> > (This is true for working WACZs.)
> 
> Hmmm, interesting. I wonder if this is true just for these new WACZs or for other WACZs we produce... Maybe let's dig.

This is true for freshly-created Perma WACZs made with Scoop.

perma_web/perma/templates/provenance-summary.html
rebeccacremona: >  I don't think it's necessary to provide the level of provenance detail Scoop provides.

Agreed.

How about time of original capture (if any), time of replacement, replacement mime type?

perma_web/perma/utils.py
rebeccacremona: I don't know why I feel like we did this once before, or something like it, and it was exposed in the templates, and it was read by a LIL service or uptimerobot or something....

perma_web/perma/models.py
rebeccacremona: It is probably in the Linear ticket for studying and designing this. I'll try and Slack you a link.

perma_web/api/tests/test_link_authorization.py
```diff
@@ -167,11 +163,12 @@ def test_should_allow_user_to_patch_with_file(self):
                                   data={'file':file_content})
 
         self.assertTrue(Capture.objects.filter(link_id=self.link.pk, role='primary').exclude(pk=old_primary_capture.pk).exists())
+        self.assertTrue(Capture.objects.filter(link_id=self.link.pk, role='provenance_summary').exclude(pk=old_primary_capture.pk).exists())
```
rebeccacremona: Thanks for adding this! I think you don't need the `.exclude(pk=old_primary_capture.pk` in this one; that won't be caught by this query anyway.

perma_web/perma/models.py
```diff
@@ -1990,22 +1989,44 @@ def write_uploaded_file(self, uploaded_file, cache_break=False):
...
+            content_type=mime_type,
+            url=warc_url
+        )
+
+        provenance_capture = Capture(
+            link=self,
+            role='provenance_summary',
+            status='success',
+            record_type='resource',
+            user_upload='True',
```
rebeccacremona: Ooh, this is a fun question: should `user_upload` be true on the provenance summary, since we are creating it? I can see it going either way.

I also observe that, in the primary capture, we have long been passing in the string `'True'` and not a python boolean `True`. This is a Boolean field. I... wonder what we have in the DB... I will check... Hopefully it is converting to a Boolean, but either way, oops....

rebeccacremona: Okay locally, looks like these are converted to Booleans:
```
In [4]: for capture in Capture.objects.filter(user_upload=True):
   ...:     print(type(capture.user_upload))
   ...: 
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>

In [5]: for capture in Capture.objects.filter(user_upload='True'):
   ...:     print(type(capture.user_upload))
   ...: 
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
<class 'bool'>
```

That is... not a thing I would have expected to happen implicitly.

In any case, how about we switch to passing in `True`?

perma_web/perma/utils.py
bensteinberg: Here's how we did it before: https://github.com/harvard-lil/perma/commit/7e0a04b0403e9d44b95ce411fd6e6b86216099d1

perma_web/perma/templates/provenance-summary.html
bensteinberg: See [63ca218](https://github.com/harvard-lil/perma/pull/3679/commits/63ca21885e5dbb5ca8aee6b232a230baedaa55c9) -- 

image

perma_web/perma/models.py
bensteinberg: I was wondering about that setting for the provenance; I'll switch that one to `False` and change the other from string to boolean.

bensteinberg: Done in [44191e7](https://github.com/harvard-lil/perma/pull/3679/commits/44191e7719a4947ed8d1d473474e45e87959ea98)

perma_web/perma/utils.py
bensteinberg: Now see [e7d9405](https://github.com/harvard-lil/perma/pull/3679/commits/e7d94053df8da3d1c5b1004ee49c066179f3d19d) -- I haven't yet set up the mechanism for interpolation in deployment.

perma_web/perma/templates/provenance-summary.html
rebeccacremona: Looks good to me!

rebeccacremona (approved):

perma_web/perma/utils.py
bensteinberg: I now have a mechanism for interpolation in deployment; I don't love it, but it works, and I may revisit it later.

PR #3680, opened 2025-01-02
Minimal upgrade to Postgres 16

bensteinberg:
Tests pass locally.

See the PG release notes for versions [13](https://www.postgresql.org/docs/13/release-13.html), [14](https://www.postgresql.org/docs/14/release.html), [15](https://www.postgresql.org/docs/15/release.html), and [16](https://www.postgresql.org/docs/16/release.html).

The RDS upgrade path from 12.19, which is what we are running now, does not extend directly to 16.6, so I took the liberty of upgrading our stage instance from 12.19 to 12.22, and then to 16.6; we can kick the tires on stage. The upgrades took some minutes; we'll go into maintenance mode when it's time to upgrade the production instances.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -140,10 +140,9 @@ services:
 
 volumes:
   node_modules:
-  postgres_data:
   postgres_data_16:
   redis_data:
-  pp_db_data:
+  pp_db_data_16:
   scoop_redis_data:
   scoop_postgres_data:
   scoop_access_key:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,13 +11,13 @@ services:
     volumes:
       - redis_data:/data:delegated
   db:
-    image: registry.lil.tools/library/postgres:12.11
+    image: registry.lil.tools/library/postgres:16.6
     environment:
       POSTGRES_PASSWORD: password
       POSTGRES_DB: perma
       POSTGRES_USER: perma
     volumes:
-      - postgres_data:/var/lib/postgresql/data:delegated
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -141,6 +141,7 @@ services:
 volumes:
   node_modules:
   postgres_data:
+  postgres_data_16:
```
rebeccacremona, line 143: I like the idea of updating the volume's name! Less confusing for folks pulling down the changes.

What's the thought re: keeping both `postgres_data` and `postgres_data_16` defined here?

rebeccacremona (approved):
LGTM! Interesting to read the changelog and see the kinds of things that they are changing.

When merged, one of us should probably post instructions in Slack for what devs should do, when they pull down the changes, to initialize a new DB, and some remark about keeping or deleting old volumes.

docker-compose.yml
bensteinberg: I had a vague notion about devs possibly wanting to keep and use the old volume, as in 
> some remark about keeping or deleting old volumes

but I think I'll remove it from the docker compose file.

rebeccacremona: 👍 Yeah. That's what I figured. I always have a few extras around, for instance, one or two prod-like dumps put together [this way](https://github.com/harvard-lil/perma/blob/develop/developer.md#import-production-like-data-into-the-database), so I'm used to having volumes available that aren't listed in the compose file. 🤷‍♀️. No opinion on whether that's good or bad practice.

PR #3681, opened 2025-01-07
Fix broken Firefox extension

cmsetzer:
Before:
* The [Perma.cc extension](https://addons.mozilla.org/en-US/firefox/addon/perma-cc/) for Firefox was broken. It would fail on every URL with an error like `Firefox can’t find the file at moz-extension://5f641aec-6742-443d-b1c3-feeefc2e0419/https://perma.cc/service/bookmarklet-create/?v=1&url=https://example.com/`.
* The error happened because we pass the (absolute) target URL into `getURL`, which is only meant for relative paths (see [[1]](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/extension/getURL), [[2]](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getURL)). Seems like it only worked by accident before, or maybe they changed the API at some point.

After:
* We no longer use `getURL`, so the extension now connects to Perma.cc as expected.

## How to test

1. In Firefox, go to `about:debugging`
2. Click "This Firefox" at left
3. Click the "Load Temporary Add-on…" button and select the `extras/perma-firefox/src/manifest.json` file from this branch
4. Now the local extension should be loaded; navigate to a website and click the Perma.cc extension button to test it out

### DIFF
--- a/extras/perma-firefox/src/bg.js
+++ b/extras/perma-firefox/src/bg.js
@@ -1,7 +1,5 @@
-chrome.browserAction.onClicked.addListener(function(tab){
-  chrome.tabs.create({
-    "url": chrome.extension.getURL('https://perma.cc/service/bookmarklet-create/?v=1&url=' + encodeURIComponent(tab.url))
+browser.browserAction.onClicked.addListener(function(tab){
+  browser.tabs.create({
+    "url": "https://perma.cc/service/bookmarklet-create/?v=1&url=" + encodeURIComponent(tab.url)
   });
 });
-
-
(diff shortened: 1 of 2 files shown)


teovin:
Looks good! I followed your instructions and was able to capture a couple of links as expected.

This was my first time playing with this extension, so I would like to confirm the flow quickly: When the extension is clicked, the user is asked to log in. After logging in, they are taken to manage/create, where they are expected to use the create perma link button with the URL prepopulated from the previous step.

teovin (approved):

cmsetzer:
Yes, that is the expected behavior. The extension is new to me too, but it's basically a quick way to populate Perma's URL field from the current site. It was created in 2016 and hasn't been touched since, so it's pretty simple.

Thanks for the review! I will look at getting this release published on [addons.mozilla.org](https://addons.mozilla.org).

PR #3694, opened 2025-02-05
Update payments image

bensteinberg:
This updates the payments image with the same point version update to Django as #3693 -- see https://docs.djangoproject.com/en/5.1/releases/4.2.19/ and https://github.com/harvard-lil/perma-payments/pull/186. 

Note that although the tests for that PR succeeded, but the build on merge failed to push the image to the registry, so the build status of the payments repo is "failed" -- I built and pushed the image locally.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -91,7 +91,7 @@ services:
       - perma_payments
 
   perma-payments:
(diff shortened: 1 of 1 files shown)


bensteinberg:
(This is not the most efficient way to do this kind of update; I think I should have updated the payments repo, merged that PR, waited for the image to be built, then made a single PR to this repo, including both the package upgrade and the image upgrade.)

teovin (approved):

PR #3709, opened 2025-04-03
queue bulk org user emails with celery

teovin:
This PR adds the ability to the bulk user addition flow so that the user emails are queued as individual Celery tasks instead of sending them inline. I tested with a form of 500 users and it was instantaneous.

### DIFF
--- a/perma_web/perma/celery_tasks.py
+++ b/perma_web/perma/celery_tasks.py
@@ -1616,11 +1616,12 @@ def send_user_email_from_bulk_addition(user_email, context, template, host=None,
     """
     if is_new_user:
         user = LinkUser.objects.get(raw_email=user_email)
-        activation_route = f'''{host}{reverse('password_reset_confirm', 
-                                              args=[
-                                                urlsafe_base64_encode(force_bytes(user.pk)),
-                                                default_token_generator.make_token(user),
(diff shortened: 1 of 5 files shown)


perma_web/perma/celery_tasks.py
```diff
@@ -1603,3 +1607,22 @@ def warn_expiring_organization_users(warning_days=None):
...
+    """
+    Handles sending emails to users created/updated via the bulk org user addition form
+    """
+    if is_new_user:
+        user = LinkUser.objects.get(raw_email=user_email)
+        activation_route = f'''{host}{reverse('password_reset_confirm', 
+                                              args=[
+                                                urlsafe_base64_encode(force_bytes(user.pk)),
+                                                default_token_generator.make_token(user),
+                                              ])}'''
```
cmsetzer: Just a suggestion, but I think this will be more legible if you generate the URL outside the f-string:
```suggestion
        path = reverse(
            'password_reset_confirm',
            args=[
                urlsafe_base64_encode(force_bytes(user.pk)),
                default_token_generator.make_token(user),
            ],
        )
        activation_route = f'{host}{path}'
```

cmsetzer (approved):

PR #3710, opened 2025-04-03
Require auth for GUI-mediated downloads.

rebeccacremona:
Only logged in users can download using `?type=warc_download` or `?type=wacz_download` URLs.


bensteinberg (approved):
👍

PR #3711, opened 2025-04-11
Modify sponsorship expiration

teovin:
This PR adds the ability to revise a sponsored user's affiliation expiration date through the manage sponsored user view. 

Admin users can see all sponsorships, while registrar users can only see the sponsorships that are affiliated with their own registrar.

Here is the before and after of how the `manage/sponsored-users/<user_id>` page will look:

**Before:**

image

**After:**

Screenshot 2025-04-16 at 8 34 53 AM

Screenshot 2025-04-16 at 8 35 02 AM

Notice that I added a new button next to the `Deactivate sponsorship` button. They both have the same initial orange bordered styling, but the `Update sponsorship` button turns green on hover and focus events while the deactivate button turns red. 

Another change I made in this view is that previously, we weren't displaying affiliation information for permanent affiliations, and with this change, I added a permanent affiliation text. I thought this would make it clearer to the admins/registrars that permanent affiliations are also modifiable. I also removed the expiration info from inactive sponsorships (which I am fine with adding back; I thought this made more sense).

**Here is how the form to update the date looks like:**

Screenshot 2025-04-11 at 9 42 25 AM

If the user checks the `Sponsor permanently` box, the date selection input will be disabled. Only when the user de-selects the checkbox will the date selection be re-enabled:

image

Just a note that I haven't run this by Clare yet; I am planning to when she returns. So there might be changes to wording, styling, and overall behavior.

### DIFF
--- a/perma_web/perma/templates/user_management/manage_single_user.html
+++ b/perma_web/perma/templates/user_management/manage_single_user.html
@@ -12,12 +12,20 @@
           <div class="settings-block">
             <p>{{ sponsorship.registrar.name }}</p>
             {% if sponsorship.status == 'active' %}
+              <!-- <br><p><strong>Expiration date:</strong> {{ sponsorship.expires_at }}</p> -->
+              <br>
+              <p>
+                <strong>Expiration:</strong>
+                {% if sponsorship.expires_at %}
+                  {{ sponsorship.expires_at }}
+                {% else %}
+                  Permanent affiliation
+                {% endif %}
+              </p>
               <a href="{% url 'user_management_manage_single_sponsored_user_remove' user_id=target_user.id registrar_id=sponsorship.registrar.id %}" class="btn cancel btn-xs leave-org-btn">Deactivate sponsorship</a>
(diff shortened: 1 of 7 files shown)


rebeccacremona (approved):
Huzzah.

LGTM! I personally like that you removed the expiration date from inactive sponsorships; agreed that looks cleaner. One suggestion for wording: it might be clearer to say, "Permanent affiliation" rather than, "Expiration: Permanent affiliation" but you mentioned you were going to talk to Clare about labels anyway 🙂.

How would you feel about adding a couple tests?

teovin:
Thanks Becky, yep, will check with Clare as well and update labels.

And tests, I guess I can 🤣

teovin:
@rebeccacremona, regarding your comment about the expiration label for the permanent affiliations, I removed that and updated the screenshots in the PR description to reflect that. I will go ahead and close this PR for now, and if there is any additional feedback on the labels, etc, from the rest of the team, I can open another PR then.

PR #3712, opened 2025-04-23
Modify organization affiliation expiration

teovin:
This PR adds the ability to revise an organization user's affiliation expiration date through the manage organization users view.

Admin users can see all affiliations, while registrar users can see user affiliations linked to the same org they are associated with.

Here is the before and after of how the **manage/organization-users/<user_id>** page will look:

**Before:**

Screenshot 2025-04-23 at 11 53 06 AM

**After:**

Screenshot 2025-04-23 at 11 51 41 AM

Notice the new button next to the Remove button. 

Another change I made in this view is that previously, we weren't displaying affiliation information for permanent affiliations, and with this change, I added a permanent affiliation text. I thought this would make it clearer to the admins/registrars that permanent affiliations are also modifiable. 

Here is how the form to update the date looks like:

Screenshot 2025-04-23 at 11 54 05 AM

If the user checks the Set permanent affiliation box, the date selection input will be disabled. Only when the user deselects the checkbox will the date selection be re-enabled:

image

I also made a small tweak to the Remove affiliation flow. Previously, the action redirected to `manage/organization-users` view, but I changed it to redirect to `manage/organization-users/<user_id>` as it seemed more intuitive to me. Let me know if I missed a nuance there.

### DIFF
--- a/perma_web/perma/templates/user_management/manage_single_user.html
+++ b/perma_web/perma/templates/user_management/manage_single_user.html
@@ -31,20 +31,25 @@
     </div>
   {% elif this_page == 'users_organization_users' and affiliations %}
     <div class="orgs-list-container">
-      <h3 class="body-bh">Remove from organization</h3>
+      <h3 class="body-bh">Edit organization affiliation</h3>
       <p class="page-dek"><strong>{{ target_user.raw_email }}</strong> is a member of the following organization{{ affiliations|pluralize }}:</p>
-      <form action="{% url 'user_management_manage_single_organization_user_remove' user_id=target_user.id %}" method="POST">
-        {% csrf_token %}
-        {% for aff in affiliations %}
-          <div class="settings-block">
-            <p>{{ aff.organization.name }}</p>
+      {% for aff in affiliations %}
+        <div class="settings-block org-user-actions">
+          <p>{{ aff.organization.name }}</p><br>
+          <p>
+          {% if aff.expires_at %}
+            Expiration: {{ aff.expires_at }}
+          {% else %}
+            Permanent affiliation
+          {% endif %}
+          </p>
+          <form action="{% url 'user_management_manage_single_organization_user_remove' user_id=target_user.id %}" method="POST">
+            {% csrf_token %}
             <button name="affiliation" value="{{ aff.id }}" class="btn btn-default btn-xs leave-org-btn" data-toggle="tooltip">Remove</button>
-            {% if aff.expires_at %}
-              <br><p>Expiration date: {{ aff.expires_at }}</p>
-            {% endif %}
-          </div>
-        {% endfor %}
-      </form>
+          </form>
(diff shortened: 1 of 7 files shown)


perma_web/perma/tests/test_views_user_management.py
```diff
@@ -979,6 +979,38 @@ def test_can_remove_self_from_organization(self):
...
+                         data={'expires_at': ''},
+                         success_url=reverse('user_management_manage_single_organization_user', args=[self.organization_user.id]))
+        affiliation.refresh_from_db()
+        self.assertEqual(affiliation.expires_at, None)
+
+    def test_registrar_user_can_modify_affiliation_of_existing_org_user(self):
+        # can only modify user affiliations if registrar is affiliated with the same org
+        self.log_in_user(self.registrar_user)
+        affiliation = UserOrganizationAffiliation.objects.get(user=self.organization_user, organization=self.organization)
+        expires_at = '2025-04-30T00:00:00+00:00'
```
cmsetzer, line 1000: Just to make sure I understand, is the difference in datetime format between lines 987 and 1000 because this one is being posted via the form, and the previous one was committed directly via the ORM? Or is it just an irrelevant detail?

cmsetzer (approved):
Looks good to me!

perma_web/perma/templates/user_management/manage_single_user.html
```diff
@@ -31,20 +31,25 @@ <h4 class="body-ch">{{ status.grouper|title }}</h4>
...
+          <p>
+          {% if aff.expires_at %}
+            Expiration: {{ aff.expires_at }}
+          {% else %}
+            Permanent affiliation
+          {% endif %}
+          </p>
+          <form action="{% url 'user_management_manage_single_organization_user_remove' user_id=target_user.id %}" method="POST">
+            {% csrf_token %}
             <button name="affiliation" value="{{ aff.id }}" class="btn btn-default btn-xs leave-org-btn" data-toggle="tooltip">Remove</button>
```
cmsetzer, line 48: Also, should we make the Remove button red (I think it's `btn-danger` instead of `btn-default`) to visually differentiate the two?

perma_web/perma/tests/test_views_user_management.py
teovin: Yep, on line 987 I am manually setting an expiration date using a format that doesn't match the format the form saves. And later on line 991 I am emptying it before the form submits to check whether it will match what the test is expecting (None). 

I will match the format on line 1000, just for consistency's sake, even though it won't make a difference in the test outcome.

perma_web/perma/templates/user_management/manage_single_user.html
teovin: I agree! I was a little undecided since that button is an already existing element, and changing it might have unexpected UI/UX implications. But I am consulting with Clare and will update it depending on her feedback.

teovin: Clare prefers the buttons to stay blue, so I am leaving them as they are.

PR #3716, opened 2025-05-06
Don't save Scoop logs if capture was successful; assorted refactoring.

rebeccacremona:
This PR's main goal is to switch things up, so that we only save the Scoop logs to the DB if the capture attempt failed. So, it makes that change, and adds an `invoke` task we can run to delete the logs of successful captures, going back in time. 

The task aims to be medium efficient: it loads the IDs of all the capture jobs made by Scoop, excluding those that failed, into RAM and zeroes out the `scoop_logs` field in batches of 500, using `UPDATE`. 
- It filters that way rather than specifically looking for capture jobs whose `scoop_logs` field is populated because there isn't an index on that field. And there are indices for job status and for capture engine... though, I note, not a composite index on both. Hopefully the query isn't so inefficient it causes a problem: I think worst case scenario, it will be slow, but not expensive... and since we are only doing this once, hopefully no big deal. 
- It loads all the IDs into a list in RAM because... I think that's fine, we have plenty of RAM, and even with the number of IDs we're talking here... I'm not worried.
- It goes with batches of 500 because... that sounds large enough to be significant, but not crazy? Dunno, could be for our particular DB a different batch size would be most efficient, but, I think that's a good starting place. If it looks like we are struggling in some way, we could quit and try again; no harm in running this multiple times.

This PR also adds a new `creation_timestamp` field for capture jobs going forward: if it turns out this table is still too larger and/or we want to shrink it further, we can use that field to delete the scoop jobs that age out, according to whatever schedule we want.

All the other changes are just things I got up to "while I was in there".

- I added tags for a couple other categories of known Scoop errors

- I removed an unneeded bit of code from the `can_play_back` method on Link, because I was thinking of changing that logic, which looks, among other things, at a link's capture job... while thinking through whether we can consider deleting the capture jobs on some reasonable schedule entirely... which I decided in the end was out of scope for this PR.

- I split the massive unwieldy invoke `dev.py` into a few separate files, with possibly lousy names, to make it easier to see which tasks belong together, which tasks are around primarily for documentation of the past, and which tasks are commonly used.

- I quieted down some logging, so that we can more easily see what more important logs look like, when they go by in Grafana. (Right now, the logs are drowned in routine playback and capture announcements.)

To be discussed: whether we want to just give this a try, and see if it's good enough for now, or whether we think a more thorough intervention will be required.


perma_web/tasks/merge_accounts.py
```diff
@@ -0,0 +1,554 @@
+"""
+Perma's email addresses used to be case-sensitive, so users could actually
+have multiple accounts, with the same address capitalized differently.
+We merged accounts using the below tasks, and made login etc. case-insenstivie.
```
bensteinberg:
```suggestion
We merged accounts using the below tasks, and made login etc. case-insensitive.
```

bensteinberg (approved):
Looks great. I found one typo :)

PR #3720, opened 2025-05-08
disable selecting past dates

teovin:
This PR prevents the users from selecting past dates in the organization user, sponsored user, and bulk org user addition forms' affiliation expiration field. 

**Example form with disabled dates:**

image

### DIFF
--- a/perma_web/perma/templates/user_management/add_multiple_users_to_org.html
+++ b/perma_web/perma/templates/user_management/add_multiple_users_to_org.html
@@ -17,18 +17,22 @@
     </form>
 
     <script>
-      const checkbox = document.getElementById("id_a-indefinite_affiliation");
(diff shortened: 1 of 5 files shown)


cmsetzer (approved):
Looks good to me.

Optional thought: Did you consider whether to add the validation logic (i.e. require an expiration date of today or later) to the form/model field validation as well? I'm not sure if it's worth doing or not; might just add needless overhead.

teovin:
> Looks good to me.
> 
> Optional thought: Did you consider whether to add the validation logic (i.e. require an expiration date of today or later) to the form/model field validation as well? I'm not sure if it's worth doing or not; might just add needless overhead.

Yep, this "nipping in the bud" approach sounded simpler to me.

PR #3723, opened 2025-06-13
feat: added copy link button for a newly created permalink

jadidbourbaki:
## Summary

This PR adds a better UX for copying perma links.

## Motivation

This is a solution to the following issue: https://github.com/harvard-lil/perma/issues/3687

### Setup

I tested this with Docker Desktop 4.41.2 on a Macbook with Apple Silicon. I set up a local instance of perma following the provided [developer guide](https://github.com/harvard-lil/perma/blob/develop/install.md).

### UX Testing

- I navigated to https://perma.test:8000 and created a test account. 
- I created a permalink for https://example.com
- The copy button is visible ✅ 

image

- The tooltip is correct ✅ 

image

- Copying works ✅ 

## Roll out and Monitoring

- If the reviewers agree with the approach here, I can also send follow-up PRs to add similar copy buttons for the list links page.

- I can commit to providing support for and/or rolling back this feature for the next two weeks in case there are any initial breakages.

Issue #3687, opened 2025-01-18
Add "Copy Perma Link" Functionality


linuxlurak:
Hi
It would be very useful if we had more direct copy functions for the perma links. 

For example instead of (we have to select and copy the link):

<img src="https://github.com/user-attachments/assets/38d0c872-feb3-427d-b15b-7a3a621aacd0" height=140px>

Something like:
<img src="https://github.com/user-attachments/assets/bf04fc0a-eb09-4969-8dbb-e73becee52f5" height=135px>

Or even simpler:

<img src="https://github.com/user-attachments/assets/49b823ee-5ae0-4650-93aa-bcdb19bfdf1f" height=150px>

rebeccacremona:
Deployed!

PR #3726, opened 2025-06-24
feat: added copy link button to link lists

jadidbourbaki:
## Summary

This PR adds a better UX for copying perma links on the list links page.

This is a follow up PR to https://github.com/harvard-lil/perma/pull/3723

## Motivation

This is a solution to the following issue: https://github.com/harvard-lil/perma/issues/3687

### Setup

I tested this with Docker Desktop 4.41.2 on a Macbook with Apple Silicon. I set up a local instance of perma following the provided [developer guide](https://github.com/harvard-lil/perma/blob/develop/install.md).

### UX Testing

- I navigated to https://perma.test:8000 and created a test account. 
- I created a permalink for https://example.com
- I created a second permalist for https://google.com
- The copy button is visible ✅ 

image

- The tooltip is correct ✅ 

image

- Copying works ✅ 

Also confirmed the original functionality is unaffected:

image

I notice that the CSS duplication in `style-responsive-archive.scss` and `style-responsive.scss` for the copy button is necessary; otherwise, the newly created link styling disappears. Happy to put this in any global scss / CSS file, if there is one, to avoid duplication. 

## Roll out and Monitoring

- I can commit to providing support for and/or rolling back this feature for the next two weeks in case there are any initial breakages.


rebeccacremona (approved):
Thanks @jadidbourbaki, this looks great! I think lots of people will enjoy this feature. We're planning a production deployment soon, so you should see it in action before long.

The duplication is fine; as I'm sure you noticed, our frontend is in a bit of a transition state presently. We can fold that into the list of things that will get cleaned up as we migrate more fully to clean components :-)

PR #3728, opened 2025-07-08
Add intro cursor rule

jcushman:
From our Cursor demo just now, here's an automatically generated rules file so Cursor will have background on the codebase when editing a particular file. I don't know if this file is accurate or even useful for getting better responses, though I suspect it's both accurate-enough and useful-enough -- just sharing for fun!

As Mike suggested I used the [new rules format](https://docs.cursor.com/context/rules) rather than .cursorrules.

### DIFF
--- a/.cursor/rules/code-base-intro.mdc
+++ b/.cursor/rules/code-base-intro.mdc
@@ -1,0 +1,234 @@
+---
+description: 
+globs: 
+alwaysApply: true
+---
+# Perma.cc Codebase Context
+
+## Project Overview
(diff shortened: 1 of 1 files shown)


rebeccacremona (commented):
> I don't know if this file is accurate or even useful for getting better responses, though I suspect it's both accurate-enough and useful-enough -- just sharing for fun!

I had plans to give this a test drive and see, but, a month later, have not had a chance. 

So! I figure let's just merge it. If someone is wrenching on Perma and thinks it's unhelpful, in the future, they can delete or change it; that's probably better than leaving this PR open indefinitely.

PR #3731, opened 2025-08-07
Allow admin to deny pending registrar without registrar user

cmsetzer:
Our team has requested the ability to deny a pending registrar (e.g., one submitted by a bot) without having to first populate a registrar user. This branch implements the logic and also does some cleanup on the conditions for whether to show/hide the Approve and Deny buttons.

To see this in action, run Perma locally and:

1. Submit a paid registrar request via the [Other Orgs sign up form](https://perma.test:8000/sign-up/firms) without setting yourself as administrator on the account
2. View the pending registrar request at `/manage/registrars/approve/xx`, replacing `xx` with the ID for your submitted registrar

## Before

image

## After

image

### DIFF
--- a/perma_web/perma/forms.py
+++ b/perma_web/perma/forms.py
@@ -139,6 +139,15 @@ class ApproveRegistrarForm(ModelForm):
             self.fields['base_rate'].required = False
             self.fields['status'].required = False
 
+    def clean_status(self) -> str | None:
+        status = self.cleaned_data.get('status')
+        registrar_user = self.cleaned_data.get('registrar_user')
+
+        if status == 'approved' and not registrar_user:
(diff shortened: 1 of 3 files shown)


rebeccacremona (approved):
Great! Looks good to me!

Do you think to be belt and suspenders, it's worth adding a line that complains if `new_status == 'approved'` and `not target_registrar_user`? Maybe not worth it: I know we can't get into that condition in the normal flow. Only asking because I noticed that, should that somehow happen, we'd run the "deny" branch.

cmsetzer:
I decided to address this at the `ApproveRegistrarForm` level in a77bd48. This seems like the right spot, since it's a data validation issue but probably shouldn't involve changing the model.

PR #3732, opened 2025-08-21
raise validation error for non utf-8 encoded csv files

teovin:
With this change, instead of throwing a 500 error, the form will raise a validation error if the CSV file isn't encoded with UTF-8 standard. 

Screenshot 2025-08-21 at 10 11 03 AM

### DIFF
--- a/perma_web/perma/tests/test_views_user_management.py
+++ b/perma_web/perma/tests/test_views_user_management.py
@@ -694,8 +694,8 @@ class UserManagementViewsTestCase(PermaTestCase):
                          ).exists())
 
     def test_add_multiple_org_users_via_csv(self):
-        def create_csv_file(filename, content):
-            return SimpleUploadedFile(filename, content.encode('utf-8'), content_type='text/csv')
+        def create_csv_file(filename, content, encoding):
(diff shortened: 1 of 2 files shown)


perma_web/perma/tests/test_views_user_management.py
```diff
@@ -712,6 +712,7 @@ def initialize_form(csv_file, data=None):
         one_more_valid_csv_file = create_csv_file('one_more_valid_users.csv', csv_data)
         invalid_csv_file = create_csv_file('invalid_users.csv', invalid_csv_data)
         another_invalid_csv_file = create_csv_file('another_invalid_users.csv', another_invalid_csv_data)
+        one_more_invalid_csv_file = create_csv_file('users.csv', csv_data, 'utf-16')
```
cmsetzer: It's minor, but maybe we could name this more descriptively since it's a different kind of invalidity? e.g.

```suggestion
        csv_file_with_invalid_encoding = create_csv_file('users.csv', csv_data, 'utf-16')
```

cmsetzer (approved):
Looks good!

Issue #2529, opened 2019-01-08
Immutable dict error when creating link batch


rebeccacremona:
What happened here? This was via a command line API call... perhaps extra data was POSTed, e.g., the 'created_by' field? Test when you have time....
```
File "./api/views.py" in post
  559.             request.data['created_by'] = request.user.pk
```
stacktrace
```
Internal Server Error: /v1/archives/batches

AttributeError at /v1/archives/batches
This QueryDict instance is immutable

Request Method: POST
Request URL: https://api.perma.cc/v1/archives/batches
Django Version: 1.11.16
Python Executable: /usr/bin/uwsgi-core
Python Version: 3.5.3
Server time: Tue, 8 Jan 2019 14:25:26 +0000


Traceback:  

File "/usr/local/share/perma/perma_web/.venv/lib/python3.5/site-packages/django/core/handlers/exception.py" in inner
  41.             response = get_response(request)

File "/usr/local/share/perma/perma_web/.venv/lib/python3.5/site-packages/django/core/handlers/base.py" in _get_response
  187.                 response = self.process_exception_by_middleware(e, request)

File "/usr/local/share/perma/perma_web/.venv/lib/python3.5/site-packages/django/core/handlers/base.py" in _get_response
  185.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/usr/local/share/perma/perma_web/.venv/lib/python3.5/site-packages/django/views/decorators/csrf.py" in wrapped_view
  58.         return view_func(*args, **kwargs)

File "/usr/local/share/perma/perma_web/.venv/lib/python3.5/site-packages/django/views/generic/base.py" in view
  68.             return self.dispatch(request, *args, **kwargs)

File "/usr/local/share/perma/perma_web/.venv/lib/python3.5/site-packages/rest_framework/views.py" in dispatch
  483.             response = self.handle_exception(exc)

File "/usr/local/share/perma/perma_web/.venv/lib/python3.5/site-packages/rest_framework/views.py" in handle_exception
  443.             self.raise_uncaught_exception(exc)

File "/usr/local/share/perma/perma_web/.venv/lib/python3.5/site-packages/rest_framework/views.py" in dispatch
  480.             response = handler(request, *args, **kwargs)

File "./api/views.py" in post
  559.             request.data['created_by'] = request.user.pk

File "/usr/local/share/perma/perma_web/.venv/lib/python3.5/site-packages/django/http/request.py" in __setitem__
[...]

PR #3738, opened 2025-09-25
Workaround for immutable dict error

rebeccacremona:
It's the weirdest thing. https://github.com/harvard-lil/perma/blob/develop/perma_web/api/serializers.py#L301 is throwing with:
> Exception Type: AttributeError at /v1/archives/
> Exception Value: This QueryDict instance is immutable

... but only for one particular api user (whom we have not heard from), under unknown circumstances I cannot seem to recreate, no matter what I try. Beats me. I would really like to know the root cause, but ????

This PR hopefully works around the issue, by setting `validation_status_code` directly on the serializer, instead of trying to inject it into the request data. That is a cleaner strategy anyway: I should have done that in the first place.

### DIFF
--- a/perma_web/api/serializers.py
+++ b/perma_web/api/serializers.py
@@ -298,7 +298,7 @@ class AuthenticatedLinkSerializer(LinkSerializer):
                             if content_length > settings.MAX_ARCHIVE_FILE_SIZE:
                                 errors['url'] = f"Target page is too large (max size {settings.MAX_ARCHIVE_FILE_SIZE / 1024 / 1024}MB)."
                         if validation_status_code := response_data.get("status_code"):
-                            self.context['request'].data["validation_status_code"] = validation_status_code
+                            self.validation_status_code = validation_status_code
                     except ScoopAPIException:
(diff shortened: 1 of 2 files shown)


cmsetzer (approved):
Looks good!

rebeccacremona:
Update: @cmsetzer figured out how to trigger this situation! 

Just like in issue [2529](https://github.com/harvard-lil/perma/issues/2529) (which has followup in [3132](https://github.com/harvard-lil/perma/issues/3132)), it's that, when you send data as 'multipart/form-data', Django creates a QuertyDict, which is immutable... whereas sending as 'application/json' creates a regular dict.

e.g. 
`curl -H "Authorization: ApiKey <redacted>" -X POST -d "url=http://example.com" https://api.perma.test:8000/v1/archives/` 
vs 
`curl -H "Authorization: ApiKey <redacted>" -H 'Content-Type: application/json' -X POST -d '{"url": "http://example.com"}' https://api.perma.test:8000/v1/archives/`

Hooray, mystery solved!

PR #3742, opened 2025-11-06
Use intended form label

rebeccacremona:
See LIL-3778.

The `name` field was supposed to be labeled "Organization name", but that code wasn't working as intended: it turns out, you can only override the label for model fields using that syntax, not set the label for extra fields. So, this PR defines the label inline, when defining the field, instead.

Before:
image

After:
image


cmsetzer (approved):
Ugh. Thanks for fixing this!

PR #3743, opened 2025-11-12
Add expiration date field to sponsored user and org user exports

rebeccacremona:
See LIL-4437.

Users with sufficient credentials can download a spreadsheet of metadata about the users affiliated with the groups they administer: org users, and/or sponsored users, depending on their particular permissions.

This PR adds a column to those spreadsheets, reporting when the users' affiliations expire (which is a "newer" Perma feature, in the scheme of things: all affiliations were originally permanent).

The column is called "affiliation_expires_at" for org users, and "sponsorship_expires_at" for sponsored users. The new column appears in both CSV and JSON export formats.

image
image
image

I didn't think it was worth updating the tests to check for the presence of that column. If you think I should, I am happy to readdress 🙂

### DIFF
--- a/perma_web/perma/views/user_management.py
+++ b/perma_web/perma/views/user_management.py
@@ -322,11 +322,13 @@ def manage_sponsored_user_export_user_list(request: HttpRequest) -> HttpResponse
         'last_login',
         'sponsorship_status',
         'sponsorship_created_at',
+        'sponsorship_expires_at'
     ]
     records = list_sponsored_users(request, export=True)
     users = records.annotate(
         sponsorship_status=F('sponsorships__status'),
         sponsorship_created_at=F('sponsorships__created_at'),
+        sponsorship_expires_at=F('sponsorships__expires_at'),
     ).values(*field_names)
     filename = 'perma-sponsored-users'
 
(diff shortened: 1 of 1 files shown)


teovin (approved):
LGTM!

As a side note, I think we should add the sponsoring registrar's name to the export as well. Without that column, all we see are a bunch of date columns. And especially considering one user might have multiple sponsorships, it doesn't look very meaningful to me as it is.

Let me know what you think, and I can create a Linear ticket for that work if needed.

rebeccacremona:
> LGTM!
> 
> As a side note, I think we should add the sponsoring registrar's name to the export as well. Without that column, all we see are a bunch of date columns. And especially considering one user might have multiple sponsorships, it doesn't look very meaningful to me as it is.
> 
> Let me know what you think, and I can create a Linear ticket for that work.

Sure, that makes sense to me for admins. 

Registrar users, the folks who would normally be doing this, would of course only see the sponsorships for their own registrar, and since you can only be a registrar user for one registrar, it would be unambiguous.

But totally agreed that the data is not that useful for admin users, without a column like that! Let's throw it in Linear! Could be, we should rethink whether admins should be able to do this at all.

PR #3744, opened 2025-11-17
Prevent long court name from erroring out during sign up

rebeccacremona:
See LIL-3723.

These signup forms are (and ever have been) so quirky: you can fill them out with any email address, and there is special logic for, if you happen to use the email address of an existing user. We have long talked about completely re-envisioning this flow.

But, in the meantime, we were seeing an occasional bug: if a user filled out the form with the email of an existing user, and supplied a court name > 45, we tried to save the overly long value to the DB and got an error. (Form validation took care of it otherwise, if a new email address was used.)

This PR takes care of that in two ways. 

First, it tweaks the logic in `[user_sign_up.py](https://github.com/harvard-lil/perma/compare/develop...rebeccacremona:prevent-long-name?expand=1#diff-d97eadd77df375d1af215119a37863572f3205e05eedf8c8fe186f6f9aaf4f90)` to validate the form, but ignore the error that complains there is already a user with that email (we know). That is sufficient to get a validation error to the user:
image

But, I also explicitly added a max_length to the char field definition. That results in the form being rendered with `<input type="text" name="requested_account_note" maxlength="45" required="" id="id_requested_account_note">`, so, the user's browser will enforce the limit.

### DIFF
--- a/perma_web/perma/forms.py
+++ b/perma_web/perma/forms.py
@@ -344,7 +344,11 @@ class CreateUserFormWithCourt(UserForm):
     add court to the create user form
     """
 
-    requested_account_note = forms.CharField(required=True)
+    requested_account_note = forms.CharField(
+        required=True,
+        # Explicitly add this max length to the form, so that it enforced by the HTML
+        max_length=LinkUser.requested_account_note.field.max_length
+    )
 
     class Meta:
         model = LinkUser
--- a/perma_web/perma/forms.py
+++ b/perma_web/perma/forms.py
@@ -346,7 +346,7 @@ class CreateUserFormWithCourt(UserForm):
 
     requested_account_note = forms.CharField(
         required=True,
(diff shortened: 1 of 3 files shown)


perma_web/perma/forms.py
```diff
@@ -346,7 +344,11 @@ class CreateUserFormWithCourt(UserForm):
     add court to the create user form
     """
 
-    requested_account_note = forms.CharField(required=True)
+    requested_account_note = forms.CharField(
+        required=True,
+        # Explicitly add this max length to the form, so that it enforced by the HTML
```
cmsetzer:
```suggestion
        # Explicitly add this max length to the form, so that it is enforced by the HTML
```

cmsetzer (approved):
Looks good. These `sign_up_*` functions are so unwieldy! Thanks for figuring out a comprehensive fix to this.

PR #3745, opened 2025-11-25
update action labels

teovin:
See LIL-4436.

This is to update the action labels on the manage users views to account for affiliation expiration modifications.


rebeccacremona (approved):
LGTM!

PR #3748, opened 2025-12-03
Public API tweaks

rebeccacremona:
This PR removes the `/public/archives/:guid` and `/public/archives/:guid/download` routes. 

It accordingly removes the `warc_download` and `wacz_download` URLs from the public link serializer and the public option from the helper function for calculating download URLs.

It reinstates the `/public/archives/` route, but imposes a 10-page limit on pagination. So, only the most recent 10 * 300 = 3,000 Perma Links are exposed.

(Here's a screenshot from my local, with different settings, showing what it looks like when you exceed the limit.)

image

It updates the docs to remove references to those two routes, and to mention the pagination limit, adding a link to the contact form, with an auto-filled subject line, for folks to get in touch if they are looking for more info for research purposes.

image
image


rebeccacremona:
@cmsetzer if you have a minute, I'd love a quick second set of eyes on that pagination code, just to confirm you don't think it's crazy 😄

perma_web/perma/templates/docs/developer/index.html
```diff
@@ -71,37 +71,18 @@ <h3 id="examples-in-curl" class="body-bh">Examples in cURL</h3>
...
 
-    <p class="body-text">(Note: these API routes are temporarily unavailable.)</p>
-
-    <p class="body-text">The Public Archives API endpoint helps you find all publicly available Perma.cc archives. You do not need an API Key to use this portion of the API.</p>
+    <p class="body-text">The Public Archives API endpoint helps you find publicly available Perma.cc archives. You do not need an API Key to use this portion of the API.</p>
     <p class="body-text">The base endpoint for this API is <code>{{base_url}}/public/archives/</code></p>
 
+    <p class="body-text">Only recently-made Perma Links are listed.</p>
+
+    <p>To inquire about a more complete dataset, please <a href="{% url 'contact' %}?subject=Public%20Perma%20Link%20Dataset%20Inquiry">please drop us a note</a>.</p>
```
cmsetzer: This language might be too subtle? If I were a user reading these docs, I might not parse this as "this API has a hard limit on the offset parameter."

I might suggest an edit like

```suggestion
    <p class="body-text">This API provides access to the most recently created Perma Links. To inquire about access to data beyond this limit, <a href="{% url 'contact' %}?subject=Public%20Perma%20Link%20Dataset%20Inquiry">please drop us a note</a>.</p>
```

perma_web/perma/settings/deployments/settings_common.py
```diff
@@ -556,6 +556,7 @@ def filter(self, record):
     'SEARCH_PARAM': 'q',  # query string key for plain text searches
     'ORDERING_PARAM': 'order_by',  # query string key to specify ordering of results
 }
+API_MAX_PAGES = 10
```
cmsetzer, line 559: Just to confirm: this isn't part of `REST_FRAMEWORK` because that only contains actual DRF config settings, right?

cmsetzer (approved):
This looks good. I would suggest we do `API_MAX_OFFSET` instead of max pages; it feels like doing the pagination math adds complexity that isn't needed. Your call!

perma_web/perma/templates/docs/developer/index.html
rebeccacremona: Thanks, I like that suggestion a lot.

perma_web/perma/settings/deployments/settings_common.py
rebeccacremona: Right. I thought about putting it inside, but wasn't positive it would reliably propagate, settings objects can be weird...

rebeccacremona:
> This looks good. I would suggest we do `API_MAX_OFFSET` instead of max pages; it feels like doing the pagination math adds complexity that isn't needed. Your call!

Thanks! Yeah, I hear you... I think I'll stick with pages for now, since this is what Jack proposed, but if we end up doing something like this for _all_ the API routes, after testing to see what we can, in fact, handle, I like the idea of switching over then.

PR #3749, opened 2025-12-04
Disable the public list route again, temporarily.

rebeccacremona:
This is here in case we need to revert, after deploying my latest.


cmsetzer (approved):

rebeccacremona:
Nice, we didn't need it.

PR #3750, opened 2025-12-17
Add a switch to disable live capturing.

rebeccacremona:
See LIL 4596.

This PR adds a Django Waffle "[switch](https://waffle.readthedocs.io/en/stable/types/switch.html)" that can be toggled on in the Django admin, in order to prevent the Perma application from continuing to submit capture requests to the Scoop API. 

The switch is intended to give users a better experience on the rare occasion that we have an extended capturing outage: users will see a message if they try to make a link, rather than a slow failure, and will be offered the opportunity to upload.

image

image


teovin:
LGTM! The switch is working as expected, disabling link creation when enabled and reenabling it when disabled.

teovin (approved):

PR #3752, opened 2026-02-03
Limit allowed page length on API requests

rebeccacremona:
See LIL-4773.

This fixes a bug in our understanding of DRF's `PAGE_SIZE` setting: it does not enforce that size as a maximum; rather, it defaults to it.

In this PR, I tweak `get_limit` to ensure that if `PAGE_SIZE` is set, it is enforced as the maximum page size.

I realized that this would actually mess up a quick fix we once made to the frontend: https://github.com/harvard-lil/perma/pull/3179

We intend to completely revisit this UI, so, I made another "quick fix": request up to two pages of our max length (300 x 2 = 600) and then print an error to the console if it looks like there is more. A full-fledged pagination implementation here would, I think, want to be lazy-loading... and I don't think that's worth it at this time. So, hopefully this will last until the redesign :-)


teovin:
When `limit` query passed exceeds max allowed page size, the `max_limit` is enforced.

image

When the request is below the `max_limit`, the requested size is returned.

image

When not passed, it defaults to 300 as expected.

image

The folder pagination looks good as well with the caveat that we don't have a parent folder with over 600 folders right now.

teovin (approved):

PR #3756, opened 2026-03-16
add daily link counts endpoint

teovin:
This new endpoint (/v1/internal/daily_link_counts/?lookback_period=int) returns the daily link counts for a given period in the following response format. The reason behind the creation of this is to prevent the status app from calling the archives endpoint repeatedly by paginating through multiple pages, looking for link creation dates. 

```
{
    "lookback_period": 30,
    "counts": [
        {
            "2026-03-14": 10
        },
        {
            "2026-03-16": 3
        }
    ]
}
```

### DIFF
--- a/perma_web/api/serializers.py
+++ b/perma_web/api/serializers.py
@@ -361,3 +361,10 @@ class LinkBatchSerializer(BaseSerializer):
 
 class DetailedLinkBatchSerializer(LinkBatchSerializer):
     target_folder = FolderSerializer(read_only=True)
+
+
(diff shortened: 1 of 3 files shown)


rebeccacremona (approved):
Looks great to me!

One thought: what would you think of adding a row for "today so far"?

teovin:
> Looks great to me!
> 
> One thought: what would you think of adding a row for "today so far"?

Hmm, I like that idea. I'll save that for another iteration, if/when we need that info, though. For now, this should be sufficient for the status app.

PR #3757, opened 2026-03-17
Alter public api route

rebeccacremona:
This PR removes the `public_archives` route from the list of available routes:
image

It is still available, it just isn't advertised via that mechanism, which may reduce the number of crawlers that hit it.

This also drops support for filtering or searching within the list of public archives; that has grown too expensive to support, over time.


rebeccacremona:
~(I may still have to update the tests)~

teovin (approved):

PR #3759, opened 2026-03-23
Adopt uv in place of pip-tools, setup.cfg

cmsetzer:
We've successfully adopted uv to manage Python dependencies on other projects. This branch swaps uv in for [pip-tools](https://pip-tools.readthedocs.io/), keeping dependencies pinned to equivalent versions from `requirements.txt` via uv's recommendation for [migrating version constraints](https://docs.astral.sh/uv/guides/migration/pip-to-project/#importing-requirements-files).

## Before

* Dependencies are listed in `requirements.in` and `requirements.txt`
* No structural distinction between production and dev dependencies
* Tool config is defined in separate `setup.cfg` file
* `pip install` from scratch takes ~30s

## After

* Dependencies are listed in `pyproject.toml` with `uv.lock` as lockfile
* Dev dependencies are listed separately from prod dependencies
* Tool config is defined in `pyproject.toml`
* Pruned outdated [mock](https://pypi.org/project/mock/) dependency in favor of `unittest.mock`
* `uv sync` from scratch takes ~9s


cmsetzer:
@rebeccacremona notes that, if and when this is approved, we should hold off on merging until we know whether prod can successfully deploy a Python project without `requirements.txt`.

perma_web/Dockerfile
```diff
@@ -90,4 +96,5 @@ RUN  apt-get update && apt install -y --no-install-recommends libdbus-glib-1-2 \
 
 RUN mkdir -p $HOME/.pki/nssdb \
     && certutil -d $HOME/.pki/nssdb -N --empty-password \
-    && playwright install chromium firefox
+    && playwright install chromium firefox \
+    && ln -sf "$(find /root/.cache/ms-playwright -name 'chrome' -type f -executable | head -1)" /usr/local/bin/playwright-chromium
```
rebeccacremona, line 100: Nice.

perma_web/pyproject.toml
```diff
@@ -0,0 +1,97 @@
...
+    "django-tree-queries>=0.19.0",
+    "django-waffle>=4.1.0",
+    "django-webpack-loader~=1.0",
+    "djangorestframework>=3.15.2",
+    "internetarchive>=3.0.0",
+    "invoke>=2.0.0",
+    "ipdb>=0.13.13",
+    "mock>=5.1.0",
+    "netaddr>=0.8.0",
+    "pip-tools>=7.4.1",
```
teovin: We don't need this anymore, right?

.github/workflows/lints.yml
```diff
@@ -9,8 +9,10 @@ jobs:
...
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
+
       - name: Flake8
         working-directory: perma_web
         run: |
-          pip install `egrep -o 'flake8==\S+' requirements.txt`  # install our version of flake8
-          flake8
+          uv run --frozen --only-group dev flake8
```
teovin, line 18: It looks like the way we run `flake8` in our local env will need to use a similar pattern to the CI. In the previous setup, I used to run the `flake8` in the web container bash. Now it returns multiple violations and doesn't respect the config in `pyproject.toml`, but if I run `uv run flake8` in perma_web root, it returns the expected violations. We might want to update developer.md to reflect this.

teovin:
I was able to create a fresh instance, install dependencies, run tests and lints, and navigate the UI and admin portal. Looks good to me overall. I only noted a couple of things and will approve after they are addressed.

perma_web/pyproject.toml
cmsetzer: Good catch! Removed in [b5864e7](https://github.com/harvard-lil/perma/pull/3759/commits/b5864e7d85bee65b39241b91ce836a2faf6c4cbb).

.github/workflows/lints.yml
cmsetzer: Oh, thanks for spotting this. I think this is flake8 on the container not respecting the settings in `pyproject.toml` (unless executed via `uv`). It would be good for us to consider swapping flake8 out for Ruff (which can re-use our existing flake8 config), but that should be a separate PR.

Anyway, I updated the language in [7bfd3b8](https://github.com/harvard-lil/perma/pull/3759/commits/7bfd3b8909f271c76079c57566c4eed0dbce2ce8) to account for the flake8 change. Thanks!

cmsetzer:
@teovin Thanks for the review; I've updated the issues you mentioned.

@rebeccacremona Do you mind double-checking my "lock dependencies + commit lockfile" logic in [64fa296](https://github.com/harvard-lil/perma/pull/3759/commits/64fa296af8497d0e87670b6567709c012aa65ff5)?

teovin (approved):

rebeccacremona (approved):
LGTM! I never know what to make of it when the hashes for particular packages change; I imagine all these are fine...

The code in the action for updating requirements.txt seems fine to me!

PR #3760, opened 2026-03-24
add a db models visualization route

teovin:
This PR installs a tool that visualizes the database models in a chart-like UI, which includes field/column information and their data types, and the relationships between models. It adds two routes to the app:

`/schema-viewer/`: This route renders the chart itself, allowing a few customization options such as app filtering and visuals.
`/schema-viewer/schema/`: This route displays the raw JSON data of the chart. It includes the primary and foreign keys information, and also the exact reference points between models. 

A screenshot from the chart (not scrolling):

image


perma_web/perma/settings/deployments/settings_dev.py
```diff
@@ -109,3 +109,17 @@
...
+
+if VIEW_SCHEMA:
+    try:
+        import schema_viewer
+        INSTALLED_APPS += (
+            'schema_viewer',
+        )
+    except ImportError:
+        VIEW_SCHEMA = False
+        print("django-schema-viewer is not installed")
```
cmsetzer, line 125: I'd probably opt to just raise the `ImportError` here, since an engineer's dev setup should match `requirements.txt` and it's one less edge case to worry about. Your call, though!

cmsetzer (approved):
Looks good! Thanks for adding this.

rebeccacremona:
Before merging: I think this setting must be defined in settings_common.py, to avoid problems in production... but I will look for real after this meeting!

teovin:
> Before merging: I think this setting must be defined in settings_common.py, to avoid problems in production... but I will look for real after this meeting!



> Before merging: I think this setting must be defined in settings_common.py, to avoid problems in production... but I will look for real after this meeting!

How about I just add it as False to be safe? Defensive driving :)

rebeccacremona:
> How about I just add it as False to be safe? Defensive driving :)

That sounds right to me :-)

Here's why: right now, VIEW_SCHEMA is only defined in `settings_dev.py`, and then overwritten in `settings_testing.py`, which inherits from `settings_dev.py`.

In production, the inheritance chain is `settings_common.py` -> `settings_production.py` -> `settings.py` (which is placed on the servers by Saltstack, with machine-specific settings, etc.). So, in production, at import time, when `urls.py` is executed... and `if settings.VIEW_SCHEMA:` runs... no such setting will be found and, unless I am mistaken, things will crash. 

The same pattern, for why `TESTING` is set as False in `settings_common.py` https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/settings/deployments/settings_common.py#L593, and then just updated to `True` in `settings_testing.py`.

We are all agreed this settings inheritance pattern isn't immediately intuitive. But, I think the documentable TLDR would be, "Whenever adding a new setting, always add something to settings_common.py, and then customize per tier or installation as needed."

rebeccacremona (changes requested):
(Per non-review discussion, let's add the new setting with a default of False to settings_common.py)

teovin:
> (Per non-review discussion, let's add the new setting with a default of False to settings_common.py)

Done!

rebeccacremona (approved):

PR #3761, opened 2026-03-24
Speed up Python tests

cmsetzer:
This branch speeds up the Python tests by about 2.7x (from >5m to <2m). It does this like so:

1. We now perform per-test storage cleanup only for those tests that actually write to storage; otherwise we skip it
2. We now use a faster password hashing algorithm (MD5 instead of PBKDF2) and exclude unnecessary middleware when testing
3. We now run tests in parallel via pytest-xdist

The latter change is the most complex, since our tests weren't written with concurrency in mind. I've needed to:

* Separate or lock shared resources such as S3 buckets, certificates, and live server ports to prevent conflicts between workers
* Add pytest hooks/fixtures to ensure JSON data is reloaded after a transactional test flushes the database
* Handle other miscellaneous edge cases with bits of extra logic

I'm curious about whether you think the extra overhead to achieve parallelization is an acceptable tradeoff for saving another ~90s in test runtime. I feel like it's worth it, but these tests are pretty baroque as it is…


perma_web/setup.cfg
```diff
@@ -13,9 +13,11 @@ exclude = node_modules,static,settings
 DJANGO_SETTINGS_MODULE = perma.settings.deployments.settings_testing
 norecursedirs = node_modules
 python_files = tests.py test_*.py *_tests.py
-addopts = --browser chromium --screenshot only-on-failure --output failed_test_files
+addopts = -n auto --dist loadscope --browser chromium --screenshot only-on-failure --output failed_test_files
 filterwarnings =
     ignore:Unverified HTTPS request
+markers =
+    uses_storage: marks tests that write to S3/MinIO storage (cleanup_storage teardown is skipped for unmarked tests)
```
rebeccacremona, line 20: I really like the idea of this marker.

docker-compose.yml
```diff
@@ -51,7 +51,7 @@ services:
     # https://github.com/elgalu/docker-selenium/issues/20
     shm_size: 2G
     environment:
-      - DJANGO_LIVE_TEST_SERVER_ADDRESS=0.0.0.0:8000
+      - DJANGO_LIVE_TEST_SERVER_ADDRESS=0.0.0.0
```
rebeccacremona, line 54: Okay, cool: I think the only reason we had this pinned to :8000, the same as the dev server runs on, was so that you could in theory interact with the test server yourself in your browser (since that port is exposed to the Docker host). I don't even remember why we wanted or needed to do that... Seems like a totally reasonable tradeoff for more speed/parallelization!

perma_web/perma/settings/deployments/settings_testing.py
```diff
@@ -38,6 +45,37 @@
...
 CELERY_BROKER_URL = 'memory://localhost/'
 
+# Use faster (but weaker) password hasher when testing
+PASSWORD_HASHERS = [
+    "django.contrib.auth.hashers.MD5PasswordHasher",  # Faster than PBKDF2 for setting passwords
+    "django.contrib.auth.hashers.PBKDF2PasswordHasher",  # Fallback for fixture verification
+]
+
+# Comment out middleware that isn't needed for testing
+MIDDLEWARE = (
```
rebeccacremona: Instead of duplicating `MIDDLEWARE` here, and then commenting out ones that we don't need, how would you feel about switching the definition of `MIDDLEWARE` in `settings_common.py` from a tuple to a list, and then using `MIDDLEWARE.remove('name-of-middleware')` here? 

I am worried that we will not remember to keep this up-to-date if we alter our middleware definition in `settings_common.py`.

rebeccacremona (approved):
Cool! This seems great to me, especially the explicit marker to indicate that storage is used.

The added complexity in locking for cert installation seems fine.

Further complexity re: the JSON fixtures is unfortunate, but is readable, and... hopefully... quite temporary! This is additional motivation for us to continue refactoring the test suite to STOP using those entirely. Similarly, it's always been unclear to me why many of the "transactional" tests need to be that way... With any luck, as we continue to modernize and improve the tests, we will eventually be able to get ride of this entirely.

I had one note about `settings_testing.py`; I'm interested to hear if you agree.

Hooray for faster tests.

perma_web/perma/settings/deployments/settings_testing.py
cmsetzer: Good call — I think that's a lot clearer and less verbose than the commenting out. I'll make the change.

PR #3763, opened 2026-04-07
cleanup user management templates' and views' auth checks

teovin:
This PR cleans up some of the user management templates and their repeated `request.user` checks by creating a context processor that's populated with ui flags. The flags include actions and which users can carry them out, which are then received by the templates, and used to decide what content to render. 

I created an access module to define the access conditions. Along with the above, I also used this to power the `@user_passes_test_or_403` checks in the `user_management` view in an attempt to reduce the repetition of the `user` checks there as well. (I see the `@user_passes_test_or_403(lambda user:...` is used outside of user management view, and can address those in a separate PR)

**EDIT:** Becky's commits make architectural changes to this PR. Now, instead of a context processor, we have a `user_management` module, and template permissions are constructed at render time per screen/group. Thanks @rebeccacremona!


rebeccacremona (commented):
Thanks for giving this a whack; these templates are just wild lol.

The templates are much easier to read with those named booleans, and the reduction of the dozens of inline lambdas in the permission checks is delightfully DRY.

I don't, however, think a context processor is the way to go: context processors run on every HTML request (homepage, settings, playback, etc.), just to build the ~20 flags needed by the user-management pages. I think we can go with a narrower helper, explicitly called by the views that need it. I think that will also be less confusing than having the default context build automatically, and then overriding it in places with calls like `context['user_management_perms'] = build_user_management_perms(request, screen=group_name)`.

I'm going to take a swing at tweaking the architecture of your code, for your consideration: we can looks at the advantages and disadvantages of both together and pick whichever aspects seem cleanest and most intuitive 🤝

rebeccacremona:
Alrighty! I made a PR to your PR: https://github.com/teovin/perma/pull/2

When you're back, let's chat about whether you like these ideas, or whether I messed up something about your original structure!

Cheers,

rebeccacremona (approved):
Let's do it! I enjoyed working on this one together!

PR #3766, opened 2026-04-09
Tweak how links are filtered when listed by the API

rebeccacremona:
We recently noticed some odd SQL: queries with a redundant WHERE filter:
```
WHERE (NOT "perma_link"."user_deleted" AND "perma_link_folders"."folder_id" IN (x, y, z) AND T4."folder_id" = x)
```
The above is checking to see if a link is in a set  of folders... and then checking to see if the link is in one particular folder.

That's silly.

This PR tweaks the code, so that we do one or the other:
- filter for links in the one folder, if a folder is specified
- filter for all the links in the accessible set, otherwise

Removing the redundancy will hopefully improve performance somewhat for users who have access to large trees of folders.


durkinm2 (approved):
Looks good!

PR #3767, opened 2026-04-09
Faster Docker builds and CI

cmsetzer:
This branch adds some optimizations to speed up our CI process, especially Docker builds:

* Cache uv setup across linting job runs
* Cache Docker layers across testing runs via buildx-cache
* Reorder Dockerfile to adopt [recommended caching practice](https://docs.docker.com/build/cache/): heavy, infrequently changed layers (build dependencies, Playwright binaries) should come before lighter, more frequently changed ones (JS and Python dependencies)
* Tweak Python dependency constraints in `pyproject.toml` to use a [compatibility constraint](https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release) (`djangorestframework~=3.15.2`) instead of an open-ended one (`djangorestframework>=3.15.2`). This will help prevent surprising incompatibilities from cropping up by accident


durkinm2 (approved):
Great work with these caching improvements, Chris. LGTM overall!

PR #3768, opened 2026-04-20
separate models file into smaller modules

teovin:
This PR separates the models.py file into smaller modules that are organized around similar-ish resources, and also removes unused imports.

Before, the models file included 22 classes. After, they are organized as follows:

**base**

- class DeletableManager(models.Manager)
- class DeletableModel(models.Model)
- class GenericStringTaggedItem(CommonGenericTaggedItemBase, TaggedItemBase)
- class CustomerModel(models.Model)

**registrar**

- class RegistrarQuerySet(QuerySet)
- class Registrar(CustomerModel)
- class Sponsorship(models.Model)

**organization**

- class OrganizationQuerySet(QuerySet)
- class Organization(DeletableModel)

**user**

- class LinkUserManager(BaseUserManager)
- class LinkUser(CustomerModel, AbstractBaseUser, PermissionsMixin)
- class UserOrganizationAffiliation(models.Model)
- class ApiKey(models.Model)

**folder**

- class FolderQuerySet(TreeQuerySet)
- class Folder(TreeNode)

**link**

- class LinkQuerySet(QuerySet)
- class Link(DeletableModel)
- class Capture(models.Model)
- class CaptureJob(models.Model)
- class LinkBatch(models.Model)

**internet_archive**

- class InternetArchiveItem(models.Model)
- class InternetArchiveFile(models.Model)


perma_web/perma/models/__init__.py
```diff
@@ -0,0 +1,55 @@
+"""
+Django models for the perma app, implemented in submodules (base.py, link.py, etc.).
+
+Application code can keep using imports such as 'from perma.models import Link'.
```
cmsetzer, line 4: This is a good choice for now. I think we should update all those imports in a future PR so this file doesn't need to know about them.

perma_web/perma/models/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    'Organization',
+    'Registrar',
+    'Sponsorship',
+    'UserOrganizationAffiliation',
+    'get_default_archive_formats',
+    'get_empty_datetime_range',
+    'link_count_in_time_period',
+    'most_active_org_in_time_period',
+    'subscription_is_active',
+]
```
cmsetzer: Do we need this? I don't think we do `from .models import *` anywhere and it just repeats the imports from above. I'd vote for dropping the `__all__`.

cmsetzer (approved):
Looks good! Everything seems to run fine on my end, and the smaller modules are much easier to digest.

I think we should remove the redundant `__all__` from `__init__.py` but otherwise this seems good to merge.

perma_web/perma/models/__init__.py
teovin: You are right. It was mostly an api documentation hint, but our models are pretty lean already, and we don't really need it.

PR #3769, opened 2026-04-22
update docs

teovin:
This PR merges the `developer.md` and `install.md` files and makes a few corrections to wording, URLs, and commands.


developer.md
```diff
@@ -297,19 +280,18 @@ Python unit tests live in `perma/tests`, `api/tests`, etc.
...
 
-Javascript tests live in `spec/`.
+JavaScript tests live in `spec/`.
 
-See the [Common tasks and commands](#common-tasks-and-commands) for the
-common techniques for running the tests.
+See [Common tasks and commands](#common-tasks-and-commands) for commands to run the tests.
 
 
 ### Linting with flake8
```
cmsetzer: We should update this heading too if we're normalizing to "Flake8" elsewhere:

```suggestion
### Linting with Flake8
```

cmsetzer (approved):
Thanks for the editing! This looks like a step forward. A couple other thoughts:

* Can we merge `install.md` and `developer.md` into one doc? I get confused about which one to read for instructions when I have to rebuild Perma.

* While we're cleaning up the docs, can we standardize the Markdown code blocks? These docs currently use a mix of:
  ````md
  ## Single backticks on own line:

  `d pytest`

  ## Indent with CLI prompt:

      $ docker compose up -d

  ## Fenced code blocks:

  ```sh
  d invoke lock --args "--upgrade-package package_name"
  ```
  ````
  My opinion: we should always use fenced blocks for non-inline code examples. What do you think?

teovin:
@cmsetzer I merged the two mds and also standardized the non-inline code blocks and Flake8. Let me know if there is anything else, otherwise, I will merge this.

cmsetzer:
@teovin Looks good to me, thanks for working on this!

PR #3772, opened 2026-05-04
separate api views into smaller modules

teovin:
This PR separates the api/views.py file into smaller modules that are organized around similar-ish resources, and also removes unused imports.

Before, the views file included 20 classes. After, they are organized as follows:

**base**

- class BaseView(APIView)

**batches**

- class LinkBatchesListView(BaseView)
- class LinkBatchesDetailView(BaseView)
- class LinkBatchesDetailExportView(LinkBatchesDetailView)

**capture_jobs**

- class CaptureJobListView(BaseView)
- class CaptureJobDetailView(BaseView)

**folders**

- class FolderListView(BaseView)
- class FolderDetailView(BaseView)

**links**

- class LinkFilter(django_filters.rest_framework.FilterSet)
- class PublicLinkListView(BaseView)
- class AuthenticatedLinkListView(BaseView)
- class AuthenticatedLinkListExportView(BaseView)
- class AuthenticatedLinkDetailView(BaseView)
- class AuthenticatedLinkDownloadView(BaseView)
- class MoveLinkView(BaseView)

**misc**

- class DeveloperDocsView(APIView)
- class InternalDailyLinkCountsView(APIView)

**organizations**

- class OrganizationListView(BaseView)
- class OrganizationDetailView(BaseView)

**users**

- class LinkUserView(BaseView)


perma_web/api/views/link_batches.py
```diff
@@ -0,0 +1,132 @@
+import csv
```
rebeccacremona, line 1: can we call this file `link_batches.py`, so that it more closely matches the corresponding model?

perma_web/api/resources/batches.py
```diff
@@ -0,0 +1,132 @@
...
+from rest_framework.response import Response
+
+from perma.models import CaptureJob, LinkBatch
+
+from ..serializers import DetailedLinkBatchSerializer, LinkBatchSerializer
+from ..utils import dispatch_multiple_requests, reverse_api_view_relative
+from .base import BaseView
+
+
+# /batches
```
rebeccacremona: I realize this is just cut-and-pasted, but the comment that hints this is served at `/batches` is incorrect, it's actually at `/archives/batches`.

I am ambivalent about these comments: I see why they are there, especially for classes that handle two different styles of access (on with a parent, one without)... but on the other hand, it's redundant with info in urls.py and easy to get wrong....

But for the time being, how about we just update this one to say `/archives/batches`?

perma_web/api/resources/batches.py
```diff
@@ -0,0 +1,132 @@
...
+        }]
+        response = dispatch_multiple_requests(request, call_for_fresh_serializer_data)
+        data = response[0]['data'].copy()
+        links_remaining = request.user.get_links_remaining()
+        data['links_remaining'] = 'Infinity' if links_remaining[0] == float('inf') else links_remaining[0]
+        data['links_remaining_period'] = links_remaining[1]
+        return Response(data, status=status.HTTP_201_CREATED)
+
+
+# /batches/:id
```
rebeccacremona: Same here.

perma_web/api/views/link_batches.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def get(self, request, pk, format=None):
+        """
+        Single link batch details
+        """
+        return self.simple_get(request, pk)
+
+
+# /batches/:id/export
+class LinkBatchesDetailExportView(LinkBatchesDetailView):
```
rebeccacremona, line 104: Same here.

perma_web/api/urls.py
```diff
@@ -6,7 +6,20 @@
...
+    AuthenticatedLinkDetailView,
+    AuthenticatedLinkDownloadView,
+    AuthenticatedLinkListExportView,
+    AuthenticatedLinkListView,
+    MoveLinkView,
+    PublicLinkListView
+)
+from .resources.misc import DeveloperDocsView, InternalDailyLinkCountsView
+from .resources.organizations import OrganizationDetailView, OrganizationListView
+from .resources.users import LinkUserView
```
rebeccacremona: Two thoughts here:

One is, since in `resources/__init__.py` you import all the individual views, as with:
```
from .base import BaseView # noqa: F401
from .batches import ( # noqa: F401
    LinkBatchesListView, 
    LinkBatchesDetailView, 
    LinkBatchesDetailExportView
)
from .capture_jobs import ( # noqa: F401
    CaptureJobListView, 
    CaptureJobDetailView
)
...
```
I think it would be more correct to here, in urls.py, do in the imports straight from `resources`, rather than redundantly from `resources.users`, etc. I will pull this down locally and see if that works like I think it does.

Second, I wonder if you could share thoughts about calling this directory `views` instead of `resources`. `views` is what I am used to seeing in Django apps, and these are indeed [DRF views](https://www.django-rest-framework.org/api-guide/views/#class-based-views), that inherit from `rest_framework.views import APIView`. Do you think `resources` is more common or more standard?

If we called the directory `views`, and switched to import straight from the `views` package (from . import views), it would let us leave the urls.py file entirely unchanged.

But interested in hearing your thoughts on this topic.

rebeccacremona (commented):
Cool, I think splitting this up seems fine!!

I had a few thoughts/questions reading it through; I took the liberty of making a PR to your branch with the proposed changes so you could easily merge if you end up being on the same page :-) https://github.com/teovin/perma/pull/1. But if not, interested in hearing thoughts!

perma_web/api/urls.py
teovin: Hello! Thank you for staging the changes in the PR, everything looked reasonable to me, and I merged it. As for the resources/views convo, you are right that these are views in Django world, and I was thinking of "resources" as a way to describe the data they are exposing. I still need to remind myself of the diff between Django views and MVC app views occasionally :) Let me know if you spot anything else, otherwise I will go ahead and merge this.

rebeccacremona: > the diff between Django views and MVC app views 
Totally: I've never really bought the reasoning for the difference in nomenclature, etc.!

And great, let's do it!

rebeccacremona (approved):

PR #3773, opened 2026-05-08
add notes field to registrar model

teovin:
This PR adds a new text field - `notes` - to the registrar model to store notes. It also adds the new field to the admin change registrar view.

image


cmsetzer (approved):
Good idea!

PR #3774, opened 2026-05-13
Disable prorating

rebeccacremona:
Until now, we have always set new monthly subscriptions to start on the 1st of the month, charging a one-time prorated fee on initial setup for the days remaining in the current month.

We plan to switch that: full price on the day you sign up, and then your next payment in a month, whichever day that falls on.

This PR takes care of the Perma side of that switch: it cannot be deployed until [Perma Payments is adjusted ](https://github.com/harvard-lil/perma-payments/blob/997d5914be26c76be07d594eeccaa2df6a53cd0f/web/perma_payments/models.py#L224C1-L227C68)to handle that change.

I configured my local dev setup to communicate with the Cybersource test tier to click the button in Perma and watch a new monthly subscription get correctly created.

I was charged $10 today on the 13th, correctly.
image
image

A new subscription was created, with a "start" date of June 13th, with a scheduled payment on the correct date.
image
image

I'll keep the PR in draft mode until the corresponding Perma Payments change has been deployed: https://github.com/harvard-lil/perma-payments/pull/194

### DIFF
--- a/perma_web/perma/models/base.py
+++ b/perma_web/perma/models/base.py
@@ -14,7 +14,6 @@ from taggit.models import CommonGenericTaggedItemBase, TaggedItemBase
 
 from perma.exceptions import InvalidTransmissionException, PermaPaymentsCommunicationException
 from perma.utils import (
-    first_day_of_next_month,
     pp_date_from_post,
     prep_for_perma_payments,
     process_perma_payments_transmission,
--- a/perma_web/perma/models/base.py
+++ b/perma_web/perma/models/base.py
@@ -18,6 +18,7 @@ from perma.utils import (
     pp_date_from_post,
     prep_for_perma_payments,
     process_perma_payments_transmission,
+    today_next_month,
     today_next_year,
 )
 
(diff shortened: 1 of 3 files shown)


cmsetzer (approved):
Looks reasonable!

I was curious about how `today_next_month` might handle cases where a relative delta of +1 month results in a nonexistent date (e.g., 1/31 + 1 month = 2/31), but the `dateutil.relativedelta` docs [helpfully mention that](https://dateutil.readthedocs.io/en/stable/relativedelta.html):

> If the result falls on a day after the last one of the month, the last day of the month is used instead.

rebeccacremona:
> Looks reasonable!
> 
> I was curious about how `today_next_month` might handle cases where a relative delta of +1 month results in a nonexistent date (e.g., 1/31 + 1 month = 2/31), but the `dateutil.relativedelta` docs [helpfully mention that](https://dateutil.readthedocs.io/en/stable/relativedelta.html):
> 
> > If the result falls on a day after the last one of the month, the last day of the month is used instead.

Yeah! I love those utilities, I feel so well-taken-care-of

PR #3776, opened 2026-06-01
update language

teovin:
This PR updates the subscription language to support the disabling of prorating in Perma & Perma Payments.


rebeccacremona (approved):
Huzzah!

PR #3778, opened 2026-06-04
Admin stats page efficiency

jcushman:
The admin stats page polls /manage/stats/celery_queues every 2 seconds, which adds up to a real amount of traffic in the logs. For efficiency:

- Add start/stop buttons for celery queues to match the other stats sub-tabs
- Lazy load sub-tabs, so no data is loaded for unused tabs
- Disable polling when document is not visible to user (document.hidden)
- Disable polling when sub-tab not selected (so if you click the start button and switch sub-tabs, that tab doesn't poll until you switch back)

This way it not only doesn't start automatically, but if you do start and then move away, it and other sub-tabs don't keep pointlessly going in the background.

This PR is AI coded and not tested by me. In my defense: I discussed this with Becky and it sounds like this contribution is useful; it's a low risk staff-only page, easy to apply (if you have Perma up and running locally) and check if it works. Feel free to take it over, or if it's wrong in some complicated way, reject without looking too hard.

### DIFF
--- a/perma_web/perma/templates/admin-stats.html
+++ b/perma_web/perma/templates/admin-stats.html
@@ -19,6 +19,9 @@
 
       <div class="tab-pane active" id="celery_data">
         <h3 class="body-ah">Celery queues</h3>
+        <button type="button" id="toggle-queues-auto-refresh">Start Auto-Refresh (every 2s)</button>
+        <br>
+        <hr>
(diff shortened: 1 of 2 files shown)


rebeccacremona (approved):
I'm into it.

If it breaks the page, it breaks the page, and I'll fix it Monday.

PR #3779, opened 2026-06-05
Add capture_queue_health invoke task for use by salt

jcushman:
This adds an `invoke dev.capture-queue-health` command that checks if more than 25 of the last 50 capture jobs are uncompleted, or if 25 jobs have been stuck for more than 20 minutes. This is primarily for use by an automatic salt check, so with --quiet it only prints if one of those conditions applies. You can also run it manually without --quiet to see the counts.

* The first check is approximately the same as what status.perma.cc/monitor used to decide to print "PROBLEM_PENDING". It's a little more correct because that check was limited to info available through the public API endpoint.
* The second check might be a way to have fewer false positives: it won't trigger if 50 links happen to be dropped into the queue and processed right when this checks, only if jobs are sticking around longer than we want. This check wasn't possible with the /monitor approach. I'm not sure what the right values are for `max_stuck` and `stuck_minutes` though, so it's a fallback mode for now.

This is first of a 3-part PR series. Once it ships, we can update the salt crontab to use it, and then update status.perma.cc to drop the old route.

### DIFF
--- a/perma_web/tasks/dev.py
+++ b/perma_web/tasks/dev.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 from invoke import task
 import json
 import os
@@ -10,9 +10,10 @@ from tqdm import tqdm
 
 from django.conf import settings
 from django.http import HttpRequest
+from django.utils import timezone
 
 from perma.email import send_user_email, send_self_email, registrar_users, registrar_users_plus_stats
-from perma.models import Link, LinkUser, Registrar
+from perma.models import Link, LinkUser, Registrar, CaptureJob
 
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


perma_web/tasks/dev.py
```diff
@@ -93,6 +94,58 @@ def count_links_without_cached_playback_status(ctx):
...
+    )
+    uncompleted = Link.objects.filter(
+        pk__in=recent_link_ids,
+        captures__status='pending',
+        captures__user_upload=False,
+    ).distinct().count()
+
+    # 2. Stuck jobs: too old to still be unfinished. Age comes from
+    # Link.creation_timestamp (write-once); CaptureJob.creation_timestamp is
+    # auto_now and so tracks last save, not submission.
```
rebeccacremona, line 130: Whoops. I don't think was the intended behavior of that field. Good to know!

perma_web/tasks/dev.py
```diff
@@ -93,6 +94,58 @@ def count_links_without_cached_playback_status(ctx):
...
 
 
+@task
+def capture_queue_health(ctx, quiet=False, window=50, max_uncompleted=25, stuck_minutes=20, max_stuck=25):
+    """
+    Report on whether captures are completing, flagging two conditions:
+
+      1. Of the most recently created {window} links, how many still have a
+         capture (other than a user upload) in the 'pending' state. This is the
+         original status.perma.cc/monitor heuristic -- a snapshot of whether
```
rebeccacremona, line 104: > how many still have a capture (other than a user upload) in the 'pending' state
I look forward to dropping this legacy way of doing it... CaptureJob.status == `pending` is at least as informative as that value. I agree better to just swap in identical logic for now, and readdress later!!

rebeccacremona (approved):
LGTM!

I bet we can pull some data on how long it normally takes 25 jobs to finish, when the queue is saturated, to inform `stuck_minutes`. I'll make a note to do so: maybe in Linear!

perma_web/tasks/dev.py
jcushman: Looking closer, this coincidentally might not be biting us if we never actually save() a CaptureJob after creating it. A lot of the updates are like `CaptureJob.objects.filter(link__user_deleted=True, status='pending').update(status='deleted')`, which I think won't trigger `auto_now` since it's applied in Django rather than in a DB trigger.

PR #3781, opened 2026-06-18
Toggle between payment apps

rebeccacremona:
We intend to support a new version of Perma Payments in coming weeks.

This PR adds (admittedly hackish) support for configuring two payment app destinations, and toggling between them using a Django Waffle switch. 

Flip the switch in the Django admin to direct traffic to the "stripe" payments app instead of the default "cybersource" app.

image

-IF- `STRIPE_PAYMENTS_APP_INTERNAL_URL` and `STRIPE_PAYMENTS_APP_EXTERNAL_URL` are configured in settings, Perma will then direct traffic to the new app. If they are not configured, it will log a warning, and will continue directing traffic to the Cybersource app.

In staging and production, STRIPE_PAYMENTS_APP_INTERNAL_URL and STRIPE_PAYMENTS_APP_EXTERNAL_URL can be the same (though we could, if we wanted to serve public routes and private routes at differently.. though almost certainly not worth it....). But locally, `STRIPE_PAYMENTS_APP_INTERNAL_URL` needs to describe the URL the Perma `web` container can hit, to access the payments app (probably http://host.docker.internal, for now), and STRIPE_PAYMENTS_APP_EXTERNAL_URL needs to describe the URL the user's web browser can hit (probably http://localhost or, if you want to set up forwarding with /etc/hosts, perma-payments.test). 

For more info on testing locally, see LIL's internal docs on the Stripe transition: we are working on them as we speak!

This PR should be safe to deploy: I intend it to be a no-op. That is why I left the original "settings.*" in place... even though that is fairly awkward. This -should- continue to work, even if I failed to call the new helper `get_payments_app_url`, in some spots. The biggest risk is typos.


perma_web/perma/views/user_settings.py
```diff
@@ -152,8 +151,8 @@ def settings_usage_plan(request):
 
     context = {
         'this_page': 'settings_usage_plan',
-        'purchase_url': settings.PURCHASE_URL,
-        'subscribe_url': settings.SUBSCRIBE_URL,
+        'purchase_url': get_payments_app_url('purchase_history'),
```
teovin: I think we need `purchase` here instead of `purchase_history`.

perma_web/perma/settings/deployments/settings_common.py
```diff
@@ -254,6 +254,10 @@
 # Subscription packages/Payments
 #
 
+# No trailing slash
+STRIPE_PAYMENTS_APP_INTERNAL_URL = None
+STRIPE_PAYMENTS_APP_EXTERNAL_URL = None
```
teovin, line 259: Nitpick: Do you mind adding all of the Cybersource and Stripe env vars to `settings.example.py` for documentation purposes?

perma_web/perma/utils.py
```diff
@@ -399,6 +399,34 @@ def remove_whitespace(s):
 
 # communication
 
+def get_payments_app_url(route):
```
teovin, line 402: It's not a dealbreaker, but a test would be helpful for this.

perma_web/perma/views/user_settings.py
teovin: I pushed this change since you are out, just fyi.

teovin (approved):

teovin (approved):

PR #3783, opened 2026-06-24
remove blog and twitter links

teovin:
This PR removes the Twitter and blog links from the footer and the blog link from the anonymous user view header.


cmsetzer (approved):
Looks good to me!

PR #3784, opened 2026-06-25
rename paths and set up redirects

teovin:
This PR renames the `docs/libraries` and `sign-up/firms` paths and sets up redirects. It also updates the references to their old naming in the code base with the new naming.


cmsetzer (approved):
LGTM!

PR #3785, opened 2026-06-26
add type and international fields to the registrar model and admin form

teovin:
This PR adds two new fields to the registrar model; a multi-select `type` field and a boolean `international` field. It also adds these to the admin registrar change (admin/perma/registrar/id/change/) view and the registrar list (admin/perma/registrar/) views.

image


perma_web/perma/models/registrar.py
```diff
@@ -22,11 +23,32 @@ class Registrar(CustomerModel):
...
+        ('academic-library', 'Academic library'),
+        ('other-library', 'Other library'),
+        ('law-firm', 'Law firm'),
+        ('court', 'Court'),
+        ('public-defender', 'Public defender'),
+        ('attorney-general', 'Attorney general'),
+        ('government', 'Government'),
+        ('nonprofit', 'Nonprofit'),
+        ('factcheck', 'Factcheck'),
+        ('news/journalism', 'News/journalism'),
```
cmsetzer, line 37: Maybe this identifier should be slug-style, i.e. hyphen instead of slash?

```suggestion
        ('news-journalism', 'News/journalism'),
```

perma_web/perma/models/registrar.py
```diff
@@ -22,11 +23,32 @@ class Registrar(CustomerModel):
     """
     This is a library, a court, a firm, or similar.
     """
+    REGISTRAR_TYPE_CHOICES = (
+        ('law-library', 'Law library'),
+        ('academic-library', 'Academic library'),
+        ('other-library', 'Other library'),
```
cmsetzer, line 29: A thought: maybe the identifiers should be in type-subtype order:

```suggestion
        ('library-law', 'Law library'),
        ('library-academic', 'Academic library'),
        ('library-other', 'Other library'),
```

Then they'll still be collocated for sorting or aggregation.

cmsetzer (approved):
Looks good. I included a couple optional suggestions about the `type` identifiers; feel free to adopt or ignore.

perma_web/perma/models/registrar.py
teovin: I think I will leave this as is since this was the preference of the product team.

perma_web/perma/models/registrar.py
teovin: same as above :) Thank you for the review!

PR #3789, opened 2026-07-07
add payment system upgrade banner

teovin:
This PR adds a banner to the dashboard and usage-plan page announcing the payment system upgrade. 

Copy/dates to be updated before merging.

image

### DIFF
--- a/perma_web/perma/models/user.py
+++ b/perma_web/perma/models/user.py
@@ -409,6 +409,19 @@ class LinkUser(CustomerModel, AbstractBaseUser, PermissionsMixin):
         """
         return not self.nonpaying or (self.is_registrar_user() and not self.registrar.nonpaying)
 
+    ### banner visibility ###
+    
+    def should_see_payment_system_upgrade_banner(self):
+        """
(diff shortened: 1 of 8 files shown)


rebeccacremona (approved):
I like how you implemented this. 

I might recommend a little more padding under the text, but 🤷‍♀️ 

I might also recommend including it on the `usage` URLs too (https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/urls.py#L109C1-L112C131), including the fallback template (https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/templates/settings/settings-usage-plan-unavailable.html)... for ease, maybe all the settings pages? (That doesn't seem too disruptive...) But of course that's others' call!!!

LGTM, once dates and copy are updated.

teovin:
Thank you! Checked with Kristi, and she would like it on the `settings/usage-plan` page as well. I will leave it on these two pages for now. Can revisit later if need be.

image

teovin:
View after the mailchimp url addition and title update: 

image

PR #3791, opened 2026-07-14
Show status message on usage plan page after payment redirects

jcushman:
I think without this when you do something like edit a subscription, it just dumps you back on the plans page without confirming what changed -- does that sound right?


rebeccacremona:
Closed in favor of https://github.com/harvard-lil/perma/pull/3794

PR #3793, opened 2026-07-14
Allow self-serve cancellation through Stripe page

jcushman:
Per [LIL-5354: Funnel all payment and billing operations Stripe-side rather than Perma-side in Usage Plan page](https://linear.app/harvard-lil/issue/LIL-5354/funnel-all-payment-and-billing-operations-stripe-side-rather-than) --

**Usage Plan page**
- Removed the "Cancel Subscription" button and its form entirely.
- Renamed the blue "Modify Subscription" button to **"Manage Subscription and Billing"**.
- Updated the on-hold help text that referenced the old button name to match.

**Subscription update page** (`settings-subscription-update.html`):
- Renamed the "Update Credit Card Information" heading and button to **"Manage Payment and Billing"**.
- Added the descriptive line under the heading: *"Update payment information, view invoices and receipts, or cancel your subscription via Stripe's secure customer portal."*
- For registrars only, added under the "Modify ___'s Subscription" heading: *"Please contact info@perma.cc to make changes to your current subscription."*

Update tests to match.

PR #3792, opened 2026-07-14
Further warn about upcoming transition period; freeze transactions

rebeccacremona:
Both behind switches.

Description, tests TBD.

PR #3794, opened 2026-07-15
User plans page updates

jcushman:
This PR combines and replaces #3791 #3792 #3793:

### Switches

Make sure behavior is driven by three waffle switches:

| Switch | Role | Default |
|---|---|---|
| `use_stripe_payments_app` | Route payment endpoints + UI to the Stripe app | off |
| `allow_cybersource_transactions` | Allow *new* CyberSource transactions | on |
| `display_cybersource_freeze_message` | Show the migration-window banner | off |

`use_stripe_payments_app` takes precedence over `allow_cybersource_transactions` in the UI.

### Rollout phases

| Phase | use_stripe / allow_cyber / freeze_msg | UI |
|---|---|---|
| **Today (prod)** | off / on / off | Unchanged CyberSource UI (Modify + Cancel) |
| **Freeze (Jul 27–31)** | off / off / on | Freeze banner; new transactions disabled; existing subscriptions still reported/reconciled |
| **Stripe go-live** | on / off / off | "Manage Subscription and Billing" → Stripe customer portal |

With defaults, the rendered pages are today's UI.

### Code changes

- **Freeze switches (Becky's commits):** disable new subscribe/purchase/modify/cancel controls and 403 the update/cancel routes when CyberSource transactions are off, while leaving status reads, purchase history, and the sync task running.
- **Stripe billing UI:** when `use_stripe_payments_app` is on, replace the CyberSource Modify/Cancel controls with a single **Manage Subscription and Billing** button that opens the Stripe customer portal (card updates, invoices, and at-period-end cancellation all live there); relabel the subscription-update page's card section; registrars are directed to email info@perma.cc. Cancellation is portal-only in Stripe mode (no self-serve Cancel button); it propagates back via Stripe webhooks → local mirror. [LIL-5354: Funnel all payment and billing operations Stripe-side rather than Perma-side in Usage Plan page](https://linear.app/harvard-lil/issue/LIL-5354/funnel-all-payment-and-billing-operations-stripe-side-rather-than)
- **Post-redirect status messages:** success/canceled banners on the usage plan page after Stripe checkout (`?subscription=success`, etc.), gated on `use_stripe_payments_app` so they're inert on CyberSource.
- **Fail-loud misconfig guard:** `get_payments_app_url` now raises `ImproperlyConfigured` (instead of silently falling back to CyberSource) if `use_stripe_payments_app` is on but `STRIPE_PAYMENTS_APP_*_URL` is unset — so the "Stripe" UI can never post to CyberSource. Surfaces most visibly as a 500 on the usage plan page.
- **Hardening flag behavior:** Verify once CyberSource transactions are switched off, no Perma route can drive a CyberSource transaction:
  - update/cancel routes are POST-only (405 on GET) in every phase;
  - cancel route 403s whenever CyberSource transactions are off (freeze **and** Stripe);
  - update route 403s during the freeze; in Stripe mode it renders but its forms target **only** the Stripe app;
  - the usage plan page serves no CyberSource-targeted subscribe/purchase form when transactions are off.
- **Testing:** A bunch of testing to express the three phases as switch combinations and assert per-phase UI, route guards, backend targeting of rendered form actions, and the misconfig raise. All existing usage-plan tests pass unchanged.

### Go-live checklist (ops)

1. Configure `STRIPE_PAYMENTS_APP_INTERNAL_URL` / `EXTERNAL_URL` **first** (required — flipping the switch without them now 500s the usage plan page).
2. Turn `use_stripe_payments_app` on; turn `allow_cybersource_transactions` off; clear the freeze banner.
   Freeze window is the same two switches: `allow_cybersource_transactions` off + `display_cybersource_freeze_message` on, `use_stripe_payments_app` still off.

### Notes

- **Minor UX gap:** a cancel completed in the Stripe portal returns with `?update=success` → shows "Your payment information was updated" rather than a cancellation-specific message (portal return doesn't distinguish the action).


rebeccacremona (approved):
LGTM:
- I confirmed that my changes from #3792 read as intended
- Big fan of the tests
- I didn't see anything in the logic from #3791 or #3793 that looks unsafe to deploy to stage or to prod
- Logic from #3791 or #3793 looks feasible; I did not pull down to test.

Three notes: 

> Fail-loud misconfig guard: get_payments_app_url now raises ImproperlyConfigured (instead of silently falling back to CyberSource)... Surfaces most visibly as a 500 on the usage plan page.

A much louder/more visible way this surfaces [subscription_status](https://github.com/harvard-lil/perma/blob/f0e56e2a6b04d5456f1435dc1e72030801e51850/perma_web/perma/models/base.py#L535C2-L548C20) is checked whenever a user attempts to make a Perma Link (https://github.com/harvard-lil/perma/blob/f0e56e2a6b04d5456f1435dc1e72030801e51850/perma_web/perma/models/user.py#L397C1-L397C94 via https://github.com/harvard-lil/perma/blob/f0e56e2a6b04d5456f1435dc1e72030801e51850/perma_web/api/views/links.py#L177C1-L177C57) or [load the dashboard](https://github.com/harvard-lil/perma/blob/f0e56e2a6b04d5456f1435dc1e72030801e51850/perma_web/perma/views/link_management.py#L17), and it is only prepared for `PermaPaymentsCommunicationException`, not `ImproperlyConfigured`, and so will prevent the creation of new links with lots of 500s.

That might be too loud; thinking about just adding the extra exception handling in a followup PR just in case, although this should never happen.

> Cancellation is portal-only in Stripe mode (no self-serve Cancel button);

I don't know what that means... these must be terms of art, since "portal-only" sounds pretty self-service. I'd like to understand, so will follow up offline with others who may understand the distinction better.

> Minor UX gap: a cancel completed in the Stripe portal returns with ?update=success → shows "Your payment information was updated" rather than a cancellation-specific message (portal return doesn't distinguish the action).

I'll see that this is mentioned in Linear somewhere.

PR #3796, opened 2026-07-16
fix wrong purchase date bug

teovin:
This PR fixes the bug that was displaying the purchase date of the personal links incorrectly. The fix is to replace the month of the year parameter with the day of the month parameter.


rebeccacremona (approved):
lololol!!!!!!!!

PR #3797, opened 2026-07-21
Fix/guard null link limit effective timestamp

durkinm2:
Two small perma-side changes for the "scheduled downgrade isn't shown in perma" bug, paired with the perma-payments-stripe change that now reports the pending change.

- Template: the pending-downgrade section showed "Through <date>" and "Beginning <date>" with the same date (a Stripe downgrade takes effect at period end, which is also paid-through). Reworded to "Current plan (through <date>)" and "Scheduled change (effective <date>)" so it reads sensibly.
- Guard: `get_subscription` no longer does `None <= now`. If perma-payments ever sends a null effective timestamp, it's treated as already applied instead of 500-ing the usage-plan page. This is just a safety net, since the payments side should always supply it.

### DIFF
--- a/perma_web/perma/models/base.py
+++ b/perma_web/perma/models/base.py
@@ -284,7 +284,12 @@ class CustomerModel(models.Model):
         self.cached_paid_through = pp_date_from_post(post_data['subscription']['paid_through'])
 
         pending_change = None
-        if subscription_change_effective <= timezone.now():
+        # Perma Payments should always supply an effective timestamp, but the
+        # field is nullable there, so a missing value would raise on the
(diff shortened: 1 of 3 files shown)


rebeccacremona (approved):
Sure, LGTM!

Hopefully soon in The Great Refactor, we can either do away with link_limit_effective_timestamp entirely.... or, have the code make it mandatory for real, instead of just mandatory in practice. I'm not sure it makes sense for us to make allowances for it being `None` in error, more than it makes sense for allowing for `link_limit` or any of the other response fields to be `None`... but yup, the architecture does permit it!

PR #3799, opened 2026-07-22
Show Manage Subscription and Billing button on canceled subs

durkinm2:
See LIL-5356

When a subscription reports Canceled, the Usage Plan page dropped the "Manage Subscription and Billing" button, since that button only lived in the current/hold branch of the template. This adds the same portal button to the canceled state.

Pairs with the perma-payments-stripe fix that reports cancellations to perma (LIL-5360).

image


rebeccacremona (approved):
Oh great, so this, then, is how folks can access the new "uncancel" option 👍

PR #3802, opened 2026-07-23
Add a Frozen flag to block link creation for dispute/refund enforcement

durkinm2:
Adds a "Frozen" checkbox to user and registrar accounts. When frozen, the account can't create links, this overrides subscription status and bonus links. Staff set it in Django admin to enforce a dispute or refund, and uncheck it to restore access. The existing bonus-links field stays editable so staff can lower it to claw back granted links.

image

This is the manual freeze/clawback half of LIL-5353. The automated subscription cancel is a separate change on the perma-payments side.


rebeccacremona (approved):
LGTM!

I saw that the test run failed because, one of the installation steps is waiting for an upstream repo to push it's latest build artifacts. We just have to wait for that to become available, and then rerun; it doesn't usually take very long.

durkinm2:
Ahh gotcha thanks for clarifying Becky - and reviewing! Will circle back

PR #3807, opened 2026-07-24
"Grandfather" existing customers subscriptions over

rebeccacremona:
It is possible that the new payments system will not yet have records for our existing customers, when we are otherwise ready to switch over. This PR adds a new field, `grandfathered`, that we can set for any customer. When `True`, Perma will display a bespoke Usage Page page and will consult its own cached subscription information when determining whether a user may create new Perma Links. 

See LIL-4959 for screenshots and further discussion.


perma_web/perma/templates/settings/settings-usage-plan-grandfathered.html
```diff
@@ -0,0 +1,115 @@
...
+                      <dd>
+                        {{ customer.link_limit }}
+                      </dd>
+                    {% endif %}
+                  </dl>
+                </div>
+
+                <p class="page-dek">Please enjoy uninterrupted service during the transition. Billing will resume and outstanding balances will be reconciled when the migration is complete.</p>
+
+                <p class="page-dek">To cancel your subscription or adjust your level of service during the migration, please <a href="{% url 'contact' %}?subject=Cancel%20Subscription%20During%20Migration">contact us</a>.</p></p>
```
teovin: We can remove the duplicated closing p tag here

rebeccacremona: Oh! Thanks!

rebeccacremona:
```suggestion
                <p class="page-dek">To cancel your subscription or adjust your level of service during the migration, please <a href="{% url 'contact' %}?subject=Cancel%20Subscription%20During%20Migration">contact us</a>.</p>
```

perma_web/perma/views/user_settings.py
teovin, line 172: For the grandfathered customers, we will still call their purchase history here, right, even though we aren't showing the history of one-time purchases in the new template? So I think we need a guard in place so this method call doesn't cause an error on the usage plan page. (Correct me if I am wrong)

rebeccacremona: I was thinking it was safe to leave in place; I was thinking I would avoid touching it to avoid merge conflicts with any ongoing work related to purchase history. Will it cause an error? I didn't notice one when testing locally, but maybe I missed it! 

What I thought:
If Perma is wired up to Cybersource Perma Payments, this will retrieve information we can ignore.
If Perma is wired up to Stripe Perma Payments, this will find no data.

teovin: Hmm I see, I was thinking of this going live before wiring Stripe. And even if Stripe is wired, because the existing customer doesn't exist in Stripe yet, will it cause issues because Stripe will return an empty response? I am not sure lol. But in any case, maybe it is worth adding a check like `if self.grandfathered and not waffle.switch_is_active('allow_cybersource_transactions'): return None` to the method in base.py just to be safe. Up to you though!

teovin (approved):

perma_web/perma/views/user_settings.py
rebeccacremona: I don't think necessary.

So, we are not designing for a situation where, neither the Perma Payments (Cybersource) or Perma Payments (Stripe) application is up. Perma should always be able to talk to one app or the other. With both apps, it must be safe to hit /purchase-history/ for any Perma user... given that most will have none. 

When Perma is still talking to the Perma Payments Cybersource app (regardless of whether `waffle.switch_is_active('allow_cybersource_transactions')` is `True` or `False`), this call will continue to retrieve whatever purchase history is recorded there. When Perma is talking to the Perma Payments Stripe app, this call should return that there is no history, just like any other Perma user who has not yet bought any links via Stripe.

I truly think this is okay.

PR #3808, opened 2026-07-27
validate form for parsing errors

teovin:
It looks like when the `organization-users/add-multiple-users` flow receives a CSV with line breaks in data fields, it throws a `new-line character seen in unquoted field - do you need to open the file in universal-newline mode?` error, which breaks that page. A Google search and LLMs say that this happens with files created with older Macs, so I wasn't able to reproduce it by myself, but an LLM did and created a problem CSV for me through which I was able to repro the problem and confirm the fix which is to surface a parsing error to the user. 

**Before:**

image

**After:** 

Screenshot 2026-07-27 at 11 57 17 AM


perma_web/perma/models/base.py
```diff
@@ -5,7 +5,6 @@
 import waffle
 
 import requests
-import waffle
```
rebeccacremona: Okay, I'm not sure what's going to happen with removing this accidentally committed double import: Mike also removes it here https://github.com/harvard-lil/perma/pull/3810/changes#diff-bd7596e16c3bbb17cf669c53a8751ef467a31a60be4128eb3b7f2e756c286230L8. I think whoever merges first will win, and the other person will then have a merge conflict and will need to rebase/whatever.

teovin: Yep, it looks like it made it into develop in a previous PR (not sure how it bypassed linting there), but I will take care of it if it causes issues after Mike's PR is merged.

perma_web/perma/forms.py
```diff
@@ -515,37 +515,42 @@ def clean_csv_file(self):
...
+
+                if email.lower() in self.user_data:
+                    raise forms.ValidationError(f"CSV file cannot contain duplicate users: {email}")
+                else:
+                    self.user_data[email.lower()] = {
+                        'raw_email': email,
+                        'first_name': row.get('first_name', '').strip(),
+                        'last_name': row.get('last_name', '').strip()
+                    }
+        except csv.Error:
```
rebeccacremona: I think the try/except should only wrap the line that is expected to throw error, rather than this whole block of code. Do you know which line is raising the exception?

I'm also wondering if we can manage to give more specific advice than, "CSV file could not be parsed. Please check that the file is formatted correctly and try again." I'm worried that users will have trouble acting on that advice.

If the line that is raising the exception is, `reader = csv.DictReader(csv_file)`, then you can probably detect the "newline" message inside the exception, and make a specific recommendation if it's found and a generic one if not. If the exception is being raised later in this block, then I think you'll be able to tell them which row has a problem (by reporting either the row that has the problem, or the row before it).

Do you agree with this critique? Lemme know if not 🙂

rebeccacremona (changes requested):
Cool beans, thanks for tackling this. It is wild how many varieties of invalid input there can be lol.

I left some suggestions for a second pass. I'd also suggest that, since you have a CSV that can reproduce the error, adding a test for that condition.

If you think we should merge this as-is, and make a second pass later, say so!

Cheers -

perma_web/perma/forms.py
teovin: Hello, I decided to take a different approach with my last commit. I basically normalized the line endings since they are difficult to fix manually. I thought handling it in the code would make it easier for the end user. I also made the utf-8-sig change in this branch as well. My research (and local testing), like yours, showed that encoding with utf-8-sig handles BOM-marked files.

rebeccacremona (approved):
Love it.

PR #3809, opened 2026-07-27
prevent 500 coming from bad emails

teovin:
See LIL-4994 for details.

The fix is to `except` the ValueErrors that are being thrown as a result of the email field including NUL characters. 

Before:

Screenshot 2026-07-27 at 2 27 21 PM

After:

Screenshot 2026-07-27 at 2 39 15 PM


durkinm2 (approved):

rebeccacremona (commented):
To stop this resulting in a 500, and to stop us getting notified about it, this seems like a fine bandaid. But, I think it is safer/better to prevent null bytes from being sent to postgres in the first place, rather than allowing it and catching the error.

It's a little surprising to me that null bytes or other control characters can actually make it this far... but, FWIW, we do have a `remove_control_characters` utility. If we can't block higher up the stack, one safe solution might be to validate for control characters and set `target_user = None` if detected 🤷‍♀️

teovin:
@rebeccacremona Great suggestion about the utility command. I pushed an update and also added a small test.

rebeccacremona (approved):
Love it.

PR #3810, opened 2026-07-27
Replace the Purchase History list with a Stripe billing portal button

durkinm2:
Implements LIL-5357.

The "Purchase History" section is removed entirely and replaced with an "Additional Links" section containing one line, "Additional Links Remaining: N Links", and a blue "Billing and Purchase History" button into the Stripe portal. All history, including Cybersource-era purchases, is handled by contacting us directly.

The per-page-load call to `get_purchase_history()` is dropped under Stripe. The list is no longer rendered, so fetching it was a wasted HTTP round trip, and summing it under-reported for anyone who bought via Cybersource. Bonus-link crediting is unaffected, that runs off the /subscription/ payload in `credit_for_purchased_links()`, not this one.

The form posts directly to the payments app, the same pattern the existing bonus package buttons use, so there's no new Perma view or URL and it's one click to the portal instead of the current two-step.

Depends on the /billing/ endpoint in perma-payments.

image

### DIFF
--- a/perma_web/perma/templates/settings/settings-usage-plan.html
+++ b/perma_web/perma/templates/settings/settings-usage-plan.html
@@ -341,14 +341,9 @@
       {% if use_stripe_payments_app %}
         <hr>
         <h3 class="body-ch">Additional Links</h3>
-        <dl class="dl-horizontal">
-          <dt>
-            Additional Links Remaining
-          </dt>
-          <dd>
-            {{ request.user.bonus_links|default:0 }} Links
-          </dd>
-        </dl>
+        {# Not a dl-horizontal: its term column is a fixed 160px and clips #}
+        {# "Additional Links Remaining" to an ellipsis. #}
+        <p class="page-dek">Additional Links Remaining: {{ request.user.bonus_links|default:0 }} Links</p>
         {% if billing_portal_url %}
           {# Posts straight to the payments app, like the purchase buttons above, #}
           {# so receipts are one click away rather than two. #}
--- a/perma_web/perma/templates/settings/settings-usage-plan.html
+++ b/perma_web/perma/templates/settings/settings-usage-plan.html
@@ -338,7 +338,30 @@
       {% else %}
         <p class="page-dek">New purchases are temporarily unavailable. If you are working on a time-sensitive project and need to make a one-time purchase of additional links, please <a href="{% url 'contact' %}?subject=Time%20Sensitive%20New%20Project%20Inquiry">contact us</a> so we can assist you directly.</p>
       {% endif %}
-      {% if purchase_history.purchases %}
+      {% if use_stripe_payments_app %}
+        <hr>
+        <h3 class="body-ch">Additional Links</h3>
+        <dl class="dl-horizontal">
+          <dt>
+            Additional Links Remaining
+          </dt>
+          <dd>
+            {{ request.user.bonus_links|default:0 }} Links
+          </dd>
+        </dl>
+        {% if billing_portal_url %}
+          {# Posts straight to the payments app, like the purchase buttons above, #}
+          {# so receipts are one click away rather than two. #}
+          <div class="row subscription-buttons">
+            <div class="col-xs-12 col-md-6">
+              <form method="post" action="{{ billing_portal_url }}">
+                <input type="hidden" name="encrypted_data" value="{{ billing_encrypted_data }}">
+                <button class="btn btn-primary" type="submit">Billing and Purchase History</button>
+              </form>
+            </div>
+          </div>
+        {% endif %}
+      {% elif purchase_history.purchases %}
         <hr>
         <h3 class="body-ch">Purchase History</h3>
         <dl class="dl-horizontal">
--- a/perma_web/perma/tests/test_views_user_settings.py
+++ b/perma_web/perma/tests/test_views_user_settings.py
@@ -8,6 +8,7 @@ from django.conf import settings
 from django.test import override_settings
 
 from perma.exceptions import PermaPaymentsCommunicationException
+from perma.models import LinkUser
 
 from conftest import submit_form
 from unittest.mock import patch, sentinel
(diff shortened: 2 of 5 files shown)


perma_web/perma/tests/test_views_user_settings.py
```diff
@@ -339,6 +340,44 @@ def test_stripe_mode_shows_subscribe_and_purchase_forms(prepped, client, user_wi
...
+    user = user_without_subscription_or_purchase_history
+    user.bonus_links = 7
+    user.save()
+
+    client.force_login(user)
+    response = client.get(reverse('settings_usage_plan'), secure=True)
+
+    assert response.status_code == 200
+    # One line, whole and unsplit. A dl-horizontal was tried first and its
+    # 160px term column clipped the label to "Additional Links Rem...".
```
rebeccacremona: This comment looks to be commenting on the development process; delete I think?

perma_web/perma/tests/test_views_user_settings.py
```diff
@@ -339,6 +340,44 @@ def test_stripe_mode_shows_subscribe_and_purchase_forms(prepped, client, user_wi
...
+    assert b'Additional Links Remaining: 7 Links' in response.content
+    assert b'Billing and Purchase History' in response.content
+    assert b'Purchase History</h3>' not in response.content
+    assert f'{STRIPE_APP_URL}/billing/' in form_actions(response.content)
+    # Every form on the page posts somewhere: a missing billing_portal_url must
+    # never render as a form that posts back to the usage plan page itself.
+    forms = BeautifulSoup(response.content, 'html.parser').find_all('form')
+    assert forms and all(form.get('action') for form in forms)
```
rebeccacremona: I'm confused about this test. We just checked `assert f'{STRIPE_APP_URL}/billing/' in form_actions(response.content)`, so we know that `billing_portal_url` is not missing in the context of this test, right? I think if you want to test that condition, you need a separate test. But I'm also not sure if we need to separately test that condition: are we anticipating that might be missing sometimes?

perma_web/perma/tests/test_views_user_settings.py
```diff
@@ -339,6 +340,44 @@ def test_stripe_mode_shows_subscribe_and_purchase_forms(prepped, client, user_wi
...
+    # The list is gone from the page, so fetching it would be a wasted round
+    # trip. Bonus links are still credited, off the /subscription/ payload.
+    LinkUser.get_purchase_history.assert_not_called()
```
rebeccacremona: I'm confused about this. 

Nit pick: the comment seems redundant to me with the explanation in the code section. I'm wondering if it, like the above, is more a relic of the development process than an intentional explanatory comment for the tests?

But the actual confusion: `LinkUser.get_purchase_history.assert_not_called()`. I don't understand how that works: `assert_not_called` is a method on a mock/magick mock, but so far as I can see, this test doesn't mock that method. Also `get_purchase_history` is an instance method rather than a class method, and so isn't usually invoked from the class `LinkUser.get_purchase_history(...)`, but rather, on an instance, e.g., `user = LinkUser.objects.get(...); user.get_purchase_history()`; so this syntax catches my eye. 

It looks like the test suite is passing, but are we sure that assertion is running and is testing what we think?

rebeccacremona (commented):
Looks good to me! I asked a couple questions and made a couple small comments on the tests.

perma_web/perma/tests/test_views_user_settings.py
rebeccacremona: Oh wait, okay, there are mocks in the fixture:
```
@pytest.fixture
def user_without_subscription_or_purchase_history(mocker, link_user, no_purchase_history):
    get_subscription = mocker.patch('perma.models.LinkUser.get_subscription', autospec=True)
    get_purchase_history = mocker.patch('perma.models.LinkUser.get_purchase_history', autospec=True)

    get_subscription.return_value = None
    get_purchase_history.return_value = no_purchase_history

    yield link_user

    get_subscription.assert_called_once_with(link_user)
```

rebeccacremona: I jumped into the debugger, after that assertion:
```
(Pdb) LinkUser.get_purchase_history.assert_not_called()

(Pdb) dir(LinkUser.get_purchase_history)
['__annotations__', '__builtins__', '__call__', '__class__', '__closure__', '__code__', '__defaults__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__get__', '__getattribute__', '__getstate__', '__globals__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__kwdefaults__', '__le__', '__lt__', '__module__', '__name__', '__ne__', '__new__', '__qualname__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__signature__', '__sizeof__', '__str__', '__subclasshook__', '_mock_children', 'assert_any_call', 'assert_called', 'assert_called_once', 'assert_called_once_with', 'assert_called_with', 'assert_has_calls', 'assert_not_called', 'call_args', 'call_args_list', 'call_count', 'called', 'method_calls', 'mock', 'mock_calls', 'reset_mock', 'return_value', 'side_effect']

(Pdb) user.get_purchase_history()
{'purchases': [], 'total_links': 0}
(Pdb) LinkUser.get_purchase_history.assert_not_called()
*** AssertionError: Expected 'get_purchase_history' to not have been called. Called 1 times.
```
So, that is mocked, and that assertion -does- work when that method is called directly on the user object in the test.

rebeccacremona: Okay, and I experimentally altered the view to still call it:

Image

And yup, the assertion failed. 

Image

So, strange but true: this syntax works!

rebeccacremona (approved):
Okay, after digging into that Mock thing, updating to "approved" :-) but still recommending addressing those comments and that one assertion.

perma_web/perma/templates/settings/settings-usage-plan.html
```diff
@@ -338,7 +338,25 @@ <h2 class="body-ah">Get More Personal Links</h2>
       {% else %}
         <p class="page-dek">New purchases are temporarily unavailable. If you are working on a time-sensitive project and need to make a one-time purchase of additional links, please <a href="{% url 'contact' %}?subject=Time%20Sensitive%20New%20Project%20Inquiry">contact us</a> so we can assist you directly.</p>
       {% endif %}
-      {% if purchase_history.purchases %}
+      {% if use_stripe_payments_app %}
+        <hr>
+        <h3 class="body-ch">Additional Links</h3>
+        {# Not a dl-horizontal: its term column is a fixed 160px and clips #}
+        {# "Additional Links Remaining" to an ellipsis. #}
```
rebeccacremona, line 345: I still this this comment is a leftover remark from your agent, but we can delete later 🙂

rebeccacremona (approved):
Great, thanks for making those changes. I feel like that mocked test is a lot more readable now :-)

perma_web/perma/templates/settings/settings-usage-plan.html
durkinm2: LOL how did I miss that - thanks Becky!

durkinm2:
thank you @rebeccacremona for flagging these and for the careful review!

PR #3814, opened 2026-07-30
Add a banner for the payments app's downgrade-canceled redirect

durkinm2:
The payments app now redirects to `?change=canceled` after canceling a scheduled downgrade. This PR adds the matching banner: "Your scheduled downgrade has been canceled. Your current plan continues unchanged."  Fixes  a small UX gap.


rebeccacremona (approved):
🤘

PR #3815, opened 2026-07-30
Adjust downgrade-canceled banner wording per QA

durkinm2:
Tiny suggested change in copy (follow-up to #3814)

"Your current plan continues unchanged" is now "Your current plan will continue unchanged."


rebeccacremona (approved):
😂

PR #3818, opened 2026-08-12
Remove cybersource-flavored Perma Payments integration

rebeccacremona:
Until now, we've supported two different "flavors" of Perma Payments integration: one, that is designed to work with Cybersource, one that is designed to work with Stripe.

This PR intends to remove all the code for the Cybersource-flavored integration.

If you want to, reviewing commit-by-commit should make it clearer what was removed, and why; but, I would not feel pressured to give a very detailed review. I am hopeful that a lot of the code that remains, we will get to refactor soon (even as soon as later today lol).

Some major components:

- We used to bundle a containerized Perma Payments app into Perma's docker compose setup. I removed that for now. The intention, back when, was to allow a fuller test of integration tests, rather than tests that rely on mocked interactions... but a) it was never fully realized and b) I think it was more work that it was worth. So, at least for now, I removed it.

- We no longer need a special "cancellation confirmation" page, so that is removed.

- Settings are overhauled: you supply the "internal" and "external" Perma Payments URLs, as before, and the corresponding routes are assembled once in setting post processing, instead of at call time via a utility function.  

- All the feature flags we needed to manage the migration are removed

- I started untangling the template logic... but there is lots more we can do there, afterwards!

The tests were difficult to update. I am not worried that I deleted any essential assertions, etc., but neither am I confident that, as it currently stands, we have a good, comprehensive suite that tests purchase and subscription-related activities. I would like us to readdress that as a separate project: I think it will be much simpler, once all this stuff is deleted and the dust settles a bit.

This PR should be a no-op in prod: none of the deleted code should be running in prod anyway, because of how the switches are configured.


rebeccacremona:
LOL I forgot to lint, and left lots of unused imports. Fixing!

perma_web/perma/views/user_settings.py
```diff
@@ -188,71 +178,38 @@ def settings_usage_plan(request):
...
-        'subscribe_url': get_payments_app_url('subscribe'),
-        'cancel_confirm_url': reverse('settings_subscription_cancel'),
+        'purchase_url': settings.PAYMENTS_APP_URLS['purchase'],
+        'subscribe_url': settings.PAYMENTS_APP_URLS['subscribe'],
         'update_url': reverse('settings_subscription_update'),
         'accounts': accounts,
         'links_remaining': request.user.get_links_remaining(),
         'purchase_history': purchase_history,
         'bonus_packages': request.user.get_bonus_packages(),
         'display_cybersource_freeze_message': waffle.switch_is_active('display_cybersource_freeze_message'),
```
teovin: I think this can be removed along with the waffle import

perma_web/perma/tests/test_models.py
teovin: This and line 192 can be removed.

perma_web/perma/settings/deployments/settings_dev.py
```diff
@@ -71,14 +71,11 @@
...
-SUBSCRIBE_URL = 'http://localhost/subscribe/'
-CANCEL_URL = 'http://localhost/cancel-request/'
-SUBSCRIPTION_STATUS_URL = 'http://perma-payments/subscription/'
-UPDATE_URL = 'http://localhost/update/'
-CHANGE_URL = 'http://localhost/change/'
+
+# For now, point at an external Perma Payments application,
+# assumed to be running on the Docker host machine on port 8002
+STRIPE_PAYMENTS_APP_INTERNAL_URL = 'http://host.docker.internal:8002'
+STRIPE_PAYMENTS_APP_EXTERNAL_URL = 'http://perma-payments.test:8002'
```
teovin, line 78: Not a blocker but maybe we can also update settings.example with Stripe urls

perma_web/perma/models/base.py
teovin: Goodbye CyberSource, here and in a couple of docstrings in test_models.py

perma_web/perma/settings/utils/post_processing.py
```diff
@@ -97,3 +97,19 @@ def post_process_settings(settings):
...
+
+    # set up Payments App URLS
+    assert 'STRIPE_PAYMENTS_APP_EXTERNAL_URL' in settings and settings['STRIPE_PAYMENTS_APP_EXTERNAL_URL'] is not None, "Set DJANGO__STRIPE_PAYMENTS_APP_EXTERNAL_URL env var!"
+    assert 'STRIPE_PAYMENTS_APP_INTERNAL_URL' in settings and settings['STRIPE_PAYMENTS_APP_INTERNAL_URL'] is not None, "Set DJANGO__STRIPE_PAYMENTS_APP_INTERNAL_URL env var!"
+    settings['PAYMENTS_APP_URLS'] = {
+        'purchase': f"{settings['STRIPE_PAYMENTS_APP_EXTERNAL_URL']}/purchase/",
+        'purchase_history': f"{settings['STRIPE_PAYMENTS_APP_INTERNAL_URL']}/purchase-history/",
+        'acknowledge_purchase': f"{settings['STRIPE_PAYMENTS_APP_INTERNAL_URL']}/acknowledge-purchase/",
+        'subscribe': f"{settings['STRIPE_PAYMENTS_APP_EXTERNAL_URL']}/subscribe/",
+        'cancel': f"{settings['STRIPE_PAYMENTS_APP_EXTERNAL_URL']}/cancel-request/",
```
teovin: I think this route isn't used anywhere

perma_web/perma/models/base.py
rebeccacremona: Thanks for a second set of eyes on things I missed!!

perma_web/perma/settings/utils/post_processing.py
rebeccacremona: Huh! Interesting. I'm going to look one more time, because that's what I see too... but it's also a route in the Perma Payments app! https://github.com/harvard-lil/perma-payments-stripe/blob/main/stripe_payments/urls.py#L19C1-L19C74, https://github.com/harvard-lil/perma-payments-stripe/blob/aa919ab80565e439d58185cbc55910989284a1fd/stripe_payments/views.py#L931

Totally unused? Hmmmmmmm.

rebeccacremona: Ah okay, there WAS a cancellation button at one point, but it was removed. LIL-5354.

👍

perma_web/perma/settings/deployments/settings_dev.py
rebeccacremona: Ah! To be discussed! I think settings.example.py is pretty useless, and it lists a bunch of stuff that I don't think belongs in there. Unclear to me why anything is listed except feature flags.

rebeccacremona:
Thanks, @teovin! I pushed all the additional deletions; I didn't add anything to settings.example.py (pending further discussion of that file)... but I did remove references to the old settings there! Re-requested review for formality 🙂

teovin (approved):

PR #3820, opened 2026-08-25
Refactor customer's "usage plan" page and customer models

rebeccacremona:
This PR does two closely related things, both to improve the legibility of Perma's codebase as pertains to paying customers.

It is easiest to read commit-by-commit; recommended if the overall diff feels overwhelming.

First, it extracts the `Customer` abstract class out into its own submodule in `models`. Why? Well, to be perfectly honest: because I -keep looking for it there- every time I want to find that logic, ever since the refactor that split up models.py into submodules lol. It's been many months, so at this point, I take this as evidence that that's a good idea 😄. Second, for consistency with the rest of the repo, it moves the small amount of remaining "helper" logic out of `models/base.py` into a `models/utils.py`. These are both stylistic; I do think helpful, but equal parts taste and substance.

Then, it refactors the templating of the main page that Perma's paying users interact with: the Usage Plan page. That page started off small and got more and more complex over the intervening decade, until during the Cybersource -> Stripe transition, it became totally unmanageable, as all complained 😂 . This PR switches it from a single, highly complex template with lots and lots of nesting and `if/then/else` to one simple base template, and then a series of includes.

The base template is nice and short:
```
{% extends "settings-layout.html" %}
{% block title %} | Settings{% endblock %}
{% block dashboardContent %}

  {% if payment_status_message %}
    <div class="alert alert-{{ payment_status_level }} alert-block">{{ payment_status_message }}</div>
  {% endif %}

  {% for account in accounts %}
    {% with subscription=account.subscription %}
      {% with customer=account.customer %}
        {% with customer_type=customer.customer_type %}

          {% if customer_type == 'Registrar' %}
            <h2 class="body-ah">Registrar Subscription &mdash; {{ customer.name }}</h2>
          {% else %}
            <h2 class="body-ah">Personal Subscription</h2>
          {% endif %}

          {% if not subscription %}
            {% include 'includes/subscription-purchase.html' %}
          {% else %}
            {% include 'includes/subscription-details.html' %}
          {% endif %}

        {% endwith %}
      {% endwith %}
    {% endwith %}
  {% endfor %}

  {% if request.user.can_purchase_bonus_links %}
    {% include 'includes/bonus-links-purchase.html' %}
  {% endif %}

{% endblock %}
```
That's it! 

The includes also make use of reusable components, like `subscription-definition.html`, for improved consistency: in the original mammoth template, there was a fair bit of near-duplication, with subtle details differing. By extracting these bits of logic into reusable files, those differences are both reduced and made easier to spot... which will smooth the way for future UI improvements. 

Finally, this PR includes a small bug fix, the problem uncovered by the refactoring (which I think proves the refactoring was worth it!). This simpler templating made it clear that, in the old setup, it would have in theory been possible for a non-paying, "unlimited" user to purchase (unnecessary) bonus links. In reality... no such user was presented with that option, but this PR adds code that removes the possibility.

Screenshots forthcoming but: everything should look exactly the same as before, no changes 😁

### DIFF
--- a/perma_web/perma/models/utils.py
+++ b/perma_web/perma/models/utils.py
@@ -2,9 +2,7 @@ from django.db import models
 from django.db.models import Count
 from django.utils import timezone
 from taggit.models import CommonGenericTaggedItemBase, TaggedItemBase
-import logging
 
-logger = logging.getLogger(__name__)
 
 ### HELPERS ###
 
--- a/perma_web/perma/models/utils.py
+++ b/perma_web/perma/models/utils.py
@@ -1,0 +1,80 @@
+from django.db import models
+from django.db.models import Count
+from django.utils import timezone
+from taggit.models import CommonGenericTaggedItemBase, TaggedItemBase
+import logging
+
+logger = logging.getLogger(__name__)
+
+### HELPERS ###
+
+# functions
+def link_count_in_time_period(links, start_time=None, end_time=None):
+    if start_time and end_time and (start_time > end_time):
+        raise ValueError("specified end time is earlier than specified start time")
+    elif start_time and end_time and (start_time == end_time):
+        links = links.filter(creation_timestamp=start_time)
+    else:
+        if start_time:
+            links = links.filter(creation_timestamp__gte=start_time)
+        if end_time:
+            links = links.filter(creation_timestamp__lte=end_time)
+    return links.count()
+
+def most_active_org_in_time_period(organizations, start_time=None, end_time=None):
+    if start_time and end_time and (start_time > end_time):
+        raise ValueError("specified end time is earlier than specified start time")
+    # unlike 'link_count_in_time_period', no special behavior required
+    # if start_time = end_time here. the end result is the same
+    else:
+        if start_time:
+            organizations = organizations.filter(links__creation_timestamp__gte=start_time)
+        if end_time:
+            organizations = organizations.filter(links__creation_timestamp__lte=end_time)
+        return organizations\
+            .annotate(num_links=Count('links'))\
+            .exclude(num_links=0)\
+            .order_by('-num_links')\
+            .first()
+
+
+# classes
+
+# django-taggit assumes the model being tagged has an integer primary key.
+# per http://django-taggit.readthedocs.io/en/latest/custom_tagging.html,
+# tag "through" this class if your model has a string as primary key.
+# tags = TaggableManager(through=GenericStringTaggedItem)
+# (copied straight from their docs)
+class GenericStringTaggedItem(CommonGenericTaggedItemBase, TaggedItemBase):
+    object_id = models.CharField(max_length=50, db_index=True)
+
+
+class DeletableManager(models.Manager):
+    """
+        Manager that excludes results where user_deleted=True by default.
+    """
+    def get_queryset(self):
+        # exclude deleted entries by default
+        return super(DeletableManager, self).get_queryset().filter(user_deleted=False)
+
+    def all_with_deleted(self):
+        return super(DeletableManager, self).get_queryset()
+
+
+class DeletableModel(models.Model):
+    """
+        Abstract base class that lets a model track deletion.
+    """
+    user_deleted = models.BooleanField(default=False, verbose_name="Deleted by user")
+    user_deleted_timestamp = models.DateTimeField(null=True, blank=True)
+
+    class Meta:
+        abstract = True
+
+    def safe_delete(self):
+        self.user_deleted = True
+        self.user_deleted_timestamp = timezone.now()
+
+
+
+
--- a/perma_web/perma/templates/includes/subscription-current-pending-change.html
+++ b/perma_web/perma/templates/includes/subscription-current-pending-change.html
@@ -6,12 +6,12 @@
   <div class="col-xs-12">
     <h3 class="body-ch">Current plan (through {{ subscription.paid_through | date:"F jS, Y" }})</h3>
     <dl class="dl-horizontal">
-      {% include 'includes/subscription-definition.html' with sub=subscription paid_though=False paid=True %}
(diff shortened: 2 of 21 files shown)


rebeccacremona:
personal-registrar-purchase
registrar-canceled
registrar-on-hold
registrar-current
personal-canceled-plus-bomus
personal-hold
personal-downgrade-pending
personal-current
personal-purchase

perma_web/perma/templates/includes/subscription-current-pending-change.html
```diff
@@ -0,0 +1,17 @@
+<p class="page-dek">
+  Your subscription is <span class="blue-text">{{ subscription.status | lower }}</span>.
+</p>
+
+<div class="row">
+  <div class="col-xs-12">
+    <h3 class="body-ch">Current plan (through {{ subscription.paid_through | date:"F jS, Y" }})</h3>
+    <dl class="dl-horizontal">
+      {% include 'includes/subscription-definition.html' with sub=subscription paid_though=False paid=True %}
```
teovin: typo fix: paid_through

perma_web/perma/templates/includes/subscription-current-pending-change.html
```diff
@@ -0,0 +1,17 @@
...
+    </dl>
+
+    <h3 class="body-ch">Scheduled change (effective {{ subscription.pending_change.effective | date:"F jS, Y" }})</h3>
+    <dl class="dl-horizontal">
+      {% include 'includes/subscription-definition.html' with sub=subscription.pending_change paid_though=False paid=False %}
```
teovin: typo fix: paid_through

perma_web/perma/templates/includes/subscription-definition.html
```diff
@@ -0,0 +1,47 @@
...
+    ${{ sub.rate | intcomma }}/year{% if paid %} (paid){% endif %}
+  </dd>
+{% endif %}
+
+{% if paid_through %}
+  <dt>
+    Paid Through:
+  </dt>
+  <dd>
+    {{ subscription.paid_through | date:"F jS, Y" }}
```
teovin: I see that we used `sub` throughout this template, except here it is `subscription`. Should we change this to sub as well? I see the including templates pass the subscription as sub.

perma_web/perma/models/utils.py
```diff
@@ -0,0 +1,80 @@
+from django.db import models
+from django.db.models import Count
+from django.utils import timezone
+from taggit.models import CommonGenericTaggedItemBase, TaggedItemBase
+import logging
+
+logger = logging.getLogger(__name__)
```
teovin: Looks like we are not using logger anywhere in this file, I think we can remove this and the import statement.

perma_web/perma/templates/includes/top_banner_grandfathered.html
```diff
@@ -1,4 +0,0 @@
-<div class="top-banner">
-  <p class="top-banner-title">Payment System Upgrade In Progress</p>
-  <p>Payment, billing, and cancellation services may be limited during this time. We apologize for any inconvenience and appreciate your patience as we improve the payment experience. We will be in touch with more details soon. Reach out to <a href="mailto:info@perma.cc?subject=payment%20system%20upgrade">info@perma.cc</a> if you have any questions.</p>
-</div>
```
teovin, line 4: I see that this was removed; however, we are still referencing it in create-link.html. This is harmless right now as we don't have any grandfathered users. I'm mostly noting it for myself to clean it up while working on that grandfathering cleanup.

teovin:
Great work @rebeccacremona! I noted a few minor things; I think it is good to go after those are addressed.

rebeccacremona:
> Great work @rebeccacremona! I noted a few minor things; I think it is good to go after those are addressed.

Thank you for the detailed read-through and review! I will fix up the things you noticed :-)

perma_web/perma/templates/includes/subscription-current-pending-change.html
rebeccacremona: Ah! I am realizing that with this typo (paid_though -> instead of paid_through), that the value of `paid_through` was still falsey (None instead of False) as far as the `include` was concerned, which is why things were still printing correctly!

Thanks for catching!

perma_web/perma/templates/includes/subscription-definition.html
rebeccacremona: Ah, that's a good one for us to maybe eliminate wholesale in this refactoring: is the nicknaming of `subscription` to `sub` to save tying `len('scription')` characters really worth it? I will either make this one reference `sub` or will change them all to `subscription`.

rebeccacremona: Okay, I started to switch them all to `subscription`... and then realized why it's `sub`: it's so that, for the template that shows when you have a pending change, which needs to show TWO sets of details, one for the current applicable subscription conditions and one for the upcoming conditions, you can pass in `sub=subscription` first and then `sub=subscription.pending_change`! So, I will correct what you noticed from `subscription` to `sub`, rather than changing everywhere.

This could be made MUCH clearer by using `only`, in the `include`s across the board, so that we aren't utilizing context variables in global scope (e.g. `{% include "name_snippet.html" with greeting="Hi" only %}`... which is why `subscription` works the same as `sub` here... but I feel like that sort of refactoring makes more sense as part of a larger project, say, one preparing us for a JS frontend :-)

perma_web/perma/templates/includes/top_banner_grandfathered.html
rebeccacremona: Ahh! Good point! I did not remember it was used on that page, and forgot that template errors like that would go silently rather than raising an exception.

👍 to leaving it here, so that the grandfathering cleanup PR is nice and self-contained.

teovin (approved):

PR #3822, opened 2026-08-28
dont increment counts in the case of a deleted resaved link

teovin:
This PR modifies the link pre-save signal to include user-deleted links in the initial link filtering. Otherwise, a previously user-deleted link is treated as a new link, which inflates the user, org, and registrar link numbers (if set). 

This is not a very common case, I think, but I thought it still needed addressing in the greater link counts rethinking effort.

**To reproduce:**

Create a link: User's link count should increment
Delete it: User's link count should decrement
Go to admin and save the link without any changes: User's link count increments. (Work is to fix this).

### DIFF
--- a/perma_web/perma/tests/test_link_count_caching.py
+++ b/perma_web/perma/tests/test_link_count_caching.py
@@ -7,26 +7,29 @@ def test_link_count_do_not_increment_after_saving_a_deleted_link(org_user, link_
         submitted_url="http://example.com",
         organization=organization,
     )
+    org_user.refresh_from_db()
+    organization.refresh_from_db()
+    registrar.refresh_from_db()
+    assert org_user.link_count == 1
+    assert organization.link_count == 1
+    assert registrar.link_count == 1
     link.safe_delete() 
-    link.save() # user should have 0 links now
+    link.save()
 
-    # get the link counts before the link is saved again
     org_user.refresh_from_db()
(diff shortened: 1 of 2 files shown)


perma_web/perma/tests/test_link_count_caching.py
```diff
@@ -1,3 +1,34 @@
...
+
+    # get the link counts before the link is saved again
+    org_user.refresh_from_db()
+    organization.refresh_from_db()
+    registrar.refresh_from_db()
+    user_count = org_user.link_count
+    org_count = organization.link_count
+    registrar_count = registrar.link_count
+
+    link.save() # user should still have 0 links
```
rebeccacremona: Instead of this comment, "# user should still have 0 links", do you want to include an `assert org_user.link_count == 0`? Like, watch it be `1` before deleting and then `0` again?

rebeccacremona (approved):
LOL, this is a funny one, and a good one. Well-spotted.

I made a small remark about the test: I'm not sure a comment serves best there? But if I'm misunderstanding, happy to hear why :-)

perma_web/perma/tests/test_link_count_caching.py
teovin: Oh, that comment was mostly for me to keep track of what's happening, but I agree, I might as well place all of the mental load on the test :)

PR #3823, opened 2026-08-28
Read pending_change from perma payments directly

durkinm2:
Step 2 of 3 for LIL-5426.

Perma infers scheduled downgrades when perma payments reports the future plan's values with a future effective date. As of yesterday, perma payments also states the downgrade in an explicit `pending_change` field (step 1, in prod). This PR switches perma to reading that field and drops the date inference.

While a downgrade is pending we still leave the local plan fields alone, same as before, just triggered by the explicit field instead of date math. If the field is ever missing (an older payments build), we treat it as no change pending rather than erroring.

### DIFF
--- a/perma_web/conftest.py
+++ b/perma_web/conftest.py
@@ -963,7 +963,8 @@ def spoof_pp_response_subscription():
                 "paid_through": "1970-01-21T00:00:00.000000Z",
                 "link_limit_effective_timestamp": "1970-01-21T00:00:00.000000Z",
                 "link_limit": "unlimited",
-                "reference_number": "PERMA-1237-6200"
+                "reference_number": "PERMA-1237-6200",
+                "pending_change": None
             },
             "purchases": []
         }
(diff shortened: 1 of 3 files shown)


rebeccacremona:
image
This happens when mkcert is in the process of making a new release... and the docker image for our architecture isn't available yet. Re-running the tests in a few minutes ought to take care of it. I think their build has been slower lately.

rebeccacremona:
image
LOL! Of course then, the single flakey end-to-end test also flaked.

Third time's the charm!

durkinm2:
😂...Voila! Thank you for clueing me in on why that was happening. Requested a review from you, no urgency, whenever you get a chance!

rebeccacremona (changes requested):
This looks great to me! I like the idea of moving towards this shape of data, which more closely matches Stripe's own. If we come to support multiple "scheduled" changes, this seems easy to expand.

One request: I think the code comments that refer to old Perma Payment builds, and previous behavior, will be confusing for future readers, who won't (and don't need to) know about the past. What do you think about buzzing through this file, and removing protections to handle old versions' of Perma Payments behavior, and adjusting comments to limit context to still-current behavior? Push back if you think this is a poor idea :-)

durkinm2:
Yup good call! I went through and tweaked those comments and removed a single protection where .get() was a safety net for talking to an old perma payments that didn't send the `pending_change` field.

I also see 'Cancellation Requested' in **ACTIVE_SUBSCRIPTION_STATUSES** which is leftover from Cybersource and our current service can never send. There is no risk leaving this other than potential developer confusion, whereas removing it could theoretically affect a rare user state and should be verified first. So I say leave it as is for now!

Think I covered it all but please LMK. Thanks Becky!

rebeccacremona (approved):
👍 ah, good catch on the "cancellation requested" status: something else for us to delete later!

PR #3824, opened 2026-08-28
add an ondemand task to correct linkuser link counts

teovin:
Now that a [previous PR](https://github.com/harvard-lil/perma/pull/3822) fixed a LinkUser `link_count` bug going forward, we need a task to fix those currently having incorrect numbers. This PR adds the ability to do that. 

Added: `dry_run` param so we can see the count of records that would be updated if the job was to be run.


perma_web/perma/models/link.py
```diff
@@ -177,6 +177,7 @@ class Meta:
             models.Index(fields=['creation_timestamp', 'guid']),
             models.Index(fields=['-creation_timestamp', 'guid']),
             models.Index(fields=['submitted_url_surt']),
+            models.Index(fields=['created_by', 'user_deleted']),
```
rebeccacremona: I am wondering about this index. 

Usually, because of the [manager that excludes soft-deleted links by default](https://github.com/harvard-lil/perma/blob/develop/perma_web/perma/models/link.py#L113) from queries, the first field in the WHERE clause of SQL generated by the ORM, and the first field needed in indices for normal `Link.object.filter` queries, is `user_deleted`, and then anything you add follows. Indicies are also sequential sorts: `'created_by', 'user_deleted'` isn't used by the db in the same conditions as `'user_deleted', 'created_by'`.   

So I'm wondering if this index `models.Index(fields=['created_by', 'user_deleted'])` indeed gets used for:
```
  non_deleted_link_count = Subquery(
        Link.objects.filter(created_by_id=OuterRef("pk"))
        .values("created_by_id")
        .annotate(count=Count("pk"))
        .values("count")
    )
```

Have you seen that getting used? I believe you if yes 😀 , I am not such an expert that I trust my instincts here in the face of... evidence 😂

perma_web/perma/celery_tasks.py
```diff
@@ -1617,6 +1618,50 @@ def warn_expiring_organization_users(warning_days=None):
                 break
 
 
+@shared_task(acks_late=True)
+def reconcile_user_link_counts(dry_run=False, batch_size=None):
```
rebeccacremona: I love the idea of these "dry run" and "batch size" parameters. Great thinking.

My instincts are, we need to further optimize this to make it safe to run in production. 

-------

First, my suspicion is that running 
```
    non_deleted_link_count = Subquery(
        Link.objects.filter(created_by_id=OuterRef("pk"))
        .values("created_by_id")
        .annotate(count=Count("pk"))
        .values("count")
    )

    actual_link_count = Coalesce(non_deleted_link_count, Value(0))
    users = LinkUser.objects.all()
    users = users.exclude(link_count=actual_link_count).order_by("pk")
```
would be expensive in prod, both because counts are expensive, and because both the link table and the user table are so large. Even if we use the "batch_size" parameter to work with the results in smaller sets, this query will still attempt to do the calculation for every user in the user table, which I think is probably too expensive for us safely run, especially given that, since new users and new links might be created while this is running and affect the results... for accuracy, the query might need to freeze both tables.

In the past, I have occasionally tested potentially expensive queries locally, by loading a production-like copy of the database into my dev environment. But I haven't done that since, probably, the last time you did, when we were working on benchmarking WARC - > WACZ!

If you think this might in fact be safe and I am worrying too much, let's test and document that testing in some way... for instance, by trying it with 100 users, then 1000 users, etc. to see if the RAM and timing seem okay... if not testing against the full `Link` and `User` tables. (I honestly don't know how testing against superset would inform how production might behave: that could be good for us to know too!)

I think likely safest to architect such that we will only ever look at a subset of users at a time. (I can imagine different batching strategies, like pk ranges... or, by the first letter of email address, or etc.)

-----

Further, if I'm reading this correctly, I think this query will re-run in its entirety every time `batch_ids = list(users.values_list("pk", flat=True)[:batch_size])` is reached in the "while" loop. So, passing a small batch size will not, I believe, particularly save us resources: it will limit the scope of the `update` statement inside the transaction, but will not save us from the expensive full-table counting. Also, I think the counting will happen a second time, for each user in the batch, during the execution of the `update`. I don't -think- any cahcing of queries will protect us there. 

------

Finally, I am concerned that this celery task will take longer to run than the allowed time limit for celery tasks. 

A pattern we often use, when we need to run a task for many batches of things is, divide the work up by batches and then dispatch a secondary task, one per batch. That has the advantage of a built-in way to track progress (because you can watch the task count go down), as well as a way to limit execution time, and a way to easily re-queue a subset if there is a failure of any kind, or the process is interrupted by a deployment, or etc.

------

While counting the total number of affected users in advance sounds -very- interesting to me, I suspect the least expensive way to clean up this problem is, iterate through the entire LinkUser table once, not attempting to filter in advance on whether their link count is accurate or not, and have the 

I understand the instinct to want to send a smaller number of large queries, rather than a larger number of smaller queries to update a single user at a time... Agreed, that sort of thing is usually a great optimization. 

I am wondering what the most bang for our buck would be here. I suspect, letting the `update` update every user's `link_count` (which would just be a no-op if the figure is already correct), rather than counting, filtering, and then counting again during the update... and estimating how many users will be corrected separately (maybe being content with superset). I suspect one-at-a-time would be fine... as would batching smaller collections of users... One-at-a-time would also fine-grained logging (say, of every user who was corrected), batches would allow for fewer queries. 

------

What are your thoughts?

How were things for you, when developing the PR?

I think this would a totally reasonable for us to chat with a well-informed model about too, to help us think through what is reasonable for our amount of data.

perma_web/perma/celery_tasks.py
```diff
@@ -1617,6 +1618,50 @@ def warn_expiring_organization_users(warning_days=None):
...
+    non_deleted_link_count = Subquery(
+        Link.objects.filter(created_by_id=OuterRef("pk"))
+        .values("created_by_id")
+        .annotate(count=Count("pk"))
+        .values("count")
+    )
+
+    actual_link_count = Coalesce(non_deleted_link_count, Value(0))
+    users = LinkUser.objects.all()
+    users = users.exclude(link_count=actual_link_count).order_by("pk")
```
rebeccacremona: Whee. Nice use of fancy ORM features to craft subtle SQL.

rebeccacremona (changes requested):
Hooray for correcting these figures!

Some thoughts below.

perma_web/perma/models/link.py
teovin: Evidence suggests your hunch :) I logged the raw query, which was doing a where filter on those fields together, but the explain analyze didn't show the composite index was being used, so I just removed it.

perma_web/perma/celery_tasks.py
teovin: Hello, I changed it a bit to reduce the DB lookups. With the current way, one query will be needed to create the link dict, one query to get the link user objects and one additional query to run the bulk update if mismatches are found. (So 2, + 1 additional). I also decided to get rid of the `batch_size` as it wasn't doing much other than updating records in batches. As for `dry_run`, I left it as is. I know the total is about 2500 in prod, but I thought we might need it in the future depending on changes to our superset permissions.

rebeccacremona (changes requested):
Cool! This is feeling safer!

A couple small thoughts:

As is, I don't think this wants to be a celery task: celery tasks are great for asynchronous, concurrent flocks of work, but this is structured as a one-time, long-running job. That seems well-suited for an "invoke" task instead. Then, we won't have to worry about celery tasks' time, RAM, or other limits: we'll just be able to do the long-running thing. Does that sound good to you too? Are there advantages to Celery I'm not thinking of?

I also think it would be a good idea to add some logging so that we know where it is in the process. How about a line before and after the building of the `link_count`, to log how long that query takes? (BTW I looked into how much RAM a dict of that size will take: non-trivial, but we have plenty of RAM on the workers. I have bogged them down with overlarge dicts in the past, but this one doesn't approach that limit).  I am not sure if `tqdm` will give us a nice progress bar through the `LinkUser.objects.only("pk", "link_count").iterator()` loop, because I can't remember if `tqdm` works well with `iterator`... but that is probably worth it too, if it works. Having run things like this in the past... I just remember how anxious a person gets, waiting and seeing no output. I 

Finally, a thought on the index: I would predict that if you reversed the column order, it would use it. That is to say, `models.Index(fields=['user_deleted', 'created_by'])` instead of `models.Index(fields=['created_by', 'user_deleted'])`. That can be hard to test locally without a lot of data: sometimes when I am testing things like that locally, I ask a model to fill my database with plausible test data, and have it clean up afterwards. I am not sure whether, for a one-time task, the performance gain if it uses the index is worth it, but I have not tested the timing of anything. If you think we're good without it, I trust you!

Marking as "changes requested", but I think effectively ready to merge, those remarks are not very big :-)

teovin:
Oh my, I swear a one-time invoke task was my intention all along, but somehow I forgot about our tasks folder and invoke. Maybe because I haven't looked in there for a while. Thank you for that, I moved it. 

I also used tqdm. The iteration there is so quick that it doesn't leave much time to see the numbers rise gradually, but they do - just super quick. 
image

rebeccacremona (approved):

teovin:
Also this:

image

PR #3825, opened 2026-08-31
make user deleted and created by fields readonly

teovin:
The `user_deleted` and `created_by` field modifications via the `admin/perma/link/<guid>/change/` form are areas that might cause LinkUser link_count mismatches. From what I understand, these were not turned on out of necessity. I am making these read-only for now.


rebeccacremona (commented):
LGTM!

rebeccacremona (approved):

Issue #2225, opened 2018-01-29
Race condition when checking user link count?


rebeccacremona:
If yes, fix using select_for_update

PR #3828, opened 2026-09-11
Organization link_count fixes

teovin:
This PR addresses these two scenarios when it comes to the `link_count` field accuracy of the organization model. Currently:

1- Moving folders from personal links folders into organization folders doesn't increment the destination organization and destination registrar counts, and similarly, moving folders from under organization folders into personal links folders doesn't decrement the source organization and source registrar counts.

2- Moving a folder from under one org to another doesn't update the source/destination org and registrar counts. The registrar count only needs to be updated if movement is happening between orgs of different registrars.

**Edited to add:** I was going to create a separate PR for the reconciliation tasks to fix org and registrar counts, but I ended up pushing them to this branch by mistake. They are pretty similar to the link user equivalent, so I think I will leave it here instead of removing that commit etc. A note on this is that I haven't gotten to registrar count fixes yet, including the new field for the sponsored links, the recon task for the registrars here is only in the context of the org count fixes. 

Screenshot 2026-09-14 at 10 36 29 AM


perma_web/perma/models/folder.py
```diff
@@ -147,6 +148,65 @@ def update_cached_path(ids, tree_root_id):
                 )[:1]
             )
 
+        def update_org_and_registrar_link_counts(links, new_org_id):
+            """ update the link counts for organization and registrar after folder movements """
```
rebeccacremona, line 152: Question about the intended use of this method. I only see it being used after a single folder tree is moved. Is that the only intended use?

If so, then I am confused about why the code is structured so as to handle the links in the passed-in queryset originally belonging to different organizations, rather than all belonging to the same one.

Am I correct that that's not necessary? If so: I think this method would be easier to read and reason about if it was clearly just decrementing the originating org's total (previous_parent.organization_id) and incrementing the destination's, by the same amount, which, if I understand correctly, is count of all the links in the queryset.

If I am incorrect: when do you picture this method being called on a heterogeneous link queryset?

perma_web/perma/models/folder.py
```diff
@@ -147,6 +148,65 @@ def update_cached_path(ids, tree_root_id):
...
+                        registrar_deltas[old_registrar_id] -= num_of_links
+
+                    if new_registrar_id:
+                        registrar_deltas[new_registrar_id] += num_of_links
+
+            for org_id, delta in org_deltas.items():
+                Organization.objects.filter(pk=org_id).update(link_count=F('link_count') + delta)
+
+            for registrar_id, delta in registrar_deltas.items():
+                Registrar.objects.filter(pk=registrar_id).update(link_count=F('link_count') + delta)
```
rebeccacremona, line 208: Not directly related, but: I got to thinking about `F` and counts, and whether here (and with the bonus link calculation below) we are okay without some lock/some [`select for update`](https://github.com/harvard-lil/perma/issues/2225), to prevent race conditions. Here in the folder-moving context, we have locks in place at the top of things, so maybe not, maybe ALL SET already...

But in reasoning about that, I got to reading https://github.com/harvard-lil/perma/blob/main/perma_web/perma/signals.py, and realized that code doesn't even use `F` expressions at all. So that, surely, is another category of bug affecting the accuracy of all these counts!! I think that doesn't affect this PR, which is fixing its own, self-contained problem. But another one for us to get into Linear!!

perma_web/perma/models/folder.py
```diff
@@ -233,6 +293,7 @@ def update_parents_cached_has_children(parent_id=None, previous_parent_id=None):
                 from .link import Link
                 # update the de-normalized reference to owning org on any links in this folder's subtree
                 links = Link.objects.filter(folders__in=subtree_ids)
+                update_org_and_registrar_link_counts(links, parent.organization_id)
```
rebeccacremona, line 296: Very very small question: do you like updating the counts before the link's `organization_id` for a practical reason, or just stylistic 🙂? 

No huge opinion here, but if just stylistic, I might recommend putting below ` links.update(organization_id=parent.organization_id)` with its own comment so that the `# update the de-normalized reference to owning org on any links in this folder's subtree` clearly stays as just a label to the original two lines.

perma_web/perma/tests/test_link_count_caching.py
```diff
@@ -116,3 +217,109 @@ def test_link_count_for_registrars(registrar_user, link_factory):
...
+
+    org_user.refresh_from_db()
+    organization.refresh_from_db()
+    registrar.refresh_from_db()
+    assert org_user.link_count == user_count
+    assert organization.link_count == org_count + 2
+    assert registrar.link_count == registrar_count + 2
+
+
+def test_moving_folder_out_of_org_updates_org_and_registrar_counts(org_user, folder_factory, link_factory):
```
rebeccacremona, line 248: Here is a possibly additional case I am wondering about: what happens if a user moves a folder out of an org into their own personal links folder, and back again?

I think your code covers it, but the bit I'm wondering if should have a test: an org folder's links can contain links `created_by` many different users. While it would be -weird- for a user to then put all those in Personal Links... it's allowed lol. Maybe worth including a test just for documentation of what happens to the different counts?

perma_web/tasks/once.py
```diff
@@ -282,3 +282,75 @@ def reconcile_user_link_counts(ctx, dry_run=False):
     logger.info(f"Updated {mismatch_count} LinkUser records.")
     return mismatch_count
 
+
+@task
+def reconcile_organization_link_counts(ctx, dry_run=False):
+    """
+    Set Organization.link_count to the count of non-deleted links under each organization.
+    Pass dry_run=True to get the count of mismatches without updating.
+    """
```
rebeccacremona, line 291: Two notes on this task:

One is, I wonder about the expected performance of the `link_counts` query in production. Since the `link_counts` query in https://github.com/harvard-lil/perma/pull/3824/changes#diff-bff751b0b0d2707b69da785eef8dd8f00e78932f323d0e06c0166cd836310587R258 was fine, and this doesn't seem materially different, then I expect this too will be okay. I remember sometimes filtering on `isnull=False` is surprisingly expensive... but surely this will be okay?

The other is, I'm wondering if we're cool to run this while the application is up and being used by users. We did not discuss that when fixing user.link_count. I haven't thought through the nuances, which will of course depend on the final details of this code. Do you have thoughts there?

perma_web/tasks/once.py
```diff
@@ -282,3 +282,75 @@ def reconcile_user_link_counts(ctx, dry_run=False):
...
+def reconcile_registrar_link_counts(ctx, dry_run=False):
+    """
+    Set Registrar.link_count to the count of non-deleted links under its organizations.
+    Sponsored and personal links are not included. Pass dry_run=True to count mismatches without updating.
+    """
+    logger.info("reconcile_registrar_link_counts: Preparing to count links per registrar.")
+    link_counts = dict(
+        Link.objects
+        .filter(organization_id__isnull=False)
+        .values("organization__registrar_id")
```
rebeccacremona, line 332: I note that because of the join, we can expect this query to be more expensive than the others... but again, hopefully still feasible.

rebeccacremona (commented):
Great work: it is just amazing to see how intricate all this is, and exciting to see more pieces of it coming into alignment.

I asked a bunch of questions, including one, "am I right that this could be simpler" question?

Marking this review as "comment" rather than "approve" or "request changes" for now; if after discussion turns out I am confused, will switch to "approved"!

perma_web/perma/models/folder.py
teovin: %100 not necessary; I was thinking of batch moves mistakenly. I simplified it.

perma_web/perma/models/folder.py
teovin: Ohh, yep, very much stylistic, and I don't think I paid much attention to that first comment 😳 I moved it!

perma_web/perma/tests/test_link_count_caching.py
teovin: Just added a test for this case to document org/registrar counts changing while link user counts staying the same.

perma_web/tasks/once.py
teovin: Here is what it looks like in my machine with a 1 year old-ish data: 
image

teovin: I think there is definitely the possibility of race conditions, like somebody creating/moving links between the counting and writing stages. Although that window is extremely small. And running this during a low-traffic time of the day would minimize it even further.

perma_web/perma/models/folder.py
teovin: Do you mind creating a ticket for that? I am happy to look at it after I am done with these.

perma_web/tasks/once.py
rebeccacremona: Cool, if it's anywhere near as quick as what you are seeing (will surely be somewhat slower... but that's plenty fast!), agreed, not really a worry. We could run it a few times in a row if we were worried 👍

perma_web/perma/models/folder.py
rebeccacremona: Sure, can do! When I wrote the above, I didn't know that you had refactored that file in [this other PR](https://github.com/harvard-lil/perma/pull/3831/changes#diff-aa14409e24744897938c83ca1bf7ed81bfbe496a38f6c8d873977af26eb8b515). Would you still like a Linear issue? I definitely can create one!

teovin: Haha, and I realized you had commented on the other PR as well after responding to this comment  :D Yep, a ticket would work. I will open a different PR for that. All the merge conflicts, rebasing etc are making these a bit cumbersome.

rebeccacremona: Cool, see LIL-5729

rebeccacremona (approved):
Great, thanks for these updates! LGTM!

PR #3831, opened 2026-09-16
add sponsored_link_count field, and add logic

teovin:
This PR addresses the below when it comes to registrar model's `sponsored_link_count` field:

1- It adds a new `sponsored_link_count` field to store the number of links in folders sponsored by a registrar, not including org links.
2- It adds logic to update the new field when links are moved around. For example, creating a new link in a sponsored folder, and moving links between org <> sponsored and personal <> sponsored folders. 
3- It adds logic to update the new field when folders are moved around, such as moving folders between personal, org, and sponsored folders. 
5- It refactors `signals.py` a bit. 
6- Adds a few tests. 

It doesn't reconcile that field yet. I will add that piece in after I rebase this when [3828](https://github.com/harvard-lil/perma/pull/3828) is merged.


perma_web/perma/models/folder.py
```diff
@@ -147,6 +147,23 @@ def update_cached_path(ids, tree_root_id):
...
+            num_of_links = links.count()
+            if not num_of_links:
+                return
+
+            if old_sponsored_by_id:
+                Registrar.objects.filter(pk=old_sponsored_by_id).update(sponsored_link_count=F('sponsored_link_count') - num_of_links)
+
+            if new_sponsored_by_id:
+                Registrar.objects.filter(pk=new_sponsored_by_id).update(sponsored_link_count=F('sponsored_link_count') + num_of_links)
+
```
rebeccacremona, line 166: This code looks a lot like what I was asking about [here](https://github.com/harvard-lil/perma/pull/3828/changes#r4074071972): which is to say, whether the method for updating org counts if a folder moves, could possible also look shorter like this, or if that is more intricate? Not to cross the review streams lol.

perma_web/perma/models/link.py
```diff
@@ -298,29 +298,43 @@ def get_canonical_guid(self, guid):
...
+        from .registrar import Registrar
+
         with transaction.atomic():
+            destination_folder = folder
             # Don't let anybody move folders around, until this link is
             # safely inside its destination folder, lest denormalized
             # ownership-related fields get out of sync
-            for folder in itertools.chain(self.folders.all(), [folder]):
-                Folder.objects.select_for_update().get(pk=folder.tree_root_id)
+            for locked_folder in itertools.chain(self.folders.all(), [destination_folder]):
```
rebeccacremona, line 319: This renaming is great.

perma_web/perma/models/folder.py
```diff
@@ -147,6 +147,23 @@ def update_cached_path(ids, tree_root_id):
                 )[:1]
             )
 
+        def update_registrar_sponsored_link_counts(links, old_sponsored_by_id, new_sponsored_by_id):
```
rebeccacremona, line 150: I note that this handles if you move a folder from one sponsoring registrar to another sponsoring registrar and if you move a folder from Personal Links into Sponsored Links, or vice versa.

Unless I'm mistaken, we don't currently have code for the possibility that someone moves a folder out of an org and into a sponsored folder, or out of a sponsored folder and into an org. Ooof, so many possibilities.....

perma_web/perma/signals.py
```diff
@@ -3,58 +3,62 @@
...
+            registrar.link_count += 1
+            registrar.save(update_fields=["link_count"])
+
     try:
+        incoming_link = instance
         # include user deleted links, so we don't hit the new link signal when a link is deleted and later saved
-        loaded_link = sender.objects.all_with_deleted().get(pk=instance.pk)
-        if loaded_link.user_deleted and instance.user_deleted:
+        existing_link = sender.objects.all_with_deleted().get(pk=incoming_link.pk)
+        if existing_link.user_deleted and incoming_link.user_deleted:
```
rebeccacremona, line 40: Good idea to change up the naming here. Signals are confusing enough.

perma_web/perma/signals.py
```diff
@@ -3,58 +3,62 @@
...
+
+    def increment_organization_link_counts(link):
+        """ plus one to user's organization and associated registrar """
+        if link.organization:
+            organization = link.organization
+            registrar = organization.registrar
+            organization.link_count += 1
+            organization.save(update_fields=["link_count"])
+            registrar.link_count += 1
+            registrar.save(update_fields=["link_count"])
```
rebeccacremona, line 34: Since you are refactoring, this seems like a great time to update all these to use `F`, to avoid race conditions! Crossing the streams on reviews again lol.

perma_web/perma/signals.py
```diff
@@ -3,58 +3,62 @@
...
-
-        if instance.organization:
-            instance.organization.link_count += 1
-            instance.organization.save()
-            instance.organization.registrar.link_count += 1
-            instance.organization.registrar.save()
+        # new link, let's add it to the user, org and registrar counts
+        incoming_link.created_by.link_count += 1
+        incoming_link.created_by.save(update_fields=["link_count"])
+        increment_organization_link_counts(incoming_link)
```
rebeccacremona, line 61: Do we need a line for updating the registrar's sponsored link out here too?

perma_web/perma/signals.py
```diff
@@ -3,58 +3,62 @@
...
+        organization_changed = existing_link.organization != incoming_link.organization
+        if organization_changed:
+            decrement_organization_link_counts(existing_link)
+
+        if incoming_link.user_deleted and not existing_link.user_deleted:
+            if existing_link.created_by.link_count > 0:
+                existing_link.created_by.link_count -= 1
+                existing_link.created_by.save(update_fields=["link_count"])
+            decrement_organization_link_counts(existing_link)
+            Registrar.adjust_sponsored_link_count(existing_link._sponsored_by_id(), None)
```
rebeccacremona, line 52: I like how that reads as a class method.

perma_web/perma/tests/test_cutover_migrations.py
```diff
@@ -81,7 +81,7 @@ def record_sql(execute, sql, params, many, context):
             assert cursor.fetchone() == ('',)
         sql = '\n'.join(statements).upper()
         assert 'ADD COLUMN SESSION_HASH' in sql
-        assert 'DROP DEFAULT' not in sql
+        assert not any('SESSION_HASH' in s.upper() and 'DROP DEFAULT' in s.upper() for s in statements)
```
rebeccacremona, line 84: Calling out in case this is accidentally in the diff because of rebasing/branch changes.

perma_web/perma/models/registrar.py
```diff
@@ -57,7 +57,8 @@ class Registrar(CustomerModel):
     notes = models.TextField(blank=True)
     manual_sort_order = models.IntegerField(default=0, db_index=True)
 
-    link_count = models.IntegerField(default=0) # A cache of the number of links under this registrars's purview (sum of all associated org links)
+    link_count = models.IntegerField(default=0) # number of non-deleted links in this registrar's orgs; not including sponsored links
+    sponsored_link_count = models.IntegerField(default=0) # number of non-deleted links in folders sponsored by this registrar; not including org links
```
rebeccacremona, line 61: Probably worth adding this new field in admin.py

rebeccacremona (changes requested):
So much work!!!!

This project feels like it's getting really close.

I made some suggestions for more code to add to this PR. If you have other plans, just lemme know.

perma_web/perma/models/folder.py
teovin: Yep, I think I handled it in the other PR, but let me know.

perma_web/perma/models/folder.py
teovin: I think a user who is an org user but also a sponsored user can move folders between org, personal and sponsored folders. Oof, agreed... :)

perma_web/perma/signals.py
teovin: I don't think so. When a new link is created in a sponsored folder, the `move_to_folder_for_user` method is called, which is already adjusting the count on line 337.

perma_web/perma/tests/test_cutover_migrations.py
teovin: This is actually something I modified. A previous [change](https://github.com/harvard-lil/perma/commit/878c2ed31cd93684296535d32b68e76e01084c10#diff-b14770525e9d8c45bea568221e3a0e91c90a3355f73a529b930bcf8776a4da27) of Jack added a test that is checking if sql stataments contain DROP DEFAULT in a table he is preventing this from happening, but instead of his table, it was applying it to the whole schema. And because my new column has a default of 0, and from what I am reading, Django, after adding the default 0 to existing rows in the db, generates a statement to remove the database default, and the test was failing. And his test was catching this. So I modified it to only look at the table with SESSION_HASH column, which is the table he is interested in. I am not 100 percent confident with this explanation, but this is what I could gather.

perma_web/perma/models/folder.py
rebeccacremona: Yep, seen, LGTM!

perma_web/perma/signals.py
rebeccacremona: > line 337

Hmmm, since this file is changing in multiple PRs, I'm having trouble spotting which line 337 is being referenced 😅.  

But, I think this is pointing out something strange, right: why would sponsored link-related counts get updated in one spot, and org-related counts get updated in a different spot? 

Unless I'm mistaken, `link.move_to_folder_for_user` is called every time a link is created. Do you agree? (https://github.com/harvard-lil/perma/blob/ee143c6310434946fe9226d254af5abeab989486/perma_web/api/views/links.py#L234)

Do you understand if there is a reason that all this code isn't in one spot, or if that's just an as-of-now unrealized refactoring/simplification opportunity?

perma_web/perma/tests/test_cutover_migrations.py
rebeccacremona: Ah, cool! Gotcha. FWIW I just asked for more info about that PR here: https://github.com/harvard-lil/perma/pull/3852#issuecomment-5799053422

But yep, I see why you had to add that for this to pass. 

That's interesting; I am not particularly informed about when/why Django does that: probably not worth any of us spending any time thinking about it, but I think it's because Django generally prefers to handle defaults on the Python side, rather than on the DB side, but sometimes uses them temporarily during migrations...

jcushman:
Thanks for cleaning up test_cutover_migrations.py. I should probably remove that entirely now that the cutover is complete; I'll hold off until this all gets settled out so I don't cause more churn.

perma_web/perma/signals.py
teovin: When we are creating the link, the pre-save signal will fire before the link is actually saved, so the folder information is not known yet. For organization, it reads the organization information from the incoming link instance itself, so that information is already available.

rebeccacremona: Interesting. I might still want to make the case that maybe ALL of this should go inside the folder handler, instead of a signal on `Link`... This is feeling like an accident (we happen to cache `organization_id`, redundantly, on the Link objects; we don't happen to cache `registrar_id`), rather than an intentional design. But right, lol, not here! TBD later, maybe!!

teovin: Cool, I might give it a shot when looking at the F ticket.

perma_web/perma/signals.py
rebeccacremona: Update: we know this is buggy and can make these counts drift over time. See LIL-5729. Will be addressed in a future PR, and we can plan to re-run the invoke tasks to re-calculate/clean up all the link counts then, once everything has settled.

perma_web/perma/models/folder.py
rebeccacremona: Marking unresolved just for now, pending further discussion: this is the only reason this is still marked "Changes requested": will mark as approved after we're on the same page. I definitely might have just missed the code that makes this work!

teovin: Just added a test for this scenario. Because I had separated the org_count and this work into separate PRs, the org count update was coming from that other change. And rebasing things added that piece. Tested locally.

perma_web/perma/tests/test_link_count_caching.py
```diff
@@ -349,6 +349,55 @@ def test_moving_folder_between_sponsored_registrars_transfers_sponsored_link_cou
...
+
+    org_user.refresh_from_db()
+    organization.refresh_from_db()
+    org_registrar.refresh_from_db()
+    sponsoring_registrar.refresh_from_db()
+    assert org_user.link_count == user_count
+    assert organization.link_count == 0
+    assert org_registrar.link_count == 0
+    assert sponsoring_registrar.sponsored_link_count == 2
+    assert sponsoring_registrar.link_count == 0
```
rebeccacremona: I love to see the test fixtures serving their purpose elegantly 🤘 🎉

rebeccacremona (approved):
Thank you for adding that test! I don't know why I am having trouble reading through and understanding how that happens: I'll re-read all this stuff later!

LGTM!

perma_web/perma/models/folder.py
rebeccacremona: Ah! Gotcha!! Thx!

PR #3832, opened 2026-09-17
Avoid archive reads and conversion jobs for HEAD requests

jcushman:
Something I haven't thought about much -- by default Django treats a HEAD request as a GET, and expects the reverse proxy like nginx to strip the body. That's inefficient if a view does a lot of work. You can avoid that by checking the request method along the way and returning a response with an empty body.

This short circuits that for the view where it most bites in Perma, archive fetches.


perma_web/perma/utils.py
```diff
@@ -895,14 +895,38 @@ def get_wacz_stream(link, stream=True):
...
     # `link.user_deleted` is checked here for dev convenience:
     # it's easy to forget that deleted Perma Links' files aren't truly deleted,
     # and easy to accidentally permit the downloading of "deleted" archive files.
     # Users of stream_archive shouldn't have to worry about / remember this.
     if link.user_deleted or not link.can_play_back():
         raise Http404
 
+    if head:
+        # Check stored files without downloading them. WARC extraction still
+        # needs the ZIP directory to verify that the requested member exists.
```
rebeccacremona, line 908: `WARC extraction still needs the ZIP directory to verify that the requested member exists.`

While technically true, in reality, I think this is not worth checking. All WACZ will definitely contain a WARC, and if the file extraction fails due to some bug related to the name of the member or whatever, I think that should correctly error out during a later extraction attempt.

I'd recommend rewriting to assume the presence of the WACZ implies the presence of the WARC inside it.

perma_web/perma/utils.py
```diff
@@ -895,14 +895,38 @@ def get_wacz_stream(link, stream=True):
...
+                storage, name, size = settings.WACZ_STORAGE, link.wacz_storage_file(), link.wacz_size
+            else:
+                raise NotImplementedError("Unsupported file format.")
+            if not size or not storages[storage].exists(name):
+                raise Http404
+        suffix, content_type = ('warc.gz', 'application/gzip') if file_format == 'warc' else ('wacz', 'application/wacz')
+        # Streaming prevents CommonMiddleware from inventing a zero file size.
+        response = StreamingHttpResponse((), content_type=content_type)
+        response['Content-Disposition'] = f'attachment; filename="{link.guid}.{suffix}"'
+        return response
```
rebeccacremona, line 928: That's interesting; I wouldn't have expected a streaming response to `HEAD`, but I guess it makes sense.

perma_web/perma/tests/test_head_downloads.py
```diff
@@ -0,0 +1,131 @@
...
+    from perma.views import playback
+    settings.WARC_TO_WACZ_ON_DEMAND = True
+    link = Mock(
+        guid='ABCD-1234', replacement_link_id=None, warc_size=1, wacz_size=0,
+        is_user_uploaded=False, user_deleted=False, submitted_description='description',
+    )
+    link.primary_capture.status = 'success'
+    link.has_wacz_version.return_value = False
+    link.is_visible_to_memento.return_value = False
+    user = Mock(is_authenticated=False)
```
rebeccacremona, line 122: It seems non-idiomatic that these tests are using `Mock` for links and users, rather than using pytest fixtures, adding new ones if the existing Link and User factories aren't sufficient.

Would you mind changing that, unless this is for some practical reason I'm not thinking of? I feel like our test suite is in disarray already, and adding another idiom if we can avoid it... should be avoided.

But also maybe I'm not thinking of some reason Mocks are better?

perma_web/perma/tests/test_head_downloads.py
```diff
@@ -0,0 +1,131 @@
...
+
+@pytest.fixture
+def link():
+    return SimpleNamespace(
+        guid='ABCD-1234', user_deleted=False, warc_size=100, wacz_size=100,
+        can_play_back=Mock(return_value=True),
+        warc_storage_file=Mock(return_value='archive.warc.gz'),
+        wacz_storage_file=Mock(return_value='archive.wacz'),
+        get_warc=Mock(), get_wacz=Mock(),
+    )
```
rebeccacremona, line 26: I think by convention for this repo, all pytest fixtures go in conftest.py

Since this `link` mocks so many methods, I can see why it is convenient to just make a totally fake object, with the methods required on it, rather than using an actual `Link` object, and then mocking in spots.

But do you agree that this would be both more idiomatic and a slightly more reliable test, if using one of the pytest-factoryboy mediated `link` objects?

rebeccacremona (changes requested):
Great: yes, very into the idea of explicitly accounting for HEAD requests.

Some remarks on the implementation here; lemme know if you think they are off base.

rebeccacremona:
P.S. I think the tests are failing because there's a mock here that's no longer needed, after your rebase.

PR #3839, opened 2026-09-18
Maint/upd dep/aug26

LegallyGeek:
Upgrading the python, node, django, and other deps. to current LTS versions.


jcushman:
This is now incorporated on `main`.

PR #3840, opened 2026-09-18
update registrar count when org registrar is changed

teovin:
This PR addresses these two scenarios when it comes to registrar `link_count` field accuracy.

1- Registrar changes made through the `manage/organizations/id` view
2- Registrar changes made through the admin `admin/perma/organization/id/change/` view

In both cases, we need to decrement the org's link count from the previous registrar and add it to the new registrar, which is what this PR proposes.


perma_web/perma/models/organization.py
```diff
@@ -80,6 +83,17 @@ def save(self, *args, **kwargs):
...
+    def _transfer_registrar_link_counts(self, old_registrar_id, new_registrar_id):
+        """ Move this org's link_count between registrars when registrar changes. """
+        if old_registrar_id == new_registrar_id or not self.link_count:
+            return
+
+        if old_registrar_id:
+            Registrar.objects.filter(pk=old_registrar_id).update(link_count=F('link_count') - self.link_count)
+
+        if new_registrar_id:
+            Registrar.objects.filter(pk=new_registrar_id).update(link_count=F('link_count') + self.link_count)
```
rebeccacremona, line 95: Small dumb question: if we reached this path, where `old_registrar_id == new_registrar_id`, in what case could either of them be falsey? Could there be a case where it was appropriate to increment _somebody_ without decrementing somebody else? In short, are the `if old_registrar_id` and `if new_registrar_id` checks necessary?

rebeccacremona (approved):
Okay WOW: I genuinely had no idea that it was possible to transfer an organization between registrars, using the UI. 

Marking as approved, along with one question about a possible refactor.

perma_web/perma/models/organization.py
teovin: Good point! Removed those checks. The forms don't allow saving without registrars anyway.

PR #3844, opened 2026-09-18
Remove retired account-merge task

durkinm2:
A follow-up to https://github.com/harvard-lil/perma/pull/3826:

`tasks/merge_accounts.py` is unused and references a removed method. Now that a record of what it did, with links to the code at pinned commits, exists in a dedicated location (see LIL-5389), I believe we can comfortably remove here.

Also unregisters it from `tasks/__init__.py`.


rebeccacremona (approved):
LGTM! Thanks for creating the wiki page, and following up here!

Will be fun to delete some more, in the near future :-)

PR #3852, opened 2026-09-21
Prepare migrations for an online Salt-to-ECS handoff

jcushman:
I didn't want to require downtime during the cutover from salt VMs to ECS, so this compacts and simplifies the migrations so they're safe to apply live. Specifically:

- Tweak the Axes migration so it provides a default, keeping it working with existing code. Do this in one step instead of two so we don't add and remove a mandatory field.
- Drop three date indexes on history tables, automatically added by a django-simple-history update. If we need those we can add them later, but I don't think they were intentional.

The old versions of these had applied to staging, I think nowhere else; I'll fix that one up manually to match what's here.


rebeccacremona:
@jcushman I am reading this PR apropos of a [change Ebru had to make to it](https://github.com/harvard-lil/perma/pull/3831/changes/BASE..ed74105e62ac312dde1ad42b2097dff81ecf47aa#diff-b14770525e9d8c45bea568221e3a0e91c90a3355f73a529b930bcf8776a4da27R84) in another PR.

The PR describes these changes as having to do with the Salt -> ECS transition, but I gather what that means in more detail is, you were preparing for us to be able to run the application on the old architecture and the new architecture at the same time against the same database, with upgraded python packages in ECS, but without upgraded packages on the original infrastucture. Is that true?

If not, could you add a comment here correcting my guess, for our records?

jcushman:
Yes, that's correct! It was actually the salt -> ECS transition combined with Jenny's cross-cutting Django upgrade, which were both going live at once. I didn't want a 10 minute downtime into an untested-in-prod hosting environment, so I changed the django migrations to be able to run live, and to work with both the before and after code, instead of requiring any locks. That let us ship the new ECS web workers, test them, flip over live traffic, test again, and be able to switch back any time if there was an issue.

Specifically:
- the axes upgrade added a new NOT NULL session_hash column. I modified the migration to provide a DB-level default so old Axes inserts wouldn't fail if we flipped back to salt.
- the django-simple-history upgrade added automatic history_date indexes, which could have locked the db for a long time, so I told it not do that.

test_cutover_migrations.py was paranoia to be sure we couldn't break prod during that cutover; it can be retired, and maybe we want to go back and remove the axes override. I don't want to add more churn for Ebru, though, I can follow up with that later if it's still needed.

rebeccacremona:
Awesome, thanks for that!

jcushman:
In retrospect the benefits of combining them (only do a full-site QA pass once for two cross-cutting changes) didn't necessarily justify the drawbacks (manage a deploy that changed the code and the hosting environment at the same time and be able to roll back from both).

